PR #30, opened 2019-10-02
Parse open files as standalone if not included in project.

TwitchBronBron:
This enhances the LanguageServer so that it supports basic file parsing and syntax checking regardless of whether you have a bsconfig.json file created or not. It will only parse "open" files, and once the files are closed their diagnostics are removed.


TwitchBronBron:
Merging

PR #32, opened 2019-11-08
Fix race condition between project run and opened files

chrisdp:
This addresses some errors with reading tokens on hovers and also addresses some files not being detected in the project correctly on open.


TwitchBronBron:
that tslint error seems unrelated to your changes, so I'm accepting the PR. Thanks for these fixes!

Issue #1, opened 2019-06-07
Class specification


TwitchBronBron:
This is a proposal for classes. Here is the brighterscript syntax:

# Syntax
```VB
class VideoPlayer
    'public constructor
    public sub New(url as string)
        m.url = url
    end sub

    'properties
    public url as string
    private isPlaying as boolean = false

    'public method
    public play() 
        m.isPlaying = true
        m.updateVideoStatus()
    end function

    'private method
    private updateVideoStatus()
        'do something private here
    end sub

end class
```

# Output

```brightscript
function VideoPlayer(url as string)
    instance = {
        url: nothing,
        'underscore denotes private
        _isPlaying: nothing,
        play: VideoPlayer_play,
        _updateVideoStatus: VideoPlayer__updateVideoStatus
    }
    
    'constructor body
    instance.firstName = firstName
    instance.lastName = lastName
    instance.fullName = firstName + " " + lastName
    
    return instance
end function

function VideoPlayer_play()
    m.isPlaying = true
    m._updateVideoStatus()
end function

'double underscore (one for namespacing, the other for indicating private

function VideoPlayer__updateVideoStatus()
    'do something private here
end sub

```

# Usage
## BrighterScript
```brightscript
'usage
sub main()
    player = new VideoPlayer("some_url")
    player.play()
end sub
```

## BrightScript
```brightscript
sub main()
    player = VideoPlayer("some_url")
    player.play()
end sub
```

georgejecook:
can you please demosntrate where the constructor (i.e. `New`) code goes to?
Also how does namespacing affect this?

georgejecook:
also - can we get the className, or some other identifying type, for class comparison purposes?

TwitchBronBron:
There already a comment in the output showing you where the constructor body goes. 

And the class name is a great idea!

georgejecook:
ah - I missed it, as I was expecting to see the `m.url = url`

aaaannndddd - we should be able to extend, even if it's just brute force overrides.

chrisdp:
Is there anyway we could do classes ether as AA out put or as Node output. I say this because internally we avoid AA's as much as possible.

georgejecook:
this is a good suggestion - so something like `NodeClass`? whereby 

- automatcially creates a className.xml, 
- imports the declaring brs file into className.xml
- the public methods become `<function>` interface entries in className.xml
- public fields become `<field>` entries on className.xml

[...]

PR #40, opened 2020-04-01
Added baseline class specification

TwitchBronBron:
Describes the approach for #1


TwitchBronBron:
@georgejecook Can you please take a look at my class inheritance BrightscriptOutput, and make sure that will satisfy all of your concerns? I think I eliminated the need for a separate builder for classes by aliasing the super methods with their index (i.e. `super0`, `super1`, `super2`, etc...)

docs/classes.md
```diff
@@ -0,0 +1,428 @@
...
+    'specified private
+    instance.setSocialSecurityNumber = sub(value as string)
+        m.socialSecurityNumber = value
+    end sub
+    return instance
+end function
+```
+</details>
+
+You will get compile-time errors whenever you access private members of a class from outside the class. However, be aware that this is only a compile-time restriction. At runtime, all members are public.
```
georgejecook, line 372: what is the position on extended properties? I see no protected keyword (I think I have submitted an issue for that, if not I shall); but for now, do we need to be explicit about how you intend  private visibility to work with inheritance?

docs/classes.md
```diff
@@ -0,0 +1,428 @@
...
+    instance.getAge = function()
+        return m.age
+    end function
+    return instance
+end function
+```
+</details>
+
+## Usage
+In order to use a class, you need to construct one. Based on our person class above, you can create a new person like this:
```
georgejecook, line 414: worth explicitly calling out the new `new` keyword?

docs/classes.md
TwitchBronBron: The position is, they don't exist. A child class and super class will share all of the same properties. The compiler will add an error if a child and parent declare a property with the same name.

docs/classes.md
TwitchBronBron: That's already explained in this section: https://github.com/rokucommunity/brighterscript/pull/40/files#diff-3031a75b5d15117a376b3869ba2a5b24R143

But I could probably add a little better documentation around it, or maybe declare that section higher up?

docs/classes.md
```diff
@@ -0,0 +1,428 @@
...
+    instance = __Duck_Build()
+    instance.new(name)
+    return instance
+end function
+
+function __BabyDuck_Build()
+    instance = __Duck_Build()
+    instance.super1_move = instance.move
+    instance.move = sub(distanceInMeters as integer)
+        instance.super1_move(distanceInMeters)
```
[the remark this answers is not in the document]
TwitchBronBron: Oops, that's a typo. You're right, it needs to be `m`. I updated the docs in master. Thanks for catching this!

PR #42, opened 2020-04-03
Throw an error when running once

slheavner:
Do this so running from the command line will recognize that linting failed


TwitchBronBron:
Nice, I was about to start on this. Thanks. I'll get the ci builds working, you didn't break them.

PR #46, opened 2020-04-07
Feature: ignore files for validation

slheavner:


slheavner:
need to normalize paths so windows works

TwitchBronBron:
Thanks for the idea. After our discussions on slack, we decided to go a different direction. See the `diagnosticFilters`  option in bsconfig.json for more information.

PR #93, opened 2020-05-28
chore(docs): Some additional info about reasons to use brighterscript

georgejecook:


TwitchBronBron:
Thanks for this, way better than what I had!

PR #103, opened 2020-06-06
Support for ternary operator

georgejecook:
- adds ConditionalExpression
- decouples questionmark token from print
- adds parsing for conditional expression
- adds support for ConditionalStatement, which needs to be removed, because we can't support it due to loverloaded equals/assign operator
- adds util method to gather variables under a given syntax branch, which will be used when transpiling conditionals to gather scope


.vscode/settings.json
```diff
@@ -22,5 +22,14 @@
     ],
     "mocha.env": {
         "TS_NODE_TYPE_CHECK": false
+    },
```
TwitchBronBron, line 25: Make sure to exclude these once you are finished working on this.

src/parser/Expression.ts
```diff
@@ -645,5 +645,112 @@ export class CallfuncExpression implements Expression {
...
+    transpile(state: TranspileState) {
+        let result = [];
+        let [testExpressions, testVarExpressions, testUniqueNames] = this.getExpressionInfo(this.test);
+        let [consequentExpressions, consequentVarExpressions, consequentUniqueNames] = this.getExpressionInfo(this.consequent);
+        let [alternateExpressions, alternateVarExpressions, alternateUniqueNames] = this.getExpressionInfo(this.alternate);
+
+        let allExpressions = [...testExpressions, ...consequentExpressions, ...alternateExpressions];
+        let allUniqueVarNames = [...new Set([...testUniqueNames, ...consequentUniqueNames, ...alternateUniqueNames])];
+
+        let mutatingExpressions = allExpressions.filter(e => e instanceof CallExpression || e instanceof CallfuncExpression || e instanceof DottedGetExpression);
```
TwitchBronBron, line 673: We might need to also include object property references as "expensive" enough to run in the deferred function, because rendezvous could be fairly slow, and the "bslib_simpleTernary" would cause TWO rendezvous.

@chrisdp what do you think?

src/parser/Expression.ts
```diff
@@ -645,5 +645,112 @@ export class CallfuncExpression implements Expression {
...
+            // eslint-disable-next-line no-return-assign
+            allUniqueVarNames.forEach(name => scope += `\n"${name}": name`);
+            scope += '}';
+
+            result.push(`bslib_scopeSafeTernary(${this.test.transpile(state)}, ${scope},`);
+            result = result.concat(this.getScopedFunction(state, this.consequent, consequentUniqueNames));
+            result = result.concat(this.getScopedFunction(state, this.alternate, alternateUniqueNames));
+            result.push(')');
+        } else {
+            const ternaryCall = `bslib_simpleTernary(${this.test.transpile(state)}, ${this.consequent.transpile(state)}, ${this.alternate.transpile((state))})`;
```
TwitchBronBron, line 691: This isn't right. transpile returns `Array<string|SourceNode` so this would end up calling `array.toString()`. You need to break these out.

Something to the effect of
```
result.push('bslib_simpleTernary(');
result.push(...this.test.transpile(state));
result.push(', ');
result.push(...this.consequent.transpile(state));
result.push(', ');
result.push(...this.alternate.transpile(state));
result.push(')');
```

src/parser/tests/expression/ConditionalExpression.spec.ts
```diff
@@ -0,0 +1,266 @@
...
+    CallExpression,
+    ConditionalExpression,
+    LiteralExpression
+} from "../../Expression";
+
+describe('parser conditional expressions', () => {
+    it('throws exception when used in brightscript scope', () => {
+        let { tokens } = Lexer.scan(`a = true ? "human" : "Zombie"`);
+        let { diagnostics } = Parser.parse(tokens, { mode: ParseMode.BrightScript });
+        expect(diagnostics[0]?.code).to.equal(DiagnosticMessages.bsFeatureNotSupportedInBrsFiles('').code);
```
TwitchBronBron, line 21: Can you change this to check the message instead? It makes diagnosing errors a little easier since you don't have to go look up the code.

src/parser/Expression.ts
```diff
@@ -645,5 +645,112 @@ export class CallfuncExpression implements Expression {
...
+            result.push(new SourceNode(this.range.start.line + 1, this.range.start.character, state.pathAbsolute, ternaryCall));
+        }
+        return result;
+    }
+
+    private getExpressionInfo(expression: Expression): [Expression[], Expression[], string[]] {
+        const expressions = getAllExpressions(expression);
+        const varExpressions = expressions.filter(e => e instanceof VariableExpression) as VariableExpression[];
+        const uniqueVarNames = [...new Set(varExpressions.map(e => e.name.text))];
+        return [expressions, varExpressions, uniqueVarNames];
```
TwitchBronBron, line 702: I'm not a big fan of using array index locations to represent different values, that's what objects are for. Can you change this to return an object with names?
```
return {
    expressions: expressions,
    variableExpressions: varExpressions,
    variableNames: uniqueVarNames
};
```

Then up in the transpile function you can extract them to local vars in a similar fashion as you were doing:
```
let { expressions, variableExpressions, variableNames } = this.getExpresssionInfo(expr);
```

src/parser/Expression.ts
```diff
@@ -645,5 +645,112 @@ export class CallfuncExpression implements Expression {
...
+        return result;
+    }
+}
+
+/**
+ * Walks an expression, getting all nested expressions
+ * @param expression
+ * @param expressions
+ */
+function getAllExpressions(expression: Expression, expressions: Expression[] = []): Expression[] {
```
TwitchBronBron, line 723: Is there a way we can handle this without needing to hardcode every known expression type? Like perhaps we make each expression EXTEND Expression instead of implement the expression interface, then we can test for Expression in the prototype chain. I'm just worried about adding new expression types over time and not remembering to update this function to account for that.

TwitchBronBron:
I did a first pass over everything and left some comments.

I know you're not totally finished with the transpile stuff, but we need to make sure to add a bunch of testing for transpile output to make sure it works right. Here's a good example test to use: https://github.com/rokucommunity/brighterscript/blob/master/src/files/BrsFile.spec.ts#L1611

src/parser/Expression.ts
georgejecook: cool! I was wondering how to do this.

src/parser/Expression.ts
georgejecook: yes - I put a comment in here saying I think each expression should implement this method. Glad you concur. I'll make it so.

src/parser/Expression.ts
georgejecook: yep - I realized this in my ternary-null coalesce pr and fixed it there. Will fix it here too.

.vscode/settings.json
georgejecook: crap. sorry these keep getting in.

src/parser/tests/expression/ConditionalExpression.spec.ts
georgejecook: ok

src/parser/Expression.ts
georgejecook: any way of knowing that up front? I don't think so, no? coz we'd need to know it's a node type. Does that mean any dotteted, get then, irrespective of the presence function calls?

TwitchBronBron: @georgejecook I spoke with @chrisdp and we both agree, there are situations that could cause a rendezvous when accessing object members, so we cannot allow properties to follow the simpleTernary approach since that could be expensive in the long run. So instead we need to restrict the simpleTernary to only variables and literals.

TwitchBronBron: No, there's no way of knowing that up front. So yeah, any dotted get would also need to use the scope-capture implementation.

georgejecook:
hey @TwitchBronBron I've updated the code quite a bit; still got a cpl more things to do; so wait until I ping in here, in case you see me resolving things. There's a lot of commits here; figured I'd leave it all raw for now, and you can make a call about squashing/rebasing, etc.

I'll be done in 10 or so mins

georgejecook:
@TwitchBronBron I think this is ready to go - I just reverted package lock changes; I tried the same with settings; but I don't have any changes with master; so I'm not sure what is going on there.

georgejecook:
@TwitchBronBron you waiting on anything else from me? keen to get this in.

TwitchBronBron:
I think the code is good. However, we need to add a bunch of documentation examples before this can be merged. These PRs serve as part of the approval process for a new language feature. I'll add a few of them this weekend sometime and get feedback from interested parties.

georgejecook:
hey @TwitchBronBron I realized you were right, that dotted get needs to be scope safe, as we will never know for sure if the values are really there, as the developer will expect the alternate and onsequent to be mutually exclusive, and may therefore use values in the either, which could be invalid at runtime. I've updated the code. 

I've also added the documentation.

I hope it's on the right track. I need to add a couple more test cases, to ensure that dotted gets are covered by scope; but I think if you are happy with the

bslib.brs
```diff
@@ -21,4 +21,16 @@ function bslib_toString(value)
 
     'when all else fails, just return the type
     return "<" + valueType + ">"
+end function
+
+'
+' Simple ternary function. Given a condition, return the
+' consequent if true, and the alternate if false
+'
+function bslib_iff(condition, consequent, alternate)
```
TwitchBronBron, line 30: I called this `iff` instead of `ternary` just to save a few characters. If you have a large project with a bunch of source code, `ternary` would start to add up. Plus, `iff` is a [known visual basic function](https://docs.microsoft.com/en-us/office/vba/language/reference/user-interface-help/iif-function) so it still makes sense (at least to me). Thoughts?

src/lexer/TokenKind.ts
```diff
@@ -73,7 +73,7 @@ export enum TokenKind {
     Semicolon = 'Semicolon', // ;
     At = 'At', // @
     Callfunc = 'Callfunc', // @.
-
+    Question = 'Question', // ?
```
TwitchBronBron, line 76: I called this `Question` instead of `QuestionMark` because I'm going to name the `??` token `QuestionQuestion` for null coalescing. It felt silly to have `QuestionMark` and `QuestionQuestion`. And I don't like `QuestionMarkQuestionMark`.

Thoughts?

TwitchBronBron:
Whew, that merge from master was rough! 

@elsassph this is ready for your review.

src/parser/Expression.ts
```diff
@@ -1290,6 +1290,90 @@ export class AnnotationExpression extends Expression {
...
+        //get all unique variable names used in the consequent and alternate, and sort them alphabetically so the output is consistent
+        let allUniqueVarNames = [...new Set([...consequentInfo.uniqueVarNames, ...alternateInfo.uniqueVarNames])].sort();
+        let mutatingExpressions = [
+            ...consequentInfo.expressions,
+            ...alternateInfo.expressions
+        ].filter(e => e instanceof CallExpression || e instanceof CallfuncExpression || e instanceof DottedGetExpression);
+
+        if (mutatingExpressions.length > 0) {
+            result.push(
+                `(function(condition, `,
```
elsassph, line 1324: I think using `condition` for the arg name is risky. Maybe `bslib_cond` will be safer?

src/parser/Parser.ts
```diff
@@ -1187,6 +1188,36 @@ export class Parser {
         return annotation;
     }
 
+    private ternaryExpression(test?: Expression): TernaryExpression {
+        this.warnIfNotBrighterScriptMode('ternary operator');
+        if (!test) {
+            test = this.expression();
+        }
+        const questionMarkToken = this.advance();
```
elsassph, line 1196: Is/should a newline allowed here?
Maybe some wrapping at places can be safely allowed - and if yes/no it might be good to document.

src/parser/Parser.ts
```diff
@@ -1187,6 +1188,36 @@ export class Parser {
...
+                ...DiagnosticMessages.malformedTernaryExpression(),
+                range: test.range
+            });
+        }
+
+        const colonToken = this.tryConsume(DiagnosticMessages.malformedTernaryExpression(), TokenKind.Colon);
+        const alternate = this.expression();
+
+        if (!consequent || !alternate) {
+            this.diagnostics.push({
```
elsassph, line 1213: Looks like a malformed ternary will result into something that we'll still be able to transpile - will it work?

src/parser/Expression.ts
TwitchBronBron: Very good point. Will address

src/parser/Parser.ts
TwitchBronBron: Good point. I don't see any reason to prevent newlines within ternary statements. In brs, you can use the question mark in place of print, so it might be good to require the condition and the question mark to be in the same line. Thoughts?

The rest can definitely be flexible to use newlines at will. 

So something like this would be supported. 

```
someVar = true = true ?
                      getTrueValue():

                       getFalseValue()
```

TwitchBronBron: Any other expression brine already supported by brs elsewhere would be automatically supported too

src/parser/Parser.ts
TwitchBronBron: Probably not. However, I've tried very hard to still produce partial AST and as much AST recovery as possible, which helps with language server stuff. 

What should malformed transpiled ternary statement look like? Throw an error? Default to a no-op? Write an intentional syntax error so the code doesn't pass the on-device compile phase?

I've kinda just been letting the transpiler write whatever.

src/parser/Parser.ts
elsassph: Basically yes, `?` can't be on a newline but otherwise the rest could be wrapped.

src/DiagnosticMessages.ts
```diff
@@ -605,6 +605,11 @@ export let DiagnosticMessages = {
         message: `Missing exception expression after 'throw' keyword`,
         code: 1118,
         severity: DiagnosticSeverity.Error
+    }),
+    malformedTernaryExpression: () => ({
```
elsassph, line 609: Isn't there an unused code you could pick?

src/parser/Expression.ts
```diff
@@ -1290,6 +1290,90 @@ export class AnnotationExpression extends Expression {
...
+            test.range.start,
+            (alternate ?? colonToken ?? consequent ?? questionMarkToken ?? test).range.end
+        );
+    }
+
+    public range: Range;
+
+    transpile(state: TranspileState) {
+        let result = [];
+        let consequentInfo = util.getExpressionInfo(this.consequent);
```
elsassph, line 1312: Does all this code work if `consequent` or `alternate` is `null`? It can happen for malformed expression parsed.

src/parser/Expression.ts
TwitchBronBron: I went with `__bsCondition`

src/parser/Parser.ts
TwitchBronBron: Done. I also added support for mixing comments throughout (but they aren't included in the transpiled output because there's not much benefit in retaining them IMO. (Let me know if you disagree). 

Here are the tests for the various ways you can use 

```
    it('supports multi-line and comments', () => {
        expectZeroDiagnostics(parseBs(`result = true ? \n"zombie"\n: \ntrue`));
        expectZeroDiagnostics(parseBs(`result = true ? "zombie"\n: \ntrue`));
        expectZeroDiagnostics(parseBs(`result = true ? \n"zombie": \ntrue`));
        expectZeroDiagnostics(parseBs(`result = true ? \n"zombie"\n: true`));
        expectZeroDiagnostics(parseBs(`result = true ? "zombie"\n: \ntrue`));
        expectZeroDiagnostics(parseBs(`result = true ? "zombie": \ntrue`));
        expectZeroDiagnostics(parseBs(`result = true ? \n\n\n"zombie": \n\n\n\ntrue`));
        //with comments
        expectZeroDiagnostics(parseBs(`result = true ?'comment\n"zombie"'comment\n:'comment\nntrue`));
    });
    ```

File: src/DiagnosticMessages.ts (hunk 12 above)
Comment (TwitchBronBron) [reply src/DiagnosticMessages.ts:609]: Oh yeah, we had one left (1109). I switched to that.

File: src/parser/Expression.ts (hunk 13 above)
Comment (TwitchBronBron) [reply src/parser/Expression.ts:1312]: No, I think it will throw a node runtime error because we call `.walk` on the expression. How do you think it should be handled?

File: bslib.brs (hunk 7 above)
Comment (georgejecook) [reply bslib.brs]: I like.

File: src/lexer/TokenKind.ts (hunk 8 above)
Comment (georgejecook) [reply src/lexer/TokenKind.ts:76]: Agree with this too.

File: src/parser/Parser.ts (hunk 10 above)
Comment (georgejecook) [reply src/parser/Parser.ts:1196]: Cool

File: docs/ternary-operator.md [hunk 14]
```diff
@@ -0,0 +1,165 @@
+# Ternary (Conditional) Operator: ?
+The ternary (conditional) operator is the only BrighterScript operator that takes three operands: a condition followed by a question mark (?), then an expression to execute (consequent) if the condition is true followed by a colon (:), and finally the expression to execute (alternate) if the condition is false. This operator is frequently used as a shortcut for the if statement. It can be used in assignments, and in any other place where an expression is valid. Due to ambiguity in the brightscript syntax, ternary operators cannot be used as standalone statements. See the [No standalone statements](#no-standalone-statements) for more information.
+
+## Basic usage
+
+```BrighterScript
+a = user = invalid ? "no user" : "logged in"
+```
+
+transpiles to:
```
elsassph, line 10: I think these explanations are a bit confusing.

What about going with:
1. ternary operator with literal values: transpiles with `bslib_ternary`
2. ternary operator with references/function calls: transpiles into a self-executing function with scope capture

docs/ternary-operator.md
```diff
@@ -0,0 +1,165 @@
...
+    else
+        return falseValue
+    end if
+end function
+```
+
+## No standalone statements
+The BrighterScript ternary operator differs from languages like C# and JavaScript in that you cannot use ternary expressions as standalone expressions. This is due to the fact that BrightScript uses `=` for both assignment and equality. As such, ternary expressions must always be part of an assignment or method call. Take a look at the following expression:
+```
+a = myValue ? "a" : "b"
```
elsassph, line 156: I think we want to say that in a simpler way like: 
```
' this is not a valid statement
condition ? doSomething() : orElse()
```
Due to Brightscript syntax where an assignment like `a = b` can not be differentiated from a condition `if a = b`.

elsassph: BTW wouldn't that be ok?
```
' expression hanging alone
(condition ? doSomething() : orElse())
```

src/parser/Expression.ts
TwitchBronBron: I updated this logic so that it will not cause a runtime error during transpile. a null consequent or alternate will be transpiled as `invalid`.

docs/ternary-operator.md
TwitchBronBron: Which parts are "thought process" like? I wanted to be transparent in the way the transpiler works, and why we chose the implementations that we did. If there are specific sections that are too verbose, point them out and I'll see if there's a way to trim them down. but I did a quick read-through and still thought most of the information was generally relevant to someone reading the docs.

PR #105, opened 2020-06-07
Null coalescence

georgejecook:
example of how this looks with both null coalesence and ternary; they are so closely related, that imho they belong together in the same function.

I will still submit them as separate prs; but wanted you to get eyes on my thinking here.


bslib.brs
```diff
@@ -21,4 +21,27 @@ function bslib_toString(value)
...
+    else
+        return alternate
+    end if
+end function
+
+'
+' Return consequent if consequent is not invalid, otherwise return alternate
+'
+function bslib_coalesce(consequent, alternate)
+    if consequent <> invalid then
```
elsassph, line 42: Can the `consequent` be "uninitialized" as opposed to `invalid`?

TwitchBronBron: I spoke with @georgejecook and @triwav about this a bit, and we decided we agreed that none of us think the `null coalescing operator` should support `<uninitialized>`. Any object property access for undefined properties will return `invalid`, so there are very few cases where `<uninitialized>` could actually show up:
1. typing the name of a variable that was not defined
2. using the name of a loop variable before its first assignment.


Both of these are edge cases should happen very infrequently in the wild. In almost all cases, getting `uninitialized` is a program error, which is technically different than `invalid`.  

Here are the reasons we don't want to support `<uninitialized>`:
1. it will add a `type()` call to every ternary operator, which will incur a performance cost.
2. `<uninitialized>` is a program error, and _should_ cause a runtime exception in my opinion
3. it's such an edge case, it shouldn't happen all that frequently
4. type checking (or `bslint`) should catch both the use-before-initialized or unknown-variable during the compile phase

Thoughts?

elsassph: Sounds good.

PR #107, opened 2020-06-09
Template string implementation

georgejecook:
This is starting to get toward a final solution now:

- tokens are gathered for backtick, then quasiliterals, and expressions (interleaved) until backtick
- templateStringExpression contains quasiLiterals and expressions
- transpile function spits out a function call, to bslib, so we can concat, and repalce \n as we wish
- throws error if at EOF with no backtick
- chucked quite a few test cases at it, and it behaves well

requires:
- linting
- more tests
- some checking on the source map, to see if I've got that wrong


georgejecook:
I'd been thinking about this, so I laid it down over a couple of hours today, to clear my head while switching between 2 big projects (I know; I'm mental.. that's just what works for me :) )

Implementation is as follows:
 - lexer identifies a backtick, and records that token for the parser
 - lexer grabs all template string quasis (i.e. string literals) and expressions (i.e. code inside of ${)
 - when parser encounters backtick
 - it iterates over all quasi literals and expressions until next backtick, addig them to a parts array
 - and creates a template literal expression with the parts
 - template literal transpiles by concatting the transpile funcs of all the parts

image
image

georgejecook:
I am seeing one odd thing here; where there is some discrepency between string literal text, and string literal value.text.. not sure what that is
oh, and this is a wip, not a finished thing - needs a lot more test cases. Happy to proceed if you have no problems with the approach.

georgejecook:
whoa..

even handles this case..

```
a = ["one", "two", `I am a complex example
            ${a.isRunning(["a","b","c", `template ${"inside" + m.items[0]} template`])}`]
```
a = [
    "one",
    "two",
    stdlib_concat(["I am a complex example
            ", a.isRunning([
        "a",
        "b",
        "c",
        stdlib_concat(["template ", "inside" + m.items[0], " template"])
    ]), ""])
]
```

TwitchBronBron:
@georgejecook I'm assigning this back to you. I think I have everything in working order. I added tagged template strings, escaping char code literals, quotes, a few other things. Kick the tires a bit, and when you're ready, let me know and we can merge it.

TwitchBronBron:
@georgejecook this is ready for your final review.

PR #108, opened 2020-06-09
allow leading space for bs:disable comments

TwitchBronBron:
Currently the `bs:disable-line` and `bs:disable-next-line` comments do not support leading spaces. There's no reason not to, so this PR fixes that.


src/files/BrsFile.ts
```diff
@@ -121,16 +121,13 @@ export class BrsFile {
             throw new Error(`File was already processed. Create a new instance of BrsFile instead. ${this.pathAbsolute}`);
         }
 
-        //split the text into lines
-        let lines = util.getLines(fileContents);
```
TwitchBronBron, line 125: This is a nice benefit, we no longer need to split the file contents by line and process it, since `getIgnores` works on the list of tokens now.

src/util.ts
```diff
@@ -712,6 +713,70 @@ export class Util {
         }
     }
 
+    /**
+     * Small tokenizer for bs:disable comments
+     */
+    public tokenizeBsDisableComment(token: Token) {
```
TwitchBronBron, line 719: Simple little tokenizer/parser function that makes dealing with disable comments easier.

src/files/BrsFile.ts
```diff
@@ -334,94 +331,71 @@ export class BrsFile {
...
-                                    range: Range.create(lineIndex, idx, lineIndex, line.length),
-                                    affectedRange: affectedRange
-                                };
-                            }
-                        }
+            let affectedRange: Range;
+            if (tokenized.disableType === 'line') {
+                affectedRange = Range.create(token.range.start.line, 0, token.range.start.line, token.range.start.character);
+            } else if (tokenized.disableType === 'next-line') {
+                affectedRange = Range.create(token.range.start.line + 1, 0, token.range.start.line + 1, Number.MAX_SAFE_INTEGER);
```
TwitchBronBron, line 348: We used to find the length of the next line, but there's no real benefit to doing that, so just use MAX_SAFE_INTEGER instead because that will definitely match anything on the next line.

PR #126, opened 2020-07-03
Source literals

### LINKED ISSUE #13: additional language support functions: BS_LINENUM(), BS_FILE(), BS_FUNCTION()
wouldn't it be great if we could drop these into the code at any time, like macros, so that they could be resolved at compile time. These are super handy for debugging, and currently require the use of third party tools.

Suggestion

`print "unknown error occurred " ; BS_FILE() ; "." BS_FUNCTION() ;  BS_LINENUM() `

Would compile to 
`print "unknown error occurred pkg:/path/ToFile.brs.LoadStuff(23)"`

TwitchBronBron:
This PR adds several brighterscript-only source literals that get replaced at transpile-time. 
 - source_file_path
 - source_line_num
 - function_name
 - source_function_name
 - source_location
 - pkg_path
 - pkg_location

Resolves #13

### DIFF
--- a/docs/source-literals.md
+++ b/docs/source-literals.md
@@ -1,5 +1,5 @@
 # Source Literals
-BrightScript provides the `LINE_NUM` literal, which represents the runtime line number of the current line of code. In a similar fashion, BrighterScript provides several more source literals.
+BrightScript provides the `LINE_NUM` literal, which represents the runtime 1-based line number of the current line of code. In a similar fashion, BrighterScript provides several more source literals.
 
 These source literals are converted into inline variables at transpile-time, so keep in mind that any post-processing to the transpiled files could  cause these values to be incorrect.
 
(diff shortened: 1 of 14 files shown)


TwitchBronBron:
@georgejecook can you review this PR?

docs/source-literals.md
```diff
@@ -0,0 +1,186 @@
...
+## PKG_LOCATION
+A combination of PKG_PATH and LINE_NUM. Keep in mind, LINE_NUM is a runtime variable provided by Roku. 
+```BrighterScript
+print PKG_LOCATION
+```
+
+transpiles to: 
+
+```BrightScript
+print "pkg:/source/main.brs:" + str(LINE_NUM)
```
triwav, line 185: Was there a reason for using Roku’s vs the SOURCE_LINE_NUM?

TwitchBronBron: I figure since the pkgPath is basically the path to the file at runtime, it made sense to use the runtime line number too.

georgejecook:
I just tried it out locally; looks good

docs/source-literals.md
TwitchBronBron: @triwav do you agree with that approach, or do you think this should be something else?

georgejecook:
imho, `source_xxx` literals should honour the sourceRoot bsconfig setting, if that has been

TwitchBronBron:
@georgejecook I agree.

PR #130, opened 2020-07-04
Fix copy of output when `stagingFolderPath` isn't specified.

### LINKED ISSUE #129: CLI crash on Mac
I've picked a random Roku sample, which works nicely in VSCode and is following the default organisation, but I can't get it to work as `.bs` (on Mac):

- command line `bsc` (no `bsconfig.json`) fails with some `[Error: EACCES: permission denied, mkdir '/source']`, which possibly means some config defaults don't work,
- VSCode "Debug" doesn't automatically transpile; did I miss maybe an argument in the launch configuration?

elsassph:
Code was using the initial `options` instead of the prepared object.

Fixes #129

### DIFF
--- a/src/Program.ts
+++ b/src/Program.ts
@@ -676,10 +676,10 @@ export class Program {
 
         //copy the brighterscript stdlib to the output directory
         promises.push(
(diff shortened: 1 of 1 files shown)


TwitchBronBron:
Nice work. Thanks for digging into this.

PR #143, opened 2020-07-09
Fix: prevent accumulating multiple validations

### LINKED ISSUE #142: Prevent multiple validations in case of multi-file changes
When multiple files are modified (e.g. changes in multiple files, branch switching...), the validation will run once for every file, which can result in very long validation when changing a lot of files at once.

elsassph:
Decoupled validation from file-change handling, allowing to throttle validations.

Fixes #142


TwitchBronBron:
This looks great. Thanks!

PR #155, opened 2020-07-17
BrighterScript Reflection

### LINKED ISSUE #124: reflection
We need to have some of the following facilities:

- Know what kind of class an object is
- Get a reference to a class constructor, by it's name, so we can do create classes at runtime, from string class names
- Potentially get pointers to functions in a namespace
- Get the transpiled name of a method at runtime, so we can add observers without knowing implementation details.. e.g. `m.dialog.observeField("close", BS.Reflection.getMethodName(MyNamespace.Function))`, which will transpile to something like `m.dialog.observeField("close", "MyNamespace_Function")`

TwitchBronBron:
This PR provides the documentation and implementation of "reflection" support within BrighterScript. This PR is not yet complete, and has been created so that we can provide feedback and make alterations to the code as a single source, rather than the tickets which can get confusing. So, if you have thoughts, drop them in here and we'll update the branch accordingly. 

Resolves #124


TwitchBronBron:
Closing this for now. We have more pressing efforts underway. We can revisit this closer to the end of v1, or even later since this would most likely be a minor release adding new functionality.

PR #157, opened 2020-07-17
Type tracking

TwitchBronBron:
This is the start of type tracking. I'm opening this PR early, in draft mode, so that I can get feedback throughout the process rather than at the end. It probably don't compile most of the time. :) 

One thing to note, I scrapped the BrsType stuff in favor of raw tokens, because it was getting very confusing having two separate type systems. 

I intend on adding a linking pass within the compiler. So this is the expected flow:
 - lex
 - parse
 - link
 - validate

The linking will happen as a separate pass in this iteration so we can let the parser do parsing, and have a separate process handling variable lookups. It will also pave the way for eventually using class names, interfaces, unions, etc, as types, which will need to be scope-aware. 

This first pass will not support custom types (like classes).

Any feedback is welcome as we work through this!


TwitchBronBron:
The core of the type tracking system is contained within the `TypeReference` class. That allows one expression to "reference" the type of another expression.

```
name = "Bob"
nameCopy = name
```

`name` would have a typeRef pointing to the raw string token. Then `nameCopy` would have a typeRef pointing at name's typeRef. 
`name.typeRef` -> `StringType`
`nameCopy.typeRef`-> `name.typeRef` -> StringType

TwitchBronBron:
This PR is very outdated, and we have several other in-progress efforts to help get us further along, so I'm going to close this PR in favor of those.

PR #169, opened 2020-08-05
Fix cli overrides

### LINKED ISSUE #167: cli doesn't use bsconfig values for unspecified arguments
The CLI doesn't read certain values from the bsconfig. This is due to yargs providing default values, and since the program builder loads bsconfig, and then overrides with values from the cli, that means the cli would always override values.

### LINKED ISSUE #168: retain-staging-folder cli argument is ignored
The CLI seems to ignore the `retain-staging-folder` argument. It's always being deleted.

TwitchBronBron:
Fixes #167 - bugs in cli that wouldn't properly read bsconfig values.
Fixes #168 - bug in cli that doesn't use `retain-staging-folder` argument properly.

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -37,7 +37,6 @@ let options = yargs
     .argv;
 
 let builder = new ProgramBuilder();
-console.log(options);
 builder.run(<any>options).then(() => {
     //if this is a single run (i.e. not watch mode) and there are error diagnostics, return an error code
     const hasError = !!builder.getDiagnostics().find(x => x.severity === DiagnosticSeverity.Error);
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -6,36 +6,38 @@ import { DiagnosticSeverity } from 'vscode-languageserver';
 let options = yargs
     .usage('$0', 'BrighterScript, a superset of Roku\'s BrightScript language')
     .help('help', 'View help information about this tool.')
-    .option('create-package', { type: 'boolean', default: true, description: 'Creates a zip package. This setting is ignored when deploy is enabled.' })
+    .option('create-package', { type: 'boolean', defaultDescription: 'true', description: 'Creates a zip package. This setting is ignored when deploy is enabled.' })
     .option('cwd', { type: 'string', description: 'Override the current working directory.' })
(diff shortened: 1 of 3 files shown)


TwitchBronBron:
@elsassph can you give this a spin and make sure it doesn't break anything?

src/cli.ts
```diff
@@ -6,36 +6,38 @@ import { DiagnosticSeverity } from 'vscode-languageserver';
...
+    .option('username', { type: 'string', defaultDescription: '"rokudev"', description: 'The username for deploying to a Roku.' })
     .option('source-root', { type: 'string', description: 'Override the root directory path where debugger should locate the source files. The location will be embedded in the source map to help debuggers locate the original source files. This only applies to files found within rootDir. This is useful when you want to preprocess files before passing them to BrighterScript, and want a debugger to open the original files.' })
-    .option('watch', { type: 'boolean', default: false, description: 'Watch input files.' })
+    .option('watch', { type: 'boolean', defaultDescription: 'false', description: 'Watch input files.' })
     .strict()
     .check(argv => {
-        let d = argv.diagnosticLevel;
-        if (d !== 'error' && d !== 'warn' && d !== 'hint' && d !== 'info') {
-            throw new Error(`Invalid diagnostic level "${d}". Value can be "error", "warn", "hint", "info".`);
+        const diagnosticLevel = argv.diagnosticLevel as string;
```
elsassph, line 30: Happy to see `const` here - I thought I'd stay on the all `let` you had ;)

src/cli.ts
```diff
@@ -6,36 +6,38 @@ import { DiagnosticSeverity } from 'vscode-languageserver';
...
+        //if we have the diagnostic level and it's not a known value, then fail
+        if (diagnosticLevel && ['error', 'warn', 'hint', 'info'].includes(diagnosticLevel) === false) {
+            throw new Error(`Invalid diagnostic level "${diagnosticLevel}". Value can be "error", "warn", "hint", "info".`);
         }
         return true;
     })
     .argv;
 
 let builder = new ProgramBuilder();
+console.log(options);
```
elsassph, line 40: Probably not needed

src/cli.ts
TwitchBronBron: Yeah...you convinced me. I've started using const other places too. Eventually I'll get around to converting most of the `let`s to `const`.

PR #170, opened 2020-08-07
Feature: plugins API

elsassph:
First phase:

- API for plugins (builder lifecycle, transforms and diagnostics)
- helpers to reflect/explore the AST (walk/edit statements)

Generated readme doc:
https://github.com/rokucommunity/brighterscript/blob/f4be04ce9040639ac52a500f20d9d4084eae691e/docs/plugins.md


src/interfaces.ts
```diff
@@ -162,3 +162,19 @@ export interface CommentFlag {
     affectedRange: Range;
     codes: number[] | null;
 }
+
+export type FileTransformPlugin = (file: BrsFile, mode: ParseMode) => void;
+
+export type FileDiagnosticPlugin = (file: BrsFile, mode: ParseMode) => BsDiagnostic[];
+
+export type XmlDiagnosticPlugin = (file: XmlFile) => BsDiagnostic[];
```
georgejecook, line 170: oh my.. I could cry right now. this is going to be awesome!

src/parser/ASTUtils.ts
```diff
@@ -0,0 +1,222 @@
...
+    return {
+        kind: kind,
+        text: text || kind.toString(),
+        isReserved: !text || text === kind.toString(),
+        range: createRange(pos),
+        literal: literal
+    };
+}
+
+// Statements reflection
```
georgejecook, line 51: personally; in place of all these functions, I'd rather see an enum on Statement, like statementType, that the statement is responsible for setting. Better encapsulate imo.

Main reason though is working with this will be much nicer with enums than a bunch of if statements.
so if I was doing something I can do
```
switch (statement.type) {
  case StatementType.block:
  case StatementType.exitForStatement:
... etc
```

src/parser/ASTUtils.ts
```diff
@@ -0,0 +1,222 @@
...
+export function isNamespaceStatement(statement: Statement): statement is NamespaceStatement {
+    return statement?.constructor?.name === 'NamespaceStatement';
+}
+export function isImportStatement(statement: Statement): statement is ImportStatement {
+    return statement?.constructor?.name === 'ImportStatement';
+}
+
+// Expressions reflection
+
+export function isBinaryExpression(expression: Expression | Token): expression is BinaryExpression {
```
georgejecook, line 128: I have the same opinion re: enums for statements

src/parser/ASTUtils.ts
elsassph: This is the Typescript way to narrow the type of objects - it's very convenient in practice and is the API you get in Typescript compiler plugins itself. Also currently anyway brighterscript statements and expressions don't have a `type` property.

georgejecook: that was my point; I thought we should add that property on the expressions and statements. If this is more idiomatic, then it makes sense to follow the idioms: I'm not a ts developer, after all :)

bsconfig.schema.json
```diff
@@ -166,6 +166,18 @@
                 ]
             }
         },
+        "transforms": {
```
TwitchBronBron, line 169: We should rename the term `transformPlugin` to just `plugins`. I know this started out as just transforms, but it has already changed into both AST handling _and_ file validation _and_ scope validation. I envision there will be more plugin points in the future, so might as well keep the keyword generic.

elsassph: Done

src/Program.ts
```diff
@@ -47,6 +49,20 @@ export class Program {
         });
     }
 
+    private loadTransformPlugins() {
+        this.transforms = this.options.transforms.reduce((acc, pathRef) => {
+            if (typeof pathRef === 'string') {
+                if (pathRef.startsWith('.')) {
+                    pathRef = path.resolve(pathRef);
```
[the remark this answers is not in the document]
elsassph: Changed to resolve from `rootDir`

bsconfig.schema.json
TwitchBronBron: @elsassph you missed the renaming `transforms` in this schema file.

elsassph: Fixed...

docs/plugins.md
```diff
@@ -0,0 +1,201 @@
...
+
+### Example diagnostic plugins
+
+```typescript
+// myDiagnosticPlugin.ts
+import { ProgramBuilder, Scope, BrsFile,  XmlFile, CallableContainerMap } from 'brighterscript';
+
+// entry point
+export function initPlugin(builder: ProgramBuilder) {
+    builder.on('program-created', program => {
```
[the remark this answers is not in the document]
elsassph: Well that was basically the initial API :D

I think it's harder to use/discover for plugin authors, and to scale to more events and objects, and I didnt want to introduce API changes otherwise.

I also don't like the way the initial source creation doesn't trigger an event though, but I couldn't think of a clean solution with the events as it happens in the constructor. Maybe it could be delayed to the first source file registration.

src/Program.ts
```diff
@@ -307,15 +348,23 @@ export class Program {
...
+                        fileContents.src,
+                        () => this.emit('file-parsed', brsFile)
+                    );
                 });
                 file = brsFile;
 
                 this.dependencyGraph.addOrReplace(brsFile.dependencyGraphKey, brsFile.ownScriptImports.map(x => x.pkgPath.toLowerCase()));
 
-
+                this.emit('file-validated', brsFile);
```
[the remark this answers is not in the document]
elsassph: That's exactly the code I had initially. As you see, such a generic function is tricky to type properly - the `any[]` is quite painful to see ;)

src/parser/ASTUtils.ts
```diff
@@ -0,0 +1,444 @@
...
+        namespaceName
+    );
+}
+
+// Statements reflection
+
+export function isBody(statement: Statement): statement is Body {
+    return statement?.constructor?.name === 'Body';
+}
+export function isAssignmentStatement(statement: Statement): statement is AssignmentStatement {
```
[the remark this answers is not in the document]
elsassph: Using `instanceof` is safe within a single codebase. However a plugin might import a different version of brighterscript and try comparing different actual class objects.

That's why if you look at the Typescript plugin system it is designed with those `isSomeType` functions to narrow the type.

src/parser/Statement.ts
```diff
@@ -93,10 +101,25 @@ export class AssignmentStatement implements Statement {
     }
 }
 
+// note: order of values should not change
+export enum BlockOf {
```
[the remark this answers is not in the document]
elsassph: Statement visiting, especially using the object map pattern, is stateless: when you get a block you don't know the parent.

Maybe I should revisit the API to provide the parent statement instead.

docs/plugins.md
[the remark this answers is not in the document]
elsassph: What about "bubbling" the events maybe? E.g. what if you could listen to all the events on the builder instance? And I'll figure something clean enough to have the source created event.

src/parser/ASTUtils.ts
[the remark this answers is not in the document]
elsassph: Also you can refactor the internals to use a `type` property if you want to optimise the the lookup without breaking the API.

docs/plugins.md
```diff
@@ -0,0 +1,201 @@
...
+
+// Scope events:
+interface ScopeEventsMap {
+    'invalidated': () => void;
+    'validate-start': ValidateHandler;
+    'validate-end': ValidateHandler;
+}
+
+// related types:
+interface SourceDestEntry {
```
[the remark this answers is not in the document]
elsassph: Oh nice!

src/Program.ts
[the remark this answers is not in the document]
TwitchBronBron: To avoid using `any[]` we can overload the function definition to deal with that. 

```
private emitPluginEvent(eventName: 'first-event-type', data: string);
private emitPluginEvent(eventName: 'second-event-type', data: {srcDestEntry: SourceDestEntry, file: File});
private emitPluginEvent(eventName: 'third-event-type', data: boolean);
private emitPluginEvent(eventName: string, data: any[]) {
```

src/parser/ASTUtils.ts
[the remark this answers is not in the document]
TwitchBronBron: Ok, that makes sense. Thanks for explaining it. 

I feel like I'd rather remove these functions and instead add a `type` property to each of the statements and expressions. We can create a string enum with all the various types, and each statement can set its own type in its own constructor. 
```typescript
export enum StatementType{
    print = "print",
   ...
}
```

Then the plugin authors can write code like:
```typescript
if (statement.type === StatementType.print) { ...
```
or
```typescript
if (statement.type === "print"){ ...

File: docs/plugins.md (hunk 6 above)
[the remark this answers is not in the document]
Comment (TwitchBronBron) [reply docs/plugins.md]: Bubbling the events would work, and does simplify things a bit. However, why is my suggestion of a default export harder for plugin authors to scale? If we exported an interface like this:
```typescript
export interface Plugin {
    'program-created'?: (program: Program) => void;
    'scope-created'?: (scope: Scope) => void;
    'scope-validate-start'?: (scope: Scope) => void;
}
```

Then plugin authors could write their plugins like this: 
```typescript
// myDiagnosticPlugin.ts
export default {
    'program-created': program => { },
    'scope-created': scope => { },
    'scope-validate-start': scope => { }
} as Plugin;
```
The type casting at the end would give the plugin developers full type checking and discoverability for all plugin features. We could use the same `Plugin` interface internally to track all of the plugins, so as we add new plugin features, we would simply update the interface and both our internal code and the plugin api get the changes together. 

Honestly, I think this approach provides _easier_ discoverability for plugin features than events scattered across the various BrighterScript internals (but bubbling the events would solve that).

Another benefit to my proposed approach is the plugins become easier to test / mock, since they're just plain functions. Not a big deal, but definitely something to consider.

src/parser/ASTUtils.ts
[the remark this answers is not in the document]
elsassph: Testing on the type like that doesn't narrow the type of the statement/expression. You can "also" add type and simplify these functions.

src/Program.ts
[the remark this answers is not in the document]
elsassph: What do you think of the API now though? You can listen to all the events directly on the ProgramBuilder.

src/parser/Statement.ts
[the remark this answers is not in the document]
elsassph: Fixed

src/parser/ASTUtils.ts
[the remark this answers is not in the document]
TwitchBronBron: *OH*, I understand now. You were looking to leverage the typescript type narrowing. Yeah, I agree then, these are definitely useful. 

Yes, let's also add the "type" property to the objects so we can simplify these functions and make it easier for non-typescript developers. Honestly you could probably just set the `type` property in each statement/expression's `constructor` function to `this.type = this.constructor.name`.

docs/plugins.md
```diff
@@ -0,0 +1,174 @@
...
+
+`Scope` is a logical group of files; for instance `source` files, or each component's individual scope (XML definition, related components and imported scripts). A file can be included in multiple scopes.
+
+`BrsFile` is a `.brs` or `.bs` file; its AST can be modified by a plugin.
+
+`XmlFile` is a component; it doesn't have a proper AST currently but it can be "patched".
+
+```typescript
+export interface CompilerPlugin {
+    getName?: () => string;
```
TwitchBronBron, line 53: @elsassph is there a reason this is a function rather than just a string?

docs/plugins.md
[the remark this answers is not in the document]
TwitchBronBron: Did you get a chance to change the `FileObj` references in favor of this one? It doesn't look like it yet.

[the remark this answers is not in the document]
elsassph: Just forgot to remove it in the doc

docs/plugins.md
elsassph: I've gone back and forth with it. Prefer a string?

TwitchBronBron: Yeah, I think I do prefer a string.

elsassph: Ok done

src/PluginInterface.ts
```diff
@@ -0,0 +1,44 @@
+import { CompilerPlugin } from './interfaces';
+import { Logger } from './Logger';
+
+// inspiration: https://github.com/andywer/typed-emitter/blob/master/index.d.ts
+
+export type Arguments<T> = [T] extends [(...args: infer U) => any]
+    ? U
+    : [T] extends [void] ? [] : [T];
+
+export default class PluginInterface<T extends CompilerPlugin = CompilerPlugin> {
```
TwitchBronBron, line 10: How about we call this `Plugin`, since it's actually a class instance and _not_ an interface.

elsassph: Not necessarily the best name but ot's not the plugin, it's the interface (not in an OOP way) to the plugins that the compiler uses to emit events.

src/interfaces.ts
```diff
@@ -162,3 +167,27 @@ export interface CommentFlag {
     affectedRange: Range;
     codes: number[] | null;
 }
+
+type ValidateHandler = (scope: Scope, files: (BrsFile | XmlFile)[], callables: CallableContainerMap) => void;
+
+export interface CompilerPlugin {
```
TwitchBronBron, line 173: I'd like to standardize the names so we're using the same tense for them all. Here's what I'm thinking:

- beforePrepublish
- beforePublish
- afterPublish
- afterProgramCreate
- beforeProgramValidate
- afterProgramValidate
- beforeScopeCreate
- afterScopeDispose
- beforeScopeValidate
- afterScopeValidate
- afterFileSourceLoad
- afterFileParse
- beforeFileValidate
- afterComponentSourceLoad
- afterComponentParse
- afterComponentValidate
- beforeTranspile
- afterTranspile

src/interfaces.ts
```diff
@@ -162,3 +167,27 @@ export interface CommentFlag {
...
+    programValidateEnd?: (program: Program) => void;
+    scopeCreated?: (scope: Scope) => void;
+    scopeDisposed?: (scope: Scope) => void;
+    scopeValidateStart?: ValidateHandler;
+    scopeValidateEnd?: ValidateHandler;
+    fileSourceLoaded?: (source: SourceInfo) => void;
+    fileParsed?: (file: BrsFile) => void;
+    fileValidation?: (file: BrsFile) => void;
+    componentSourceLoaded?: (source: SourceInfo) => void;
+    componentParsed?: (file: XmlFile) => void;
```
TwitchBronBron, line 189: I don't think we need the `componentSourceLoaded` and `componentParsed` events. The component files should emit the `fileSourceLoaded` and `fileParsed` events instead. We may add additional file types in the future, and the plugin author can detect the type of file in their event handler.

src/interfaces.ts
```diff
@@ -162,3 +167,27 @@ export interface CommentFlag {
...
 }
+
+type ValidateHandler = (scope: Scope, files: (BrsFile | XmlFile)[], callables: CallableContainerMap) => void;
+
+export interface CompilerPlugin {
+    name: string;
+    beforePrepublish?: (files: FileObj[]) => void;
+    beforePublish?: (files: FileObj[]) => void;
+    afterPublish?: (files: FileObj[]) => void;
+    programCreated?: (program: Program) => void;
```
TwitchBronBron, line 178: I'd like ensure there are `before` and `after` for every single events. Plugin authors won't necessarily understand the full flow of the compiler, and they shouldn't need to know it. For example, there's nothing telling plugin authors that `beforePublish` happens after `beforePrepublish`. They would expect to see a `beforePrepublish` and `afterPrepublish`. This also better insulates our plugin system from unexpected issues if we were to add some additional step between `beforePrepublish` and `beforePublish`. Here are the events I think we are missing:

- afterPrepublish
- beforeProgramCreate
- afterScopeCreate
- beforeScopeDispose
- beforeFileSourceLoad
- beforeFileParse
- afterFileValidate

testProjects/plugins/removePrint.js
```diff
@@ -0,0 +1,19 @@
+"use strict";
```
TwitchBronBron, line 1: This file should be excluded from the project since it's transpiled from `removePrint.ts`

TwitchBronBron (changes requested):
Hey, I have requested several more changes, mostly centered around standardizing the event names.

src/Program.ts
```diff
@@ -325,15 +359,22 @@ export class Program {
                 let xmlFile = new XmlFile(pathAbsolute, pkgPath, this);
                 //add the file to the program
                 this.files[pathAbsolute] = xmlFile;
-                let fileContents = await getFileContents();
-                await xmlFile.parse(fileContents);
+                let fileContents = {
```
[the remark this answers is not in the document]
elsassph: This is to allow a plugin to modify the source before parsing.

src/Program.ts
```diff
@@ -357,6 +397,17 @@ export class Program {
         return file;
     }
 
+    /**
+     * Ensure source scope is created.
+     * Note: automatically called internally, and no-op is it exists already.
```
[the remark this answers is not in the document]
elsassph: Using it in the tests because a lot of tests expect source scope to exist.

src/ProgramBuilder.ts
```diff
@@ -24,14 +25,24 @@ export class ProgramBuilder {
     private watcher: Watcher;
     public program: Program;
     public logger = new Logger();
+    public plugins: PluginInterface;
+    private pluginModules = [] as CompilerPlugin[];
```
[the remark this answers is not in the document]
elsassph: Hmm yes, I used to defer plugins intialisation but now that's not needed.

src/interfaces.ts
elsassph: I like it - actually making `afterFileSourceLoad` -> `beforeFileParse`

src/Program.ts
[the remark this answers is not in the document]
TwitchBronBron: oh, ok. nice!

src/interfaces.ts
elsassph: Ok I tweaked a the events.

TwitchBronBron: Not sure I would say it's "really" wasteful, as these events should be essentially no-ops if no plugins are subscribing to them. However, as long as our documentation clearly outlines the execution order of the events, I can be ok with omitting `beforeFileValidate`.

src/Program.ts
```diff
@@ -696,30 +753,36 @@ export class Program {
...
+                file: file,
+                outputPath: outputPath
+            };
+        });
 
-            let result = file.transpile();
+        this.plugins.emit('beforeTranspile', entries);
+
+        const promises = entries.map(async ({ file, outputPath }) => {
+            const result = file.transpile();
```
TwitchBronBron, line 775: There could be value in emitting a `beforeFileTranspile` and `afterFileTranspile` event in this function. It removes the need for plugin authors to walk every file themselves.

src/interfaces.ts
```diff
@@ -162,3 +168,26 @@ export interface CommentFlag {
...
+    name: string;
+    beforeProgramCreate?: (builder: ProgramBuilder) => void;
+    beforePrepublish?: (builder: ProgramBuilder, files: FileObj[]) => void;
+    afterPrepublish?: (builder: ProgramBuilder, files: FileObj[]) => void;
+    beforePublish?: (builder: ProgramBuilder, files: FileObj[]) => void;
+    afterPublish?: (builder: ProgramBuilder, files: FileObj[]) => void;
+    afterProgramCreate?: (program: Program) => void;
+    beforeProgramValidate?: (program: Program) => void;
+    afterProgramValidate?: (program: Program) => void;
+    beforeTranspile?: (entries: TranspileObj[]) => void;
```
TwitchBronBron, line 184: can we rename this to `beforeProgramTranspile` and `afterProgramTranspile`?

TwitchBronBron (commented):
Added a few comments about event names.

src/Program.ts
elsassph: done

src/interfaces.ts
elsassph: done

src/Program.ts
```diff
@@ -303,20 +328,29 @@ export class Program {
...
+                    pathAbsolute: pathAbsolute,
+                    source: await getFileContents()
+                };
+                this.plugins.emit('beforeFileParse', fileContents);
+
                 this.logger.time(LogLevel.info, ['parse', chalk.green(pathAbsolute)], () => {
-                    brsFile.parse(fileContents);
+                    brsFile.parse(
+                        fileContents.source,
+                        () => this.plugins.emit('afterFileParse', brsFile)
```
TwitchBronBron, line 346: `BrsFile` already has a reference to `Program`, so I don't think we need to pass in a callback function for parse, because the `BrsFile` can just call `this.program.plugins.emit('afterFileParse', this)`.

src/Program.ts
```diff
@@ -326,15 +360,21 @@ export class Program {
...
-                let fileContents = await getFileContents();
-                await xmlFile.parse(fileContents);
+                let fileContents: SourceObj = {
+                    pathAbsolute: pathAbsolute,
+                    source: await getFileContents()
+                };
+                this.plugins.emit('beforeFileParse', fileContents);
+                await xmlFile.parse(
+                    fileContents.source,
+                    () => this.plugins.emit('afterFileParse', xmlFile)
```
TwitchBronBron, line 370: `XmlFile` already has a reference to `Program`, so I don't think we need to pass in a callback function for parse, because the `XmlFile` can just call `this.program.plugins.emit('afterFileParse', this)`.

src/astUtils/reflection.ts
```diff
@@ -0,0 +1,193 @@
...
+}
+
+export function isXmlFile(file: (BrsFile | XmlFile)): file is XmlFile {
+    return file.extension === '.xml';
+}
+
+// Statements reflection
+
+export function isBody(statement: Statement): statement is Body {
+    return statement?.constructor?.name === 'Body';
```
TwitchBronBron, line 22: I thought we were going to add a `type` property to statements and expressions that contained this info so we didn't have to use `.constructor.name`?

src/astUtils/reflection.ts
```diff
@@ -0,0 +1,193 @@
...
+export function isDottedSetStatement(statement: Statement): statement is DottedSetStatement {
+    return statement?.constructor?.name === 'DottedSetStatement';
+}
+export function isIndexedSetStatement(statement: Statement): statement is IndexedSetStatement {
+    return statement?.constructor?.name === 'IndexedSetStatement';
+}
+export function isLibraryStatement(statement: Statement): statement is LibraryStatement {
+    return statement?.constructor?.name === 'LibraryStatement';
+}
+export function isNamespaceStatement(statement: Statement): statement is NamespaceStatement {
```
TwitchBronBron, line 87: Looks like we're missing `isClassStatement`. Are there any other statements not represented here?

src/astUtils/reflection.ts
elsassph: I would suggest doing that as part of a separate task - it will be transparent from an API perspective. Do you want me to do it as part of this PR?

src/astUtils/reflection.ts
elsassph: Eh I took all the statements in `Statements.ts` and forgot the class statements which is in its own file :)

src/Program.ts
elsassph: Good point - that's not needed anymore.

src/astUtils/reflection.ts
TwitchBronBron: no, that's a good point. we can do it later.

src/astUtils/visitors.ts
```diff
@@ -0,0 +1,169 @@
+import { CancellationToken } from 'vscode-languageserver';
+import { Statement, Body, AssignmentStatement, Block, ExpressionStatement, CommentStatement, ExitForStatement, ExitWhileStatement, FunctionStatement, IfStatement, IncrementStatement, PrintStatement, GotoStatement, LabelStatement, ReturnStatement, EndStatement, StopStatement, ForStatement, ForEachStatement, WhileStatement, DottedSetStatement, IndexedSetStatement, LibraryStatement, NamespaceStatement, ImportStatement } from '../parser/Statement';
+import { Expression } from '../parser/Expression';
+import { isExpression, isBlock, isIfStatement } from './reflection';
+
+
+/**
+ * Create a filtered visitor for use with `editStatements`
+ */
+export function createStatementVisitor(
```
TwitchBronBron, line 10: Are we missing class statement here too? What about class fields/methods?

src/interfaces.ts
TwitchBronBron: You updated the docs to show the order of the events, that's good enough. :)

src/astUtils/visitors.ts
elsassph: This visitor only visits blocks - I don't think you can have a class in a Block?
Visiting classes/fields for now is meant to be visited using `file.parser.classStatements` collection.

TwitchBronBron: Ok, thanks for explaining.

PR #172, opened 2020-08-10
Support optional `project` and `extends` props

### LINKED ISSUE #171: Optional config file paths
There should be a way to specify that a config file (via `extends` or `project` or `--project`, etc...) is *optional*. This would allow more flexibility when sharing a project across teams.

TwitchBronBron:
Adds ability to specify that a config path (via `extends` or `project`) is optional, so the parser should not fail if the file is not found.
Fixes #171


TwitchBronBron:
I need to add some descriptions to `project` and `extends` in the interface and the schema yet.

PR #178, opened 2020-08-25
Fix typo for deprecate

TwitchBronBron:
We incorrectly use the word "depricate" several places when it should be with an `e` "deprecate". This PR fixes that issue.


TwitchBronBron:
Thanks @triwav for making me aware of this.

PR #181, opened 2020-08-28
Upgrade BS1104 and BS1105 to DiagnosticSeverity.Error

TwitchBronBron:
BS1104 and BS1105 are fairly critical issues and should be marked as errors instead of warnings.

Also, the wording is a bit off, so this fixes that.


TwitchBronBron:
@elsassph could you give this a quick review?

elsassph (approved):

PR #186, opened 2020-09-03
Detect and flag duplicate component names.

### LINKED ISSUE #185: Prevent Multiple Components With Same Name
Thought this was reported before but I guess not. Basically you can have multiple components with the same name. When your code gets run the Roku will choose one of the two components. Trying to diagnosis this can be a real time waster and should be easy enough to address.

TwitchBronBron:
Detect when there are multiple components with the same name, and flag each occurrence as an error.

Resolves #185

### DIFF
--- a/src/Program.spec.ts
+++ b/src/Program.spec.ts
@@ -195,7 +195,7 @@ describe('Program', () => {
             `);
             await program.addOrReplaceFile({ src: `${rootDir}/components/component2.xml`, dest: 'components/component2.xml' }, `
                 <?xml version="1.0" encoding="utf-8" ?>
-                <component name="Component1" extends="Scene" extends="Scene">
+                <component name="Component1" extends="Scene">
                 </component>
             `);
             await program.validate();
--- a/src/Program.spec.ts
+++ b/src/Program.spec.ts
@@ -1,7 +1,7 @@
 import { assert, expect } from 'chai';
 import * as pick from 'object.pick';
 import * as sinonImport from 'sinon';
-import { CompletionItemKind, Position, Range, DiagnosticSeverity } from 'vscode-languageserver';
+import { CompletionItemKind, Position, Range, DiagnosticSeverity, Location } from 'vscode-languageserver';
 import * as fsExtra from 'fs-extra';
 import { DiagnosticMessages } from './DiagnosticMessages';
 import { BrsFile } from './files/BrsFile';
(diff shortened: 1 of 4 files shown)


src/Program.spec.ts
```diff
@@ -185,6 +186,46 @@ describe('Program', () => {
...
+        it('catches duplicate XML component names', async () => {
+            //add 2 components which both reference the same errored file
+            await program.addOrReplaceFile({ src: `${rootDir}/components/component1.xml`, dest: 'components/component1.xml' }, `
+                <?xml version="1.0" encoding="utf-8" ?>
+                <component name="Component1" extends="Scene">
+                </component>
+            `);
+            await program.addOrReplaceFile({ src: `${rootDir}/components/component2.xml`, dest: 'components/component2.xml' }, `
+                <?xml version="1.0" encoding="utf-8" ?>
+                <component name="Component1" extends="Scene" extends="Scene">
```
triwav, line 198: Not sure if that's even valid xml with the double extends

src/files/XmlFile.ts
```diff
@@ -176,6 +184,16 @@ export class XmlFile {
                             Position.create(lineIndex, idx),
                             Position.create(lineIndex, idx + 10)
                         );
+                        //calculate the rane of the component's name (if it exists)
```
triwav, line 187: assume you meant name not rane?

elsassph: More like range

src/Program.ts
```diff
@@ -466,10 +468,52 @@ export class Program {
...
+                if (!map[lowerName]) {
+                    map[lowerName] = [];
+                }
+                map[lowerName].push(file);
+            }
+            return map;
+        }, {} as { [lowerComponentName: string]: XmlFile[] });
+
+        for (let componentName in componentsByName) {
+            let xmlFiles = componentsByName[componentName];
```
elsassph, line 494: Const ;)

src/files/XmlFile.ts
```diff
@@ -176,6 +184,16 @@ export class XmlFile {
...
+                        let match = /(.*?name\s*=\s*(?:'|"))(.*?)('|")/.exec(this.lines[lineIndex]);
```
elsassph, line 188: Const

src/Program.spec.ts
TwitchBronBron: oops. copy-paste error.

src/files/XmlFile.ts
TwitchBronBron: yeah, range. Good catch

src/files/XmlFile.ts
TwitchBronBron: yeah, might as well. This is a pretty big codebase, so haven't flipped the eslint switch yet to enforce this. I think I'll do this once we get your plugin PR merged, and the ternary PR merged.

TwitchBronBron:
@elsassph @triwav I've addressed all of your comments now.

elsassph (approved):

PR #187, opened 2020-09-23
Plugin enhancements

TwitchBronBron:
Good thing we released the plugin system in alpha! :D 

Moves the majority of the walking functionality onto the AST nodes (`Statement`, `Expression`) to simplify usage for plugin authors and match the object-oriented design of the rest of the project.

Currently if plugin authors wish to modify the AST, they have to also various items onto parser reference lists (like `parser.functionExpressions`, `parser.importStatements`, etc...). This is not the most intuitive approach, and can be quite tedious for plugin authors. To simplify this process, a plugin author can now simply call `BrsFile.parser.clearReferences()` if they modify AST, and the parser will automatically recalculate all of the necessary references. 

I realize this is not the most efficient way to handle AST manipulation, but since this `clearReferences()` call is optional, there's not a huge downside in my opinion. As we design new and more efficient ways to handle the AST manipulations, we can drive plugin authors to those better ways. And let the "free market" decide which plugins win, because nobody wants a slow plugin. 

The `parser.findReferences()` call (which does a full AST walk) can eventually be rolled into the future branch tracking code that @elsassph is working on so there is only a single full-walk of the AST. Also, many of the items in `references` could probably be removed once we add validation into the full walk. One step at a time. :D 

**NOTE:** This is still a work in progress, but I wanted to get feedback earlier rather than later. 

Notable changes:
 - move parser ast node lists (such as `parser.functionExpressions`, `parser.importStatements`, etc...) Into to `parser.references`
 - add option for full AST walk to re-collect `parser.references`. only necessary if AST has changed (indicated by calling `parser.clearReferences()`
 - move AST walking functionality out of standalone functions and onto their corresponding AST nodes. This best matches the object-oriented design of the rest of the project, and makes walking easier for plugin authors to utilize.
 - AST walking no longer emits a visitor event for itself. For example, if you call `Block.walk()`, you won't get a visitor event for that instance of `Block`. But that's ok, because you already _have_ the block reference since that's what you called `walk()` on in the first place. 
 - Create `WalkOptions` to specify various AST walking functionality (ability to walk statements only, walk expressions, step into sub functions). 
 - Rename `createStatementVisitor` to `createVisitor` and added `Expression` keys. 
 - Removed several `Array.forEach` loops in favor of the more efficient standard for loop. 
 - Renamed a few expressions classes to end with the literal word `Expression` to be consistent with rest of expression classes.
 - Refactored `reflection.ts` methods to allow passing in any object, since we're not always sure if we are providing an Expression, or Statement, or some other object.
 - statements and expressions now extend base class instead of implementing interface
 - moved class statements into the main Statements.ts file to mitigate typescript circular dependencies issue related to them being in a separate file. 

**TODO:** Add a `WalkOptions` option to restrict the walk to only visiting own expressions (which is how the original `Expression.walk()` function worked).


src/astUtils/reflection.ts
```diff
@@ -19,160 +18,166 @@ export function isXmlFile(file: (BrsFile | XmlFile)): file is XmlFile {
 
 // Statements reflection
 
-export function isBody(statement: Statement): statement is Body {
+export function isBody(statement: any): statement is Body {
```
TwitchBronBron, line 21: Changed these parameters to the `any` type, since we might pass other object types into them. It _could_ be `Statement | Expression`, but since we're checking for `.constructor.name`, might as well just let developers pass in any object.

src/astUtils/visitors.ts
```diff
@@ -33,137 +78,51 @@ export function createStatementVisitor(
...
-    return (statement: Statement, parent: Statement): Statement | void => {
+    return <WalkVisitor>((statement: Statement, parent: Statement): Statement | void => {
         return visitor[statement.constructor.name]?.(statement, parent);
-    };
-}
-
-/**
- * Create a statement -> expression visitor
- */
-export function createStatementExpressionsVisitor(
```
TwitchBronBron, line 46: This functionality has been moved into the AST nodes (`Expression` and `Statement`). 

**TODO: ** Add `WalkOptions` flag for _only_ walking own expressions (right now it'll still walk sub statements and expressions).

src/astUtils/visitors.spec.ts
```diff
@@ -119,6 +120,7 @@ describe('astUtils visitors', () => {
                 'Block:2',                // while block
                 'PrintStatement:3',       // print 6
                 'ForStatement:1',         // for a = 1 to 10
+                'AssignmentStatement:2',  // a = 1
```
TwitchBronBron, line 123: @elsassph My walk function emits this statement. I think this is correct, but your test didn't include it.

src/parser/Expression.ts
```diff
@@ -41,11 +42,10 @@ export class BinaryExpression implements Expression {
         ];
     }
 
-    walk(visitor: ExpressionVisitor, parent?: Expression, cancel?: CancellationToken): void {
-        if (!cancel?.isCancellationRequested) {
-            visitor(this, parent);
-            this.left.walk(visitor, this, cancel);
-            this.right.walk(visitor, this, cancel);
+    walk(visitor: WalkVisitor, options?: WalkOptions) {
+        if (options?.walkExpressions) {
```
TwitchBronBron, line 46: I'm not sure we actually need this condition for these expressions. Generally the only reason you would call `expression.walk` is if you wanted to walk the expression. I put them there "just in case", but I could see an argument being made to exclude them to improve performance (ever so slightly).

src/parser/Parser.spec.ts
```diff
@@ -14,6 +14,31 @@ describe('parser', () => {
         });
     });
 
+    describe('findReferences', () => {
+        it('gets called if references are missing', () => {
```
TwitchBronBron, line 18: Here's a great example of how a plugin would request the parser to regenerate the references.

src/PluginInterface.ts
```diff
@@ -9,21 +9,21 @@ export type Arguments<T> = [T] extends [(...args: infer U) => any]
...
-    constructor(private plugins: CompilerPlugin[], private logger: Logger) {}
+    constructor(private plugins: CompilerPlugin[], private logger: Logger) { }
 
     /**
      * Call `event` on plugins
      */
     public emit<K extends keyof T & string>(event: K, ...args: Arguments<T[K]>) {
-        this.plugins.forEach((plugin) => {
-            if (plugin.hasOwnProperty(event)) {
+        for (let plugin of this.plugins) {
```
elsassph, line 18: I believe there is little performance difference nowadays between `forEach` and `for of`

src/Scope.ts
```diff
@@ -237,7 +237,7 @@ export class Scope {
         let namespaceLookup = {} as { [namespaceName: string]: NamespaceContainer };
         let files = this.getFiles();
         for (let file of files) {
-            for (let namespace of file.parser.namespaceStatements) {
+            for (let namespace of file.parser.references?.namespaceStatements ?? []) {
```
elsassph, line 240: Can't we assume `references` always exists?

src/files/BrsFile.ts
```diff
@@ -222,8 +223,8 @@ export class BrsFile {
         }
 
         let statements = [
-            ...this.parser.libraryStatements,
-            ...this.parser.importStatements
+            ...this.parser.references?.libraryStatements ?? [],
```
elsassph, line 226: Optional needed?

src/parser/ClassStatement.ts
```diff
@@ -327,6 +328,12 @@ export class ClassStatement implements Statement {
         result.push(`end function`);
         return result;
     }
+
+    walk(visitor: WalkVisitor, options?: WalkOptions) {
+        for (let i = 0; i < this.body.length; i++) {
```
elsassph, line 333: As I noted, `for i` isn't faster than `forEach` nowadays. Even slower in some benchmarks.

src/parser/Parser.ts
```diff
@@ -98,65 +100,58 @@ export class Parser {
...
+            this._references = this.findReferences();
+        }
+        return this._references;
+    }
 
-    /**
-     * All function statements defined in this file
-     */
-    public functionStatements = [] as FunctionStatement[];
+    private _references: References = {
```
elsassph, line 121: Looks like there is no reason to have all the optional chaining and `?? []`. Optional guard generates quite a lot of overhead.

src/parser/Parser.ts
```diff
@@ -98,65 +100,58 @@ export class Parser {
...
+        newExpressions: []
+    };
 
     /**
-     * All function expressions defined in this file
+     * Clears the references. This should be called anytime the AST has been manipulated.
      */
-    public functionExpressions = [] as FunctionExpression[];
+    clearReferences() {
+        this._references = undefined;
```
elsassph, line 136: Should reset to an object with empty arrays IMHO. Maybe even better should be an `invalidateReferences` so you know they need to be refreshed but you don't lose the data.

src/parser/Parser.ts
```diff
@@ -2375,6 +2369,54 @@ export class Parser {
         }
     }
 
+    /**
+     * References are found during the initial parse.
+     * However, sometimes plugins can modify the AST, requiring a full walk to re-compute all references.
+     * This does that walk.
+     */
+    findReferences() {
+        const references: References = {
```
elsassph, line 2378: Should be a `validateReferences` which would run only if needed.

src/parser/Statement.ts
```diff
@@ -466,6 +534,14 @@ export class PrintStatement implements Statement {
         }
         return result;
     }
+    walk(visitor: WalkVisitor, options?: WalkOptions) {
+        if (options?.walkExpressions) {
+            for (let i = 0; i < this.expressions.length; i++) {
+                //TODO the original walk code did an `isExpression()` check before walking. Do we need that?
```
elsassph, line 540: I think there were cases where there can be "tokens" in such a list instead of expressions.

elsassph (commented):
Looks good overall!

src/Scope.ts
TwitchBronBron: Yes, good point. Since the references object is created at construct time, and it's a getter which regenerates itself. I'll update these to assume it always exists.

src/files/BrsFile.ts
TwitchBronBron: Nope. It's guaranteed to exist, so no need for optional. Thanks.

src/parser/ClassStatement.ts
TwitchBronBron: `for i` is definitely faster than `forEach`. I just ran a few local benchmarks testing various loop implementations. See [this gist](https://gist.github.com/TwitchBronBron/16f3b20f76c66527169a61f5c2e1d0ce) for the test results. 

However, in this situation, we actually do need the loop index, because we pass it into the `walk` helper function.

src/parser/Parser.ts
TwitchBronBron: Agreed. I have removed all of the optional chaining stuff.

src/parser/Parser.ts
TwitchBronBron: I don't see any value in keeping the old references around since `Parser.references` is a getter that auto-regenerates `references` if they are cleared. If a plugin author has modified the AST, then none of these references can be considered valid anymore anyway.

src/parser/Parser.ts
TwitchBronBron: I will change the signature of this method to `private`. Nobody should be calling this except the `Parser.references` getter, which has the logic to conditionally run this method.

src/parser/Parser.ts
elsassph: Ok but still `invalidateReferences` would be more consistent.

src/parser/Parser.ts
elsassph: Makes sense now

src/util.ts
```diff
@@ -950,6 +950,42 @@ export class Util {
...
+            }
+        };
+    }
+
+    /**
+     * Create a `Range` from two `Position`s
+     */
+    public createRangeFromPositions(startPosition: Position, endPosition: Position): Range {
+        return {
+            start: startPosition,
```
elsassph: Is it safe to use the object without cloning?

TwitchBronBron: yeah I think so. At least in our project, we treat all of the Range and Position objects as read-only.

elsassph: Could that be documented, or could the util have a default-false `clone` option, so that if a plugin author uses the util he would be aware the objects aren't cloned. Otherwise it might be generally safer and not terribly more expensive to avoid using the objects as-is.

src/astUtils/reflection.ts
elsassph: I was bitten by this `any` when converting my code - it is definitely a risk of bugs IMHO to not narrow to `Statement | Expression`.

src/util.ts
TwitchBronBron: I don't see any valid use cases for messing with the token positions. However, you're right, it wouldn't be measurably different to just clone, so I'll change that.

src/astUtils/reflection.ts
TwitchBronBron: Sure. I guess `Statement`s and `Expression`s are the most common use case. I do see valid reasons for passing in any variable to determine whether it's a statement or an expression. However, in those cases the variable can be cast as `any`.

src/parser/Statement.ts
```diff
@@ -560,7 +560,7 @@ export class PrintStatement extends Statement {
         if (options.walkExpressions) {
             for (let i = 0; i < this.expressions.length; i++) {
                 //sometimes we have semicolon `Token`s in the expressions list (should probably fix that...), so only emit the actual expressions
-                if (isExpression(this.expressions[i])) {
+                if (isExpression(this.expressions[i] as any)) {
```
TwitchBronBron, line 563: Cast as `any` as a result of https://github.com/rokucommunity/brighterscript/pull/187#discussion_r496621290

src/astUtils/visitors.ts
```diff
@@ -1,14 +1,77 @@
...
+
+    //the object we're visiting
+    let element: any = keyParent[key];
+    if (!element) {
+        return;
+    }
+
+    //notify the visitor of this element
+    if (
+        (isExpression(element) && options.walkMode & WalkModeInternal.visitExpressions) ||
```
TwitchBronBron, line 42: @elsassph do you have any suggestions for a more efficient way to handle the walk/visit logic? Right now, we have to determine each element's type before we can know for certain that we should notify the visitor.

elsassph: Ouch.
This reminds me that `isExpresion` isn't reliable, now that you added `walk` functions to some statements.

Most efficient would be that all expressions and statements have a `walkMode` property, so you could just write:
```
if (element.walkMode & opens.walkMode) {...}
```

TwitchBronBron: Oh, I like that approach. And we can set that in the superclass constructor so it doesn't clutter the Statements and Expressions.

src/util.ts
TwitchBronBron: Done.

src/parser/Parser.ts
TwitchBronBron: done.

src/astUtils/reflection.ts
TwitchBronBron: done.

PR #188, opened 2020-09-23
Support for type definition files

TwitchBronBron:
This adds support for loading type definition files (`d.bs` extension). 

**Notable changes:**
 - load type definition files during file parse.
 - cache type files to reduce excess file system loading.
 - generate type definitions during transpile when `emitDefinitions` is true in the bsconfig


TwitchBronBron:
Forgot to mark this as "draft". It's not ready yet.

src/Program.ts
```diff
@@ -277,6 +280,27 @@ export class Program {
...
+     * Get the type definitions for the specified file.
+     * The results are cached for future performance boosts.
+     */
+    private async getTypeDefinitionsForFile(pathAbsolute: string) {
+        //you can only get type definitions for .brs files
+        if (!pathAbsolute.endsWith('.brs')) {
+            return;
+        }
+
+        let typeDefinitionPath = pathAbsolute.replace(/.brs$/, '.d.bs');
```
elsassph, line 293: Do you really need to do this lookup? When are the definitions concretely needed?
`.d.bs` files will be collected like other files in the Program so they will be in the files map.

`BrsFile`s could have a `get definitions` to dynamically look up for a definition file (collcted in Program)

TwitchBronBron: Currently within this PR in the `BrsFile.ts`, if we detect that there's a definition file for the `BrsFile`, we bypass parsing the `.brs` file. This should significantly reduce parse times for libraries that include `d.bs` files since we don't need to parse the whole `.brs` file, only its definitions. 

The main problem is, we aren't guaranteed _when_ the files will be passed into `Program`. There are three methods to adding files to the program, all of which will emit files when they find them, in no apparent order

 - The vscode file watcher while in language server mode (of which we have no control other than file matcher globs)
 - The glob search during a single cli run
 - the chokidar file watcher during a `--watch` cli run

This may lead to race conditions where the `.brs` files are handed to the program _before_ the `.d.bs` file.

If we don't detect the `d.bs` file early, then we will end up paying parse costs for every library file, only to ignore most of what we parsed once we detect the definitions file in your proposed `get definitions` getter.

elsassph: Ok if you want to avoid parsing it makes some sense.

Instead of doing this `getFileContents` though what about: 
if `.d.ts` file exits, and it's not in the file map, then automatically add and parse it, and link its `BrsFile` to the `.brs`'s `BrsFile`.

Also you may still want to open the library's `.brs` file in VSCode, then you still want the editor to understand the file. You will need then to parse the real file.

src/Program.ts
```diff
@@ -283,36 +283,38 @@ export class Program {
...
         //you can only get type definitions for .brs files
-        if (!pathAbsolute.endsWith('.brs')) {
+        if (!pathAbsolute.toLowerCase().endsWith('.brs')) {
             return;
         }
 
-        let typeDefinitionPath = pathAbsolute.replace(/.brs$/, '.d.bs');
-
+        let typedefPath = pathAbsolute.replace(/.brs$/i, '.d.bs');
+        let lowerTypedefPath = typedefPath.toLowerCase();
```
elsassph, line 293: Maybe `typedefKey` instead? You could use another algorithm here.

src/Program.ts
```diff
@@ -283,36 +283,38 @@ export class Program {
      * Get the type definitions for the specified file.
      * The results are cached for future performance boosts.
      */
-    private async getTypeDefinitionsForFile(pathAbsolute: string) {
+    private async getTypedefsForFile(pathAbsolute: string) {
```
elsassph, line 286: What if the file is a `d.bs` file already?

src/parser/Statement.ts
```diff
@@ -95,6 +99,20 @@ export class Body extends Statement {
         return result;
     }
 
+    getTypedef(state: TranspileState) {
```
elsassph, line 102: Why not having a "transpilation mode", even as a static flag somewhere, to control that instead of doubling half the transpilation logic with those `getTypedef` methods?

TwitchBronBron: I considered that, but most expressions and statements will not emit any typedef information. Having a separate function allows me to detect which items we should even walk in the first place (and avoid needless `transpile` recursing).  See 
`if ('getTypedef' in statement) {` a few lines below this comment.

Also, if most of the transpile function can be reused, I'm already calling it within the `getTypedef` method so we don't actually duplicate any logic.  (see [comment statement](https://github.com/rokucommunity/brighterscript/blob/4be26ae9e5691446d4bb81338e4490260338ff44/src/parser/Statement.ts#L278-L280)). Meanwhile, something like a  [NamespaceStatement](https://github.com/rokucommunity/brighterscript/blob/4be26ae9e5691446d4bb81338e4490260338ff44/src/parser/Statement.ts#L1093-L1110) doesn't actually transpile itself, so that's where having a separate function keeps the logic clean between the two concepts (`transpile` and `getTypedef`). Or ClassStatement, whose `transpile` function is extremely complex and doesn't have _any_ similarities with what its `typedef` would look like.

src/Program.ts
TwitchBronBron: That's not currently possible. In `Program.addOrReplaceFile`, there's an [if branch](https://github.com/rokucommunity/brighterscript/blob/4be26ae9e5691446d4bb81338e4490260338ff44/src/Program.ts#L366) that refreshes the `typedefCache` for `d.bs` files. So this function won't be called for d.bs files.

src/Program.ts
TwitchBronBron: Good suggestion. I'll do that.

src/Program.ts
TwitchBronBron: >Instead of doing this getFileContents (and parse the .d.bs twice) 

In the current logic, the `d.bs` file will only be parsed once (within its corresponding `.brs` file). Also, since most developers aren't actively editing `.brs` files that have associated `d.bs` files, the current design won't cause performance issues for normal development. 

>if .d.bs file exits, and it's not in the file map, then automatically add and parse it, and link its BrsFile to the .brs's BrsFile.

I like this plan and want to move that direction eventually. However, I think this will need to be a future enhancement. I am trying very hard to get basic type definition support sooner rather than later to correspond with the ropm release and the upcoming meetup and roku developer "converence". 

> Also you may still want to open the library's .brs file in VSCode, then you still want the editor to understand the file. You will need then to parse the real file.

Very valid use case, and something we definitely need to support at some point. In general, most developers will not be hand-writing `d.bs` files in the short-term (within the next 1-3 months). Especially not with `d.bs` files being brand new. So I think we'll address this use case at a later date. 

Most of the implementation details of `d.bs` files are internal, so we can change their functionality as the design evolves.

elsassph: I mean, aren't the `.d.bs` files part of the files you will collect anyway?

elsassph: If you add a library with typedefs in your project, brighterscript will collect all the brs and the bs files, which will include the d.bs one's.

src/Scope.ts
```diff
@@ -208,25 +207,34 @@ export class Scope {
         }
     }
 
+    public enumerateFiles(callback: (file: BscFile) => void) {
+        const files = this.getFiles();
+        for (const file of files) {
+            //skip files that have a typedef
+            if (file.extension === '.brs' && (file as BrsFile).hasTypedef) {
```
elsassph, line 214: Probably could simplify by having this `hasTypedef` both in XmlFile and BrsFile?

src/files/BrsFile.ts
```diff
@@ -41,20 +43,31 @@ export class BrsFile {
...
+        this.extension = extname(this.pkgPath).toLowerCase();
 
         //all BrighterScript files need to be transpiled
-        if (this.extension === '.bs') {
+        if (this.extension.endsWith('.bs')) {
             this.needsTranspiled = true;
         }
+        this.isTypedef = this.extension === '.d.bs';
+        if (!this.isTypedef) {
+            this.typedefKey = util.getTypedefPath(this.pathAbsolute);
```
elsassph, line 54: Feels a bit wasteful to calculate for every file - should it be undefined until it "has" a typedef?

src/files/BrsFile.ts
```diff
@@ -121,14 +134,93 @@ export class BrsFile {
...
+    /**
+     * The key to find the typedef file in the program's files map.
+     * A falsey value means this file is ineligable for a typedef
+     */
+    public typedefKey?: string;
+
+    /**
+     * If the file was given type definitions during parse
+     */
+    public get hasTypedef() {
```
elsassph, line 159: This feels like a hidden "relatively heavy" operation, and the `cache` approach doesn't feel right. I'd think a more explicit pattern for setting the flag would be preferable.

src/files/BrsFile.ts
```diff
@@ -121,14 +134,93 @@ export class BrsFile {
...
+     * A reference to the typedef file (if one exists)
+     */
+    public get typedefFile() {
+        if (this.typedefKey) {
+            return this.cache.getOrAdd('typedefFile', () => {
+                const file = this.program.getFileByPathAbsolute<BrsFile>(this.typedefKey);
+                return file;
+            });
+        }
+    }
```
elsassph, line 173: This type of code should have an explicit `return undefined` fallback.

src/Scope.ts
TwitchBronBron: Done.

src/files/BrsFile.ts
TwitchBronBron: done.

src/Program.ts
TwitchBronBron: This comment is no longer relevant...the typedefs are added as actual files into the program like other files.

src/files/BrsFile.ts
TwitchBronBron: We need the key for every file right away because we use that key to subscribe to typedef file changes (add/update/delete of typedefs).

src/files/BrsFile.ts
TwitchBronBron: Yeah, good point. I refactored my code accordingly.

TwitchBronBron: I thought I needed the evaluation of typedef to be lazy, but after some testing, I verified that the dependency graph events are emitted after the file is finalized, so it _is_ safe to simply resolve these at the time of the event. [Here](https://github.com/rokucommunity/brighterscript/pull/188/files#diff-abb7837cc5611725991e1566d0a4924f98e12e39cecd95dc7f9ccceb6ff658f9R200) and [here](https://github.com/rokucommunity/brighterscript/pull/188/files#diff-abb7837cc5611725991e1566d0a4924f98e12e39cecd95dc7f9ccceb6ff658f9R59) are the relevant locations where I resolve this information now.

src/files/BrsFile.ts
```diff
@@ -121,14 +138,91 @@ export class BrsFile {
...
+     * Also notify the dependency graph of our current dependencies so other dependents can be notified.
+     */
+    public attachDependencyGraph(dependencyGraph: DependencyGraph) {
+        if (this.unsubscribeFromDependencyGraph) {
+            this.unsubscribeFromDependencyGraph();
+        }
+
+        //anytime a dependency changes, clean up some cached values
+        this.unsubscribeFromDependencyGraph = this.program.dependencyGraph.onchange(this.dependencyGraphKey, () => {
+            this.logDebug('clear cache because dependency graph changed');
```
elsassph, line 194: Log message is unclear now that here isn't a cache

src/files/BrsFile.ts
```diff
@@ -121,14 +138,91 @@ export class BrsFile {
...
+
+    /**
+     * Was parsing skipped because the file has a typedef?
+     */
+    private wasParseSkipped = false;
+
     /**
      * Calculate the AST for this file
      * @param fileContents
+     * @param typedefContents - a string containing the d.bs file contents (type definitions)
```
elsassph, line 225: Param doesn't exist

src/files/BrsFile.ts
```diff
@@ -137,8 +231,16 @@ export class BrsFile {
...
+            //if we have a typedef file, skip parsing this file
+            if (this.hasTypedef) {
+                this.wasParseSkipped = true;
+                this.parseDeferred.resolve();
+                return;
+            }
+
             //tokenize the input file
             let lexer = this.program.logger.time(LogLevel.debug, ['lexer.lex', chalk.green(this.pathAbsolute)], () => {
+                //scan the d.bs file if present, otherwise the fileContents
```
elsassph, line 243: Still relevant comment?

src/files/XmlFile.ts
```diff
@@ -129,7 +136,11 @@ export class XmlFile {
      */
     public getAvailableScriptImports() {
         return this.cache.getOrAdd('allAvailableScriptImports', () => {
-            let allDependencies = this.getAllScriptImports();
+
+            let allDependencies = this.getAllDependencies()
+                //skip typedef files
+                .filter(x => extname(x) !== '.d.bs');
```
elsassph, line 142: Should they be replaced by `.brs` files?

src/util.ts
```diff
@@ -633,6 +632,20 @@ export class Util {
...
+     * Only `.brs` files can have typedef path, so return undefined for everything else
+     */
+    public getTypedefPath(brsSrcPath: string) {
+        const typedefPath = brsSrcPath
+            .replace(/\.brs$/i, '.d.bs')
+            .toLowerCase();
+
+        if (typedefPath.endsWith('.d.bs')) {
+            return typedefPath;
+        }
```
elsassph, line 646: `return undefined`

src/Program.ts
TwitchBronBron: resolving this because it has been implemented.

src/files/BrsFile.ts
TwitchBronBron: resolved.

src/files/BrsFile.ts
TwitchBronBron: nice catch.

src/files/BrsFile.ts
TwitchBronBron: nope! Nice catch.

src/util.ts
TwitchBronBron: resolved.

src/files/XmlFile.ts
TwitchBronBron: No, the .bs/.brs files are already there. For every .bs/.brs file, the XmlFile _also_ adds the corresponding `.d.bs` file during initialization (see [this](https://github.com/rokucommunity/brighterscript/pull/188/files#diff-c2462ef143ff214d1683501300a6cc2960cf54d7ba36e7606f7383c95021a240R430) new block of code)

TwitchBronBron:
I believe this is functional enough for an initial release. I'm sure we'll run into some quirks, but luckily enough, this is an opt-in feature so it shouldn't have too many adverse effects for teams who aren't opting in to this feature.

PR #189, opened 2020-09-27
Various performance tweaks

TwitchBronBron:
`vscode-languageserver`'s `Range.create()` seems to be significantly slower than creating the ranges manually. 
image

So, by switching the Lexer to use our own `createRange` function, we can improve its performance. With this change, scanning a 400 line file went from 746 ops/sec to 931 ops/sec
image

All benchmarks were performed against a 400 line brighterscript file. 

**Notable Changes:**

1. eliminate `Range.create()` in favor of `util.createRange()` and `util.createRangeFromPositions()`. 
    - ~25% increase in `Lexer` performance
    - ~12% increase in `Parser` performance
2. In `Lexer.ts`, don't return previous character in the `advance()` method.
    - ~3% increase in `Lexer` performance
3. In `Lexer.ts`, replace switch statement with a static map.
    - ~27% increase in `Lexer` performance

Overall, this has improved the performance of the the `Lexer+Parser` cycle by ~16% (from 300 ops/sec to 350 ops/sec) for my 400 line test file.

### DIFF
--- a/src/lexer/Lexer.ts
+++ b/src/lexer/Lexer.ts
@@ -1,3 +1,4 @@
+/* eslint-disable func-names */
 import { TokenKind, ReservedWords, Keywords } from './TokenKind';
 import { Token } from './Token';
 import { isAlpha, isDecimalDigit, isAlphaNumeric, isHexDigit } from './Characters';
@@ -90,7 +91,6 @@ export class Lexer {
         this.columnEnd = 0;
         this.tokens = [];
         this.diagnostics = [];
-
         while (!this.isAtEnd()) {
             this.scanToken();
         }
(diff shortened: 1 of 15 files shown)


src/lexer/Lexer.ts
```diff
@@ -122,218 +122,206 @@ export class Lexer {
         return this.current >= this.source.length;
     }
 
+    /**
+     * Map for looking up token functions based solely upon a single character
+     * Should be used in conjunction with `tokenKindMap`
+     */
+    private static tokenFunctionMap = {
```
TwitchBronBron, line 129: This isn't the prettiest approach in the world, but as a result, it yielded much higher throughput, and it's not terribly difficult to understand, so I think it's worth the tradeoff.

TwitchBronBron:
I'm sure there are more improvements we can make, bit these are a good start.

PR #190, opened 2020-09-28
Minor fixes to brs-plugins refactoring

elsassph:


src/astUtils/visitors.ts
```diff
@@ -26,34 +26,37 @@ export type WalkVisitor = <T = Statement | Expression>(stmtExpr: Statement | Exp
...
     //notify the visitor of this expression
-    const result = visitor(keyParent[key] as any, parent ?? keyParent as any);
+    const result = visitor(element, parent ?? keyParent as any);
 
-    //Replace the value on the parent if the visitor returned a Statement or Expression (this is how visitors can edit AST)
-    if (isExpression(result) || isStatement(result)) {
+    //replace the value on the parent if the visitor returned a Statement or Expression (this is how visitors can edit AST)
+    if (result && (isExpression(result) || isStatement(result))) {
         (keyParent as any)[key] = result;
+        element = result;
```
elsassph, line 47: Should we actually walk an element just created?

TwitchBronBron: Yeah, good point. Thanks for catching that.

src/parser/Parser.ts
```diff
@@ -384,9 +384,8 @@ export class Parser {
                 if (this.check(TokenKind.Function, TokenKind.Sub) || (this.check(TokenKind.Identifier, ...AllowedProperties) && this.checkNext(TokenKind.LeftParen))) {
                     let funcDeclaration = this.functionDeclaration(false);
 
-                    //remove this function from the lists because it's a class method
-                    this._references.functionExpressions.pop();
```
TwitchBronBron, line 388: Nice, I see you're no longer popping the functionExpressions list. This is because currently we still want to have a direct way to walk all class methods without having to walk through the class AST nodes directly.

TwitchBronBron:
Thanks for the fixes!

PR #191, opened 2020-09-29
Implement onWorkspaceSymbol, onSignatureHelp, onDocumentSymbol, onReferences, improve onDefinition

triwav:
This PR covers moving onSignatureHelp, onWorkspaceSymbol, onDocumentSymbol support in. It also adds a much larger support for onDefinition as well.


src/LanguageServer.ts
```diff
@@ -915,21 +931,134 @@ export class LanguageServer {
...
+    }
+
+    private async onSignatureHelp(params: SignatureHelpParams) {
+        console.log('onSignatureHelp', 'params', params);
+
+        // Courtesy of George Cook :) He might call it whack but it works ¯\_(ツ)_/¯
+        //get the position of a symbol to our left
+        //1. get first bracket to our left, - then get the symbol before that..
+        //really crude crappy parser..
+        //TODO this is whack - it's not even LTR ugh..
```
elsassph, line 992: It's ok: simple and good enough. That was the basis of how I did code completion for a much more complex language 😁

georgejecook:
@triwav I pinged you on slack; but a cursory read, and my trial with your vsix gives me the impression that brighterscript is not supported at all by this.

I've got code in maestro's binding plugin that looks up methods. I proprose in the short term, seeing as we don't have type support, and to save you a ton of work, that we merely dump all classes, and class members into the symbols repo, maybe in `getDocumentSymbol`.. I think @elsassph and @TwitchBronBron would have the best idea around this though, as they've got the current up-to-date knowledge on where the cached references are kept, for max performance.

As it stands, this totally regresses bs support.; hopefully it will not be too much work to add brighterscript namespace and class support; fwiw - Bron makes me support brs projects whenever I do bs changes, too :)

georgejecook:
one odd thing.. I noticed that callfunc `@.` completions were also not working. that vexed me somewhat, as we get those for free from document symbols, which you currently support: not sure whats happening there.

TwitchBronBron:
@georgejecook, you say " gives me the impression that brighterscript is not supported at all by this.". What specifically makes you think that? Do you have any reproducible steps to show that classes/methods/fields, etc, are not showing up in some situation? I know that I worked with @triwav specifically to get brighterscript support in this effort, so while it's possible we missed something, it would be helpful to know what exactly is missing and how to reproduce it.

georgejecook:
Here's the current lsp:
current-lsp
Here's the vsix from this branch:
lsp-change

Note the missing method sig popups. Ill get symbol gifs, presently.

georgejecook:
go to symbol, find symbol, from vsix from this branch: note - no symbols; does nothing. 
lsp-change-symbols-defs
current extension - does a resonable job of finding defs/symbols, even for the call func case:
lsp-current-symbols-defs

georgejecook:
I only had a cursory look at the pr, and didn't see anything in the code that seemed to look up class functions, in the way I've been doing over the last 2 weeks; but, like I say; I only skimmed it; so apologies for wrongly assuming this isn't accounted for. Clearly a bug then.

src/LanguageServer.ts
```diff
@@ -915,24 +937,171 @@ export class LanguageServer {
         this.connection.sendNotification('build-status', 'success');
     }
 
-    private async onDefinition(params: TextDocumentPositionParams): Promise<Location[]> {
-        //WARNING: This only works for a few small xml cases because the vscode-brightscript-language extension
-        //already implemented this feature, and I haven't had time to port all of that functionality over to
-        //this codebase
-        //TODO implement for brs/bs also
+    public async onWorkspaceSymbol(params: WorkspaceSymbolParams) {
+        const results = util.flatMap(
```
TwitchBronBron, line 941: We probably need `await this.waitAllProgramFirstRuns();` at the start of this function since there's no guarantee that all the workspaces are loaded when it gets activated.

src/files/BrsFile.ts
```diff
@@ -1004,6 +1013,190 @@ export class BrsFile {
...
+    }
+
+    /**
+     * Builds a single SymbolInformation object for use by LanguageServer's onWorkspaceSymbol functionality
+     */
+    private generateWorkspaceSymbols(statement: Statement, containerStatement?: ClassStatement | NamespaceStatement) {
+        let symbolKind: SymbolKind;
+        const symbols = [];
+
+        if (statement instanceof FunctionStatement) {
```
TwitchBronBron, line 1096: Don't use instanceof for any of the AST traversal. Now that the plugin system exists, plugins can inject their own AST, which might not be the same binary instance. Prefer the `astUtils/reflection` functions `isFunctionStatement`, `isNamespaceStatement`, etc...)

TwitchBronBron (commented):
Hey. I have requested a few changes. I haven't finished reviewing, but figured I should get these to you sooner rather than later.

src/LanguageServer.spec.ts
```diff
@@ -323,6 +350,449 @@ describe('LanguageServer', () => {
...
+                function pi()
+                    return 3.141592653589793
+                end function
+
+                function buildAwesome()
+                    return 42
+                end function
+            end namespace`, 'bs');
+
+            for (let i = 0; i < 2; i++) {
```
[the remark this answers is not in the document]
triwav: We use a cache after the first time. This ensures it works right each time. Will add a comment

src/LanguageServer.spec.ts
```diff
@@ -323,6 +350,449 @@ describe('LanguageServer', () => {
...
+                const symbols = await svr.onWorkspaceSymbol();
```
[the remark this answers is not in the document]
triwav: What would we sort by? Suppose if I change the name to not be the same for each I can just do it by name

src/LanguageServer.ts
```diff
@@ -915,24 +937,171 @@ export class LanguageServer {
...
+            await Promise.all(this.getWorkspaces().map(workspace => {
+                return workspace.builder.program.getWorkspaceSymbols();
+            })),
+            c => c
+        );
+
+        // Remove duplicates
+        const allSymbols = Object.values(results.reduce((map, symbol) => {
+            const key = symbol.location.uri + symbol.name;
```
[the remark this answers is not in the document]
triwav: Not sure I follow you. This all the symbols for the entire project. If you don't dedupe and you have the same file in multiple workspaces you will have duplicate symbols

src/LanguageServer.ts
```diff
@@ -915,24 +937,171 @@ export class LanguageServer {
...
 
+    private async onSignatureHelp(params: SignatureHelpParams) {
+        await this.waitAllProgramFirstRuns();
+
+        const info = this.getIdentifierInfo(params);
+        if (!info) {
+            return;
+        }
+
+        const { tokens } = Lexer.scan(info.line);
```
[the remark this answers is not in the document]
triwav: not in Brightscript 😏 has to be all on one line. I believe calling a function can be split across lines but not definition of a function

src/LanguageServer.spec.ts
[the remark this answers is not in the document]
triwav: Changed to a for loop with a switch see if that works for you.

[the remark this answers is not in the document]
TwitchBronBron: Sure, that works. I was thinking you'd sort by name. 

expect(symbols.map(x=> x.name).sort()).to.eql([
    'name1',
    'name2',
    'name3',
    ...
]);

src/Program.ts
```diff
@@ -655,24 +657,45 @@ export class Program {
...
         let file = this.getFile(pathAbsolute);
         if (!file) {
             return [];
         }
-        let results = [] as Location[];
-        let scopes = this.getScopesForFile(file);
-        for (let scope of scopes) {
-            results = results.concat(...scope.getDefinition(file, position));
+
+        if (file instanceof BrsFile) {
```
TwitchBronBron, line 686: Found another `instanceof` here. You should use `isBrsFile`.

src/files/BrsFile.ts
```diff
@@ -1004,6 +1011,195 @@ export class BrsFile {
...
+     */
+    public async getDefinition(position: Position) {
+        await this.isReady();
+
+        let results: Location[] = [];
+
+        //get the token at the position
+        const token = this.getTokenAt(position);
+
+        let definitionTokenTypes = [
```
TwitchBronBron, line 1140: [Certain keywords](https://github.com/rokucommunity/brighterscript/blob/master/src/lexer/TokenKind.ts#L418) are allowed to be used as local variable names, so we should probably include those in this list.

Also, I'm not overly familiar with the full reach of this function, but [some keywords](https://github.com/rokucommunity/brighterscript/blob/07954eed46225aa26408ec00406bf48286f28cb3/src/lexer/TokenKind.ts#L340) are allowed to be object property names, so you _might_ want to include these as well:

src/files/BrsFile.ts
```diff
@@ -1064,6 +1260,93 @@ export class BrsFile {
...
+
+    public async getReferences(position: Position) {
+        await this.isReady();
+
+        const callSiteToken = this.getTokenAt(position);
+
+        let locations = [] as Location[];
+
+        // No need to actually look if they didn't select a token we can search against
+        if (callSiteToken?.kind !== TokenKind.Identifier) {
```
TwitchBronBron, line 1321: Same comment as [this one](https://github.com/rokucommunity/brighterscript/pull/191/files#r520495177)...you need to check for more than just `TokenKind.Identifier`

src/util.ts
```diff
@@ -911,14 +911,44 @@ export class Util {
...
-    public getFulllyQualifiedClassName(className: string, namespaceName?: string) {
+    public getFullyQualifiedClassName(className: string, namespaceName?: string) {
         if (className.includes('.') === false && namespaceName) {
             return `${namespaceName}.${className}`;
         } else {
             return className;
         }
     }
 
+    public splitStringIntoLines(string: string) {
```
TwitchBronBron, line 922: Could we shorten this name a bit, perhaps call it `splitLines` or `splitIntoLines` since we can see it's a `string` based on the parameter type?

TwitchBronBron:
Looks great. Thanks for addressing all of the concerns.

PR #199, opened 2020-10-06
Type enhancements

TwitchBronBron:
The project BrighterScript originally forked from, `brs`, had a runtime type system. That's why many tokens had a `literal` property, because those would be converted into actual values. There was also the `BrsType` classes which helped with runtime data handling. Since we don't actually need to run any of the code in the AST in node, most of that was not overly useful. 

Also, a while back I had created a small basic type system in the `src/types` folder (confusingly also called `BrsType`). Those will be expanded to become the basis of our type tracking system, and already are functional for the built-in types in BrightScript. 

So, this PR removes the `brs` type system, and replaces all of that with our `BrighterScript` type system. This PR does not actually introduce any new type support, it just removes unused code and unifies all of the AST with the new type system. 

Don't let the large files changed list scare you, the majority of the changes were to unit tests.

**Notable changes:**
 - remove all the `BrsType` objects leftover from the upstream `brs` project. Things like `ValueKind`, `BrsType`, the `Token.literal` property. 
 - rename the brighterscript `BrsType` class to `BscType` for more distinction from the  now deleted from-upstream `BrsType`. 
 - Modify the `createToken` function in `astUtils/creators.ts` to accept a range, or use a default negative range. 
 - Use the `BscType` objects for basic parser type tracking (this replaces `ValueKind` and `BrsType` from upstream `brs` project). 
 - minor AST property changes for `ForStatement` and `FunctionStatement`, 
 - any `ValueKind` references in code have been replaced with an instance of a `BscType` (which will be the backbone of future type tracking)
 - Updated `ForStatement` to no longer include synthetic `step 1` tokens when those were not included in the source file.
 - remove `reflection.ts` functions for eliminated `BrsType` items.


src/astUtils/creators.ts
```diff
@@ -1,52 +1,50 @@
...
-import { BrsType, BrsString, BrsInvalid, Int32, Float } from '../brsTypes';
-import util from '../util';
 
-export function createRange(pos: Position) {
-    return util.createRange(pos.line, pos.character, pos.line, pos.character);
-}
+export const interpolatedRange = Range.create(-1, -1, -1, -1);
 
-export function createToken<T extends TokenKind>(kind: T, pos: Position, text?: string, literal?: BrsType): Token & { kind: T } {
+export function createToken<T extends TokenKind>(kind: T, text?: string, range = interpolatedRange): Token & { kind: T } {
```
TwitchBronBron, line 8: Changed this to a range instead of a position, so a real range can be passed in if desired.

src/parser/Expression.ts
```diff
@@ -223,6 +236,59 @@ export class FunctionExpression extends Expression {
     }
 }
 
+export class FunctionParameterExpression extends Expression {
```
TwitchBronBron, line 239: Moved this into the `Expressions.ts` file, Mostly unchanged.

src/parser/Expression.ts
```diff
@@ -223,6 +236,59 @@ export class FunctionExpression extends Expression {
...
+    constructor(
+        public name: Identifier,
+        public typeToken?: Token,
+        public defaultValue?: Expression,
+        public asToken?: Token
+    ) {
+        super();
+        if (typeToken) {
```
TwitchBronBron, line 247: This is new code, it derives the `BscType` based on the provided `typeToken`

src/util.ts
```diff
@@ -999,6 +965,72 @@ export class Util {
         }
         return result;
     }
+
+    /**
+     * Convert a token into a BscType
+     */
+    public tokenToBscType(token: Token) {
```
TwitchBronBron, line 972: This is a helpful utility function, which converts various token types into a BscType.

src/types/BscType.ts
```diff
@@ -0,0 +1,5 @@
+export interface BscType {
```
TwitchBronBron, line 1: Renamed from `src/types/BrsType.ts`

src/astUtils/reflection.ts
```diff
@@ -191,56 +195,48 @@ export function isFunctionParameterExpression(element: Expression | Statement):
...
 }
-export function isRoSGNode(e: any): e is RoSGNode {
-    return e?.constructor.name === 'RoSGNode';
+export function isNumberType(e: any): e is IntegerType | LongIntegerType | FloatType | DoubleType {
+    return [
+        IntegerType.name,
+        LongIntegerType.name,
+        FloatType.name,
+        DoubleType.name
+    ].includes(e?.constructor.name);
```
elsassph, line 226: I don't think that's better than using a `kind` property. The new types don't have them?
Also this creates a temporary array so it's not a great runtime profile - this could hoisted as a const outside the function.

src/parser/tests/expression/AssociativeArrayLiterals.spec.ts
```diff
@@ -49,15 +49,15 @@ describe('parser associative array literals', () => {
                 token(TokenKind.LeftCurlyBrace, '{'),
                 identifier('foo'),
                 token(TokenKind.Colon, ':'),
-                token(TokenKind.IntegerLiteral, '1', new Int32(1)),
+                createToken(TokenKind.IntegerLiteral, '1'),
```
elsassph, line 52: Why using sometimes `token` and sometimes `createToken`?

TwitchBronBron: That was an oversight on my part. the `token()` function is a test helper function, and `createToken` is from `astUtils/creators.ts`. I'll go through and change all of these tests back to `token` since there's no functional difference. Thanks!

src/astUtils/reflection.ts
TwitchBronBron: BrighterScript will eventually support more advanced types. `kind` only works for the built-in BrightScript types. For example, I would like to add typescript-like interfaces. So we would have something like:
```
new InterfaceType([{
    name: 'name',
    type: new StringType()
}, {
    name: 'age',
    type: new IntegerType()
}])
```

There's no way to represent that with a `kind` property. The closest we get is `Object` which will not provide enough the level of type safety I'd like to eventually enforce in BrighterScript. 

But good point about the runtime array. I'll hoist that out into a const.

src/astUtils/reflection.ts
```diff
@@ -191,56 +195,49 @@ export function isFunctionParameterExpression(element: Expression | Statement):
...
-export function isInvalid(value: any): value is BrsInvalid {
-    return value?.kind === ValueKind.Invalid;
-}
-export function isBoolean(value: any): value is BrsBoolean {
-    return value?.kind === ValueKind.Boolean;
+// BscType reflection
+export function isStringType(value: any): value is StringType {
+    return value?.constructor.name === StringType.name;
 }
-export function isString(value: any): value is BrsString {
```
TwitchBronBron, line 202: Removed the ValueKind and BrsXYZ reflection functions in favor of common BscType checks instead.

src/interfaces.ts
```diff
@@ -128,22 +128,6 @@ export interface VariableDeclaration {
     lineIndex: number;
 }
 
-//copied from brs (since it's not exported from there)
```
TwitchBronBron, line 131: Deleted ValueKind interface (no longer being used)

src/lexer/Lexer.ts
```diff
@@ -457,7 +455,7 @@ export class Lexer {
 
         //replace escaped quotemarks "" with a single quote
         value = value.replace(/""/g, '"');
-        this.addToken(TokenKind.StringLiteral, new BrsString(value));
```
TwitchBronBron, line 460: The concept of a `literal` type on tokens has been removed, because we don't need it anymore.

src/lexer/Token.ts
```diff
@@ -12,8 +11,6 @@ export interface Token {
     text: string;
     /** True if this token's `text` is a reserved word, otherwise `false`. */
     isReserved: boolean;
-    /** The literal value (using the BRS type system) associated with this token, if any. */
-    literal?: BrsType;
```
TwitchBronBron, line 16: Removed `literal` property because we're no longer tracking the "real" value, we can derive the Type information from the token kind.

src/parser/Expression.ts
```diff
@@ -125,8 +126,20 @@ export class FunctionExpression extends Expression {
         readonly parentFunction?: FunctionExpression
     ) {
         super();
+        if (this.returnTypeToken) {
```
TwitchBronBron, line 129: Compute the return type of the function in its constructor.

src/parser/Statement.ts
```diff
@@ -719,19 +719,17 @@ export class StopStatement extends Statement {
 
 export class ForStatement extends Statement {
     constructor(
-        readonly tokens: {
```
TwitchBronBron, line 722: Removed the `tokens` collection in favor of direct token properties (to be more consistent with rest of AST in bsc).

TwitchBronBron: Most of the churn in this function is renamed properties

src/parser/Statement.ts
```diff
@@ -1566,19 +1564,9 @@ export class ClassMethodStatement extends FunctionStatement {
...
                 newStatements.push(
                     new AssignmentStatement(
-                        {
-                            kind: TokenKind.Equal,
-                            isReserved: false,
-                            range: field.name.range,
-                            text: '=',
-                            leadingWhitespace: ''
-                        },
+                        createToken(TokenKind.Equal, '=', field.name.range),
```
TwitchBronBron, line 1567: use `create` helper instead of inline object.

src/types/ArrayType.ts
```diff
@@ -1,13 +1,13 @@
-import { BrsType } from './BrsType';
```
TwitchBronBron, line 1: Only changes for the `types/` folder is that I renamed the bsc interface `BrsType` to `BscType`

src/util.ts
```diff
@@ -335,38 +333,6 @@ export class Util {
         });
     }
 
-    public valueKindToBrsType(kind: ValueKind): BrsType {
```
TwitchBronBron, line 338: Deleted in favor of using `BscType` everywhere.

PR #204, opened 2020-10-11
Diagnostic diff

TwitchBronBron:
The original diagnostic functionality would send the entire list of diagnostics, for every file, after every validate. For a project with a large number of diagnostics, this causes extremely poor performance anytime the diagnostic window is open, because the VSCode diagnostic window re-renders every diagnostic since the references have all changed (since we sent new copies of everything).

This PR fixes that problem by only sending messages for files whose diagnostics have changed. If a file's diagnostics haven't changed since the last time they were sent, that file is skipped. Files with added/changed/removed diagnostics are the only ones that will be updated. 

VSCode handles diagnostics on a file-by-file basis, so by "skipping" a file when sending diagnostics, this causes that the file's previous diagnostics to persist.

**Notable changes:**
 - created new `DiagnosticCollection` that handles tracking the previous diagnostic landscape, and generates a diff of the current diagnostics compared to the previous one.

### DIFF
--- a/src/brsTypes/Callable.ts
+++ b/src/brsTypes/Callable.ts
@@ -7,7 +7,6 @@ import { Range } from 'vscode-languageserver';
 import { TranspileState } from '../parser/TranspileState';
 import { walk, InternalWalkMode, WalkOptions, WalkVisitor } from '../astUtils';
 import { Expression, LiteralExpression } from '../parser/Expression';
-import util from '../util';
 
 /** An argument to a BrightScript `function` or `sub`. */
 export interface Argument {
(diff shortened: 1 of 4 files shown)


src/brsTypes/Callable.ts
```diff
@@ -57,7 +56,7 @@ export class StdlibArgument implements Argument {
     }
 
     /** A fake location exists only within the BRS runtime. */
-    static InternalRange = util.createRange(-1, -1, -1, -1);
+    static InternalRange = Range.create(-1, -1, -1, -1);
```
TwitchBronBron, line 59: This was to work around a ts-node mocha testing circular reference issue that only showed up during local debugging of BrighterScript.

PR #208, opened 2020-10-12
New statement namespace bugfix

### LINKED ISSUE #203: new keyword defaults to calling namespace
```
namespace Tests
  function example()
    manager = new Manager() ' this class lookup is done as `Tests.Manager`, and therefore fails.
end namespace
```

I fix this by doing the following:

image

TwitchBronBron:
Fixes some bugs related to constructing new BrighterScript classes.

**Notable changes:**
 - fixed bug that wouldn't find the correct root-level class when calling `new` from within a namespace (fixes #203)
 - fixed bug that was not adding an omitted namespace to a class whenever that class resides in the same namespace as the `new` statement.

### DIFF
--- a/src/validators/ClassValidator.ts
+++ b/src/validators/ClassValidator.ts
@@ -35,7 +35,12 @@ export class BsClassValidator {
      */
     private getClassByName(className: string, namespaceName?: string) {
         let fullName = util.getFulllyQualifiedClassName(className, namespaceName);
-        return this.classes[fullName.toLowerCase()];
+        let cls = this.classes[fullName.toLowerCase()];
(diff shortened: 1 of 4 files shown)


src/validators/ClassValidator.ts
```diff
@@ -35,7 +35,12 @@ export class BsClassValidator {
      */
     private getClassByName(className: string, namespaceName?: string) {
         let fullName = util.getFulllyQualifiedClassName(className, namespaceName);
-        return this.classes[fullName.toLowerCase()];
```
TwitchBronBron, line 38: Fixes #203

PR #210, opened 2020-10-12
Not all the BrsFile fields need to be filled at parse time

elsassph:
Some fields are really needed only when we actively use code completion on one file so it should be computed on demand. 
This cuts up to 30% parsing time.


TwitchBronBron:
Great catch!

PR #211, opened 2020-10-12
Skip validation for "super" function calls.

### LINKED ISSUE #202: super calls are wrongly subjected to missing callToUnknownFunction diagnostics
all super calls in my project result in this erroneous diagnostic.

I'm guessing super is a keyword in this context; but I don't know the references code well enough to know if it should be there or not.
In any case, I workaround for the time being by adding this code:
image

TwitchBronBron:
Calling `super()` in class constructor methods are current emitting bs1001 ("Cannot find function with name..."). This PR adds temporary relief for that issue by skipping bs1001 checks for any function call with the exact text "super" that is found in `.bs` files. This is slightly heavy-handed, but the current validation logic does not provide enough information to know if the parent function is a class method or not. 

In future validation enhancements we will tighten up this logic.

Also, this PR adds the ability for class constructors to have non-`m.` based statements before a `super()`  call. 

Fixes #202

### DIFF
--- a/src/Scope.ts
+++ b/src/Scope.ts
@@ -559,7 +559,12 @@ export class Scope {
     private diagnosticDetectCallsToUnknownFunctions(file: BscFile, callablesByLowerName: { [lowerName: string]: CallableContainer[] }) {
         //validate all expression calls
         for (let expCall of file.functionCalls) {
-            let lowerName = expCall.name.toLowerCase();
+            const lowerName = expCall.name.toLowerCase();
+            //for now, skip validation on any method named "super" within `.bs` contexts.
+            //TODO revise this logic so we know if this function call resides within a class constructor function
+            if (file.extension === '.bs' && lowerName === 'super') {
+                continue;
+            }
 
             //get the local scope for this expression
             let scope = file.getFunctionScopeAtPosition(expCall.nameRange.start);
--- a/src/validators/ClassValidator.ts
+++ b/src/validators/ClassValidator.ts
@@ -1,13 +1,13 @@
 import { Scope } from '../Scope';
 import { DiagnosticMessages } from '../DiagnosticMessages';
 import { BsDiagnostic } from '..';
-import { CallExpression, VariableExpression } from '../parser/Expression';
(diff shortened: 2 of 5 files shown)


src/validators/ClassValidator.ts
```diff
@@ -110,43 +111,45 @@ export class BsClassValidator {
...
-                ancestorNewMethod
+                classStatement.parentClass
             ) {
-                //verify there's a `super()` as the first statement in this member's "new" method
-                let firstStatement = ((newMethod as ClassMethodStatement).func?.body?.statements[0] as ExpressionStatement)?.expression as CallExpression;
+                //prevent use of `m.` anywhere before the `super()` call
+                const cancellationToken = new CancellationTokenSource();
+                let superCall: CallExpression;
+                newMethod.func.body.walk(createVisitor({
+                    VariableExpression: (expression, parent) => {
```
TwitchBronBron, line 126: Does this `VariableExpression` catch all of the situations where `m.` could be used?

src/Scope.ts
```diff
@@ -559,7 +559,12 @@ export class Scope {
     private diagnosticDetectCallsToUnknownFunctions(file: BscFile, callablesByLowerName: { [lowerName: string]: CallableContainer[] }) {
         //validate all expression calls
         for (let expCall of file.functionCalls) {
-            let lowerName = expCall.name.toLowerCase();
+            const lowerName = expCall.name.toLowerCase();
+            //for now, skip validation on any method named "super" within `.bs` contexts.
+            //TODO revise this logic so we know if this function call resides within a class constructor function
+            if (file.extension === '.bs' && lowerName === 'super') {
```
TwitchBronBron, line 565: This is a stopgap measure to hide those pesky "Cannot find function with name 'super'" error messages that show up inside class constructor methods. We'll refactor this at a later time to be more precise about when these statements exist within a class constructor block.

src/validators/ClassValidator.ts
elsassph: If you don't get a `VariableExpression` for `m` in `thing.m` then that should do it.

PR #212, opened 2020-10-12
Allows transpilation of files that do not come from roku deploy's fil…

georgejecook:
…e entries, so generated files can be included in the build output

### DIFF
--- a/src/Program.ts
+++ b/src/Program.ts
@@ -756,11 +756,12 @@ export class Program {
         const entries = Object.values(this.files).map(file => {
             let filePathObj = fileEntries.find(x => s`${x.src}` === s`${file.pathAbsolute}`);
             if (!filePathObj) {
-              //this file has been added in-memory, from a plugin, for example
-              filePathObj =  {
-                src: file.pkgPath,
-                dest: file.pkgPath
-              };
+                //this file has been added in-memory, from a plugin, for example
+                filePathObj = {
+                    //add an interpolated src path (since it doesn't actually exist in memory)
+                    src: `bsc:/${file.pkgPath}`,
+                    dest: file.pkgPath
+                };
             }
             //replace the file extension
             let outputPath = filePathObj.dest.replace(/\.bs$/gi, '.brs');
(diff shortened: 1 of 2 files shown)


TwitchBronBron:
@georgejecook  Looks good. Just add a few tests and fix the eslint errors and this will be good to go!

src/Program.ts
```diff
@@ -754,9 +754,14 @@ export class Program {
...
     public async transpile(fileEntries: FileObj[], stagingFolderPath: string) {
         const entries = Object.values(this.files).map(file => {
-            const filePathObj = fileEntries.find(x => s`${x.src}` === s`${file.pathAbsolute}`);
+            let filePathObj = fileEntries.find(x => s`${x.src}` === s`${file.pathAbsolute}`);
             if (!filePathObj) {
-                throw new Error(`Cannot find fileMap record in fileMaps for '${file.pathAbsolute}'`);
+                //this file has been added in-memory, from a plugin, for example
+                filePathObj = {
+                    //add an interpolated src path (since it doesn't actually exist in memory)
+                    src: `bsc:/${file.pkgPath}`,
```
TwitchBronBron, line 762: I'm not sure how this will work with sourcemaps...since it will definitely use this path. @georgejecook can you test this out locally and see what happens when you try to debug one of these in-memory-only files with this branch?

I don't think there's anything else we can do....perhaps embed the entire source within the sourcemap for these files? We don't have support for that type of debugging in the extension right now, but I could see that being a future enhancement.

TwitchBronBron: Actually, these files probably shouldn't have sourcemaps at all.

TwitchBronBron: Yeah, you're right. We should have sourcemaps. However, there are certain situations where sourcemaps won't be valid. For example, adding a completely in-memory file will not point to a valid location on disk, but it would still derive a path for the sourcemap. I won't hold up this PR for this scenario, but it's something we will need to consider for the future.

Currently the debugger throws an error anytime it can't load a source file with the given path. So we need to handle that more gracefully.

TwitchBronBron: Tracked at roku-debug https://github.com/rokucommunity/roku-debug/issues/20

PR #215, opened 2020-10-14
Performance refactoring: property names completion, classes transpilation

elsassph:
- Drop expensive AST walking for collecting property names and instead collect them as part of parsing.
- Use visitor for `super` AST mod.
- Removed `util.findAllDeep`

### DIFF
--- a/src/parser/Parser.ts
+++ b/src/parser/Parser.ts
@@ -8,7 +8,8 @@ import {
     DisallowedLocalIdentifiersText,
     AllowedProperties,
     Lexer,
-    BrighterScriptSourceLiterals
+    BrighterScriptSourceLiterals,
+    isToken
 } from '../lexer';
 
 import {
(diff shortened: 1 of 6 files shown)


src/parser/Parser.ts
```diff
@@ -80,9 +80,23 @@ import {
 } from './Expression';
 import { Diagnostic, Range } from 'vscode-languageserver';
 import { Logger } from '../Logger';
-import { isCallExpression, isCallfuncExpression, isClassMethodStatement, isDottedGetExpression, isIndexedGetExpression, isVariableExpression } from '../astUtils/reflection';
+import { isCallExpression, isCallfuncExpression, isClassMethodStatement, isCommentStatement, isDottedGetExpression, isIndexedGetExpression, isVariableExpression } from '../astUtils/reflection';
 import { createVisitor, WalkMode } from '../astUtils/visitors';
 
+function createReferences(): References {
```
TwitchBronBron, line 86: Can you move this `createReferences` somewhere other than the top of the file? I prefer keeping the primary class as the topmost item in a file. It could go to the bottom of the file, get set as a static method on the class, or even just a regular method since javascript prototypes should keep this function very fast and it doesn't have any use outside of this class anyway.

PR #217, opened 2020-10-14
Use in-memory source code for diagnostic printing

georgejecook:
improves diagnostic printing, by going to program to retrieve file contents, and allowing in-memory files to be used, not failing when there is no content, and improving efficiency

### DIFF
--- a/src/ProgramBuilder.ts
+++ b/src/ProgramBuilder.ts
@@ -8,9 +8,8 @@ import { standardizePath as s, util, loadPlugins } from './util';
 import { Watcher } from './Watcher';
 import { DiagnosticSeverity } from 'vscode-languageserver';
 import { Logger, LogLevel } from './Logger';
-import { getPrintDiagnosticOptions, printDiagnostic } from './diagnosticUtils';
 import PluginInterface from './PluginInterface';
-
+import * as diagnosticUtils from './diagnosticUtils';
 /**
  * A runner class that handles
  */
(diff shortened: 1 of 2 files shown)


src/ProgramBuilder.ts
```diff
@@ -239,16 +238,23 @@ export class ProgramBuilder {
...
             //load the file text
-            let fileText = await util.getFileContents(pathAbsolute);
-            let lines = util.getLines(fileText);
-
-            for (let diagnostic of sortedDiagnostics) {
-                //default the severity to error if undefined
-                let severity = typeof diagnostic.severity === 'number' ? diagnostic.severity : DiagnosticSeverity.Error;
-                //format output
-                printDiagnostic(options, severity, filePath, lines, diagnostic);
+            try {
```
TwitchBronBron, line 243: We should not be wrapping and then throwing away any runtime exceptions that happen here...that would cause even more confusion for plugin authors and developers because there wouldn't even be runtime error messages or stack traces if something went wrong.

src/ProgramBuilder.ts
```diff
@@ -239,16 +238,16 @@ export class ProgramBuilder {
             if (!emitFullPaths) {
                 filePath = path.relative(cwd, filePath);
             }
-
             //load the file text
-            let fileText = await util.getFileContents(pathAbsolute);
-            let lines = util.getLines(fileText);
+            const file = this.program.getFileByPathAbsolute(pathAbsolute);
```
TwitchBronBron, line 242: This loads the file contents from the in-memory BrsFile/XmlFile. Previously the contents were being loaded from disk, which does not work for in-memory-only files injected by plugins.

PR #220, opened 2020-10-16
Performance: refactored getFileByPkgPath

elsassph:
Massively improve validation performance by refactoring `getFileByPkgPath`.
Like 40% faster validation for me.


TwitchBronBron:
awesome find!

PR #221, opened 2020-10-16
Performance: parser micro-optimisation

elsassph:
Micro optimisation of hot parser functions - seems about 10% faster parse.

### DIFF
--- a/src/parser/Parser.ts
+++ b/src/parser/Parser.ts
@@ -296,9 +296,9 @@ export class Parser {
     private declaration(...additionalTerminators: BlockTerminator[]): Statement | undefined {
         try {
             // consume any leading newlines
-            while (this.match(TokenKind.Newline)) { }
+            while (this.match1(TokenKind.Newline)) { }
 
-            if (this.check(TokenKind.Class)) {
+            if (this.check1(TokenKind.Class)) {
                 return this.classDeclaration();
             }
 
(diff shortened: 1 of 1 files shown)


src/parser/Parser.ts
```diff
@@ -2318,28 +2326,39 @@ export class Parser {
         return this.previous();
     }
 
-    private checkPrevious(...tokenKinds: TokenKind[]) {
-        if (this.isAtEnd()) {
+    private checkPrevious(tokenKind: TokenKind) {
+        return this.previous().kind === tokenKind;
+    }
+
+    private check1(tokenKind: TokenKind) {
```
TwitchBronBron, line 2333: Rather than calling this `check1`, how about we rename the original `check` to `checkAny`, and call this one `check`? This would apply to all of the `1` methods you created.

src/parser/Parser.ts
```diff
@@ -2318,28 +2326,39 @@ export class Parser {
...
+        return this.peekNext().kind === tokenKind;
     }
 
     private checkNext(...tokenKinds: TokenKind[]) {
         if (this.isAtEnd()) {
             return false;
         }
-
-        return tokenKinds.some(tokenKind => this.peekNext().kind === tokenKind);
+        const nextKind = this.peekNext().kind;
```
TwitchBronBron, line 2360: Nice.

TwitchBronBron:
Looks great, thanks!

PR #223, opened 2020-10-18
Now that validation is fast, reduce timings

elsassph:
Make it more responsive!
Tested in a 480+ brs files project.

### DIFF
--- a/src/Throttler.ts
+++ b/src/Throttler.ts
@@ -63,7 +63,7 @@ export class Throttler {
     private async runInternal(job) {
         this.runningJobPromise = util.sleep(this.delay).then(() => {
             //run the job
-            job();
+            return job();
         }).catch((e) => {
             //log the error, but keep moving
             console.error(e);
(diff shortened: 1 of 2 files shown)


src/Throttler.ts
```diff
@@ -63,9 +63,7 @@ export class Throttler {
     private async runInternal(job) {
         this.runningJobPromise = util.sleep(this.delay).then(() => {
             //run the job
-            return Promise.resolve(
-                job()
-            );
+            job();
```
TwitchBronBron, line 66: this isn't quite right. previously the job could return a promise, and our throttler would wait for that job. Now you're not returning the job results, meaning if it's async, the throttler will continue on as-is. I think you need to `return` the call to job at least. The promise spec mandates that value gets awaited if it's a promise. 

```
return job();
```

elsassph: Oops

TwitchBronBron:
Looks great, thanks!

PR #234, opened 2020-10-23
Feature: AST annotations

elsassph:
Annotations are metadata expressions attached to AST statements.

- any statement can receive multiple annotations,
- annotations can be on the same line as the statement,
- annotations can have arguments; arguments can be any valid BrightScript expression,
- name of the annotation must be a valid identifier; it can't be a keyword (e.g. `sub`, `while`...),
- annotations have a method to obtain JS-friendly types (`annotation.getArguments()`); basic types, arrays and roAssocArrays are converted.

Limitation:

- annotations aren't validated if they are mis-named; could be declared by the plugins using them in the future,
- currently these expressions are not walked, for various reasons.

Example:
```brightscript
@someFlag
sub someFunction()
  '...
end sub

@configuration("param", [1, 2, 3], { prop: "value" })
@andAnotherFlag
function anotherFunction()
  @translate print "this can also have meta"
end function

@replaceWith(45 * 3, sub ()
  print "you could have whole expressions here - why tho?"
end sub)
sub replaceMe()
    '...
end sub
```

A plugin can explore the AST and look at the annotations:
```typescript
const annotations = statement.annotations;
if (annotations) {
    annotations.forEach(a => console.log(a.name, a.getArguments()));
    // ex: configuration ["param", [1, 2, 3], { prop: "value" }]
}
```

### DIFF
--- a/src/parser/Parser.ts
+++ b/src/parser/Parser.ts
@@ -85,7 +85,8 @@ import {
     EscapedCharCodeLiteralExpression,
     TemplateStringQuasiExpression,
     TaggedTemplateStringExpression,
-    SourceLiteralExpression
+    SourceLiteralExpression,
+    AnnotationExpression
 } from './Expression';
 import type { Diagnostic, Range } from 'vscode-languageserver';
 import { Logger } from '../Logger';
@@ -188,6 +189,11 @@ export class Parser {
     private allowedLocalIdentifiers: TokenKind[];
 
     /**
+     * A meta statement which should be attached to the next statement
+     */
+    private pendingAnnotations: AnnotationExpression[] = [];
+
+    /**
      * Get the currently active global terminators
      */
     private peekGlobalTerminators() {
@@ -246,6 +252,11 @@ export class Parser {
                 ) {
                     let dec = this.declaration();
                     if (dec) {
+                        //attach annotations to statements
+                        if (this.pendingAnnotations.length) {
+                            dec.annotations = this.pendingAnnotations;
+                            this.pendingAnnotations = [];
+                        }
                         body.statements.push(dec);
                     }
                 }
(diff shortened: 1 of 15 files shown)


georgejecook:
simply marvellous. I will put this to good use. Thanks @elsassph

src/parser/Parser.ts
```diff
@@ -246,6 +252,11 @@ export class Parser {
                 ) {
                     let dec = this.declaration();
                     if (dec) {
+                        //attach annotations to statements
+                        if (this.pendingAnnotations.length) {
```
TwitchBronBron, line 256: shouldn't this be `this.pendingAnnotations.length > 0`?

src/parser/Parser.ts
```diff
@@ -314,6 +325,11 @@ export class Parser {
                 return this.namespaceStatement();
             }
 
+            if (this.check(TokenKind.At) && this.checkNext(TokenKind.Identifier)) {
```
TwitchBronBron, line 328: BrightScript uses the `identifier@otherIdentifier` to dereference xml attributes. Should we `checkPrevious()` to ensure there isn't a leading identifier before the `TokenKind.At`? It might not matter, but perhaps to verify, we should add a unit test to ensure a top-level `Identifier@otherIdentifier` doesn't trip this up.

src/parser/Parser.ts
```diff
@@ -2503,11 +2540,11 @@ function createReferences(): References {
 export interface References {
     assignmentStatements: AssignmentStatement[];
     classStatements: ClassStatement[];
-    namespaceStatements: NamespaceStatement[];
     functionStatements: FunctionStatement[];
     functionExpressions: FunctionExpression[];
```
TwitchBronBron, line 2544: If you're ordering these alphabetically, might as well swap `functionStatements` and `functionExpressions` too.

src/parser/Parser.ts
elsassph: Should that happen? I'm matching when something "starts" with an @

TwitchBronBron: You're right, it probably won't happen. It was just a thought.

TwitchBronBron:
@elsassph this looks great. As soon as you address [this comment](https://github.com/rokucommunity/brighterscript/pull/234#discussion_r511841477) I'll be ready to merge.

PR #235, opened 2020-10-27
XML AST

### LINKED ISSUE #177: Use a proper AST for XML files as well
For instance: https://www.npmjs.com/package/@xml-tools/ast
which seem to handle error recovery and visitor pattern.

elsassph:
Implement a proper SG AST with sourcemaps support and basic component interface validation.
It's flexible enough to allow plugins to modify the AST and create AST nodes.

A few helpers are included in `src/astUtils/xml.ts`.

Performance: parsing is 2x faster, and transpilation 2x slower, but overall shouldn't really impact current timings as XML processing represents a small part of the compiler job.
```
xml-parser@0.16.9 x 12.728 ops/sec
xml-parser@local x 25.517 ops/sec

xml-transpile@0.16.9 x 73.601 ops/sec
xml-transpile@local x 29.476 ops/sec
```

Notes:
- The XML parser is now synchronous, which allows a large number of `async` code to be removed.
- Many changes were required in the tests just because the new parser does not allow whitespace before the XML prolog.

**Cool addition**:
Now you can quickly visualize sourcemaps using:
```
npm install source-map-visualize -g
source-map-visualize dist/source/main.brs
source-map-visualize dist/components/MainScene.xml
```

Fixes #177


src/Scope.spec.ts
```diff
@@ -488,12 +488,12 @@ describe('Scope', () => {
             const validateStartScope = sinon.spy();
             const validateEndScope = sinon.spy();
             await program.addOrReplaceFile('source/file.brs', ``);
-            await program.addOrReplaceFile('components/comp.xml', `
-                <?xml version="1.0" encoding="utf-8" ?>
+            await program.addOrReplaceFile('components/comp.xml',
```
TwitchBronBron, line 491: If you didn't like having to move all opening backticks to the next line (and then trailing backticks as well), you could use the `trim` template function. [Here's](https://github.com/rokucommunity/brighterscript/blob/abae186cc0857eb866d54f05f5a7a0b86cf41197/src/testHelpers.spec.ts#L35) its definition, [here's](https://github.com/rokucommunity/brighterscript/blob/abae186cc0857eb866d54f05f5a7a0b86cf41197/src/files/BrsFile.spec.ts#L2235) how it's used. I think that would help reduce the churn on a bunch of these tests.

src/files/BrsFile.spec.ts
```diff
@@ -2463,6 +2463,30 @@ describe('BrsFile', () => {
     });
 });
 
+export function trimSource(source: string, fixIndent = true) {
```
TwitchBronBron, line 2466: Can this be removed in favor of the `trim` function mentioned earlier?

elsassph: This method doesn't do the same as trim I think - it does normalize indentation.

elsassph: Oh gotcha it does normalise indentation too.

src/Program.spec.ts
```diff
@@ -662,9 +664,9 @@ describe('Program', () => {
         it('picks up new files in a scope when an xml file is loaded', async () => {
             program.options.ignoreErrorCodes.push(1013);
             let xmlPath = s`${rootDir}/components/component1.xml`;
-            await program.addOrReplaceFile({ src: xmlPath, dest: 'components/comonent1.xml' }, `
+            await program.addOrReplaceFile({ src: xmlPath, dest: 'components/comonent1.xml' }, trim`
                 <?xml version="1.0" encoding="utf-8" ?>
-                <component name="HeroScene" extends="Scene" >');
```
TwitchBronBron, line 667: Interesting that these tests still succeeded with this junk in the XML.

src/files/BrsFile.ts
```diff
@@ -187,7 +175,7 @@ export class BrsFile {
     /**
      * Find and set the typedef variables (if a matching typedef file exists)
      */
-    private resolveTypdef() {
+    private resolveTypedef() {
```
TwitchBronBron, line 178: Nice typo catch!

src/Program.spec.ts
elsassph: I know right :D

src/files/BrsFile.ts
```diff
@@ -224,13 +212,9 @@ export class BrsFile {
     public parse(fileContents: string) {
         try {
             this.fileContents = fileContents;
-            if (this.parseDeferred.isCompleted) {
```
TwitchBronBron, line 227: Deleting this condition no longer guards against `parse` being called more than once. I'm ok with that, but then we need to ensure that any setup/initialization steps are replicated in the `parse` method (like clearing out the `diagnostics` array as one example of many).

src/files/XmlFile.spec.ts
```diff
@@ -86,69 +87,114 @@ describe('XmlFile', () => {
...
+                    <script type="text/brightscript" uri="ChildScene.brs" />
+                </component>`
+            );
+            expect(file.diagnostics).to.be.lengthOf(2);
+            expect(file.diagnostics[0]).to.deep.include({ // expecting opening tag but got prolog
+                code: 1008,
+                range: Range.create(1, 16, 1, 22)
+            });
+            expect(file.diagnostics[1]).to.deep.include({
+                code: 1008,
```
TwitchBronBron, line 128: Rather than hardcoding the `1008` code here, use the spread operator on `DiagnosticMessages.xmlGenericParseError`. 

```
expect(file.diagnostics[1]).to.deep.include({
   ...DiagnosticMessages.xmlGenericParseError('Syntax error: whitespace found before the XML prolog'),
   range: Range.create(0, 0, 1, 16)
});
```

src/files/XmlFile.spec.ts
```diff
@@ -445,38 +491,14 @@ describe('XmlFile', () => {
         expect(actual).deep.equal(expected);
     });
 
-    describe('findExtendsPosition', () => {
```
TwitchBronBron, line 448: What happened to these tests? It's important that we can properly identify the exact location of the `extends` attribute.

elsassph: I figured that it wasn't needed now that we have a range coming from the AST.

elsassph: The `findExtends` function itself doesn't exist anymore

src/files/XmlFile.spec.ts
```diff
@@ -524,24 +546,68 @@ describe('XmlFile', () => {
...
+            let file = await program.addOrReplaceFile(
+                { src: s`${rootDir}/components/SimpleScene.xml`, dest: 'components/SimpleScene.xml' },
+                trim`
+                <?xml version="1.0" encoding="utf-8" ?>
+                <component name="SimpleScene" extends="Scene" >
+                    <script type="text/brightscript" uri="SimpleScene.bs"/>
+                </component>
+            `);
+            expect(file.needsTranspiled).to.be.true;
+        });
```
TwitchBronBron, line 606: There's a `bsconfig` flag called `autoImportComponentScript` which will automatically import a .brs or .bs script with the same name as the component, without needing to add a script tag. Can you add another test that verifies `needsTranspiled` is true when that flag is enabled, and `components/SimpleScene.bs` exists and is not directly referenced by a script tag?

src/files/XmlFile.spec.ts
elsassph: I'll add back equivalent tests

src/files/XmlFile.spec.ts
elsassph: `needsTranspiled` isn't affected by the auto import - it's resolved for transpilation.

src/Scope.spec.ts
TwitchBronBron: nice, that helped a lot. Thanks!

src/files/XmlFile.spec.ts
TwitchBronBron: the auto-import results in an additional `<script uri="path/to/codebehind/file.brs">` tag being injected into the XML at transpilie-time, which most definitely would require `needsTranspiled` to be true.

elsassph: Yes but this is resolved dynamically when transpiling as a result of calling `getMissingImportsForTranspile` - this forces transpilation.

TwitchBronBron: Oh. Ok, cool.

src/DiagnosticMessages.ts
```diff
@@ -590,6 +590,11 @@ export let DiagnosticMessages = {
         message: `There are multiple components with the name '${componentName}'`,
         code: 1115,
         severity: DiagnosticSeverity.Error
+    }),
+    xmlUnknownTag: (tagName: string) => ({
+        message: `Unknown tag '${tagName}'`,
```
elsassph, line 595: Maybe "unexpected" is better?

src/files/BrsFile.ts
```diff
@@ -75,15 +75,6 @@ export class BrsFile {
      */
     public extension: string;
 
-    private parseDeferred = new Deferred();
```
TwitchBronBron, line 78: Gosh it feels so nice to not need this anymore!

src/files/XmlFile.ts
```diff
@@ -84,20 +45,19 @@ export class XmlFile {
...
-    /**
-     * If the file was given type definitions during parse. XML files never have a typedef
-     */
-    public hasTypedef = false;
-
     /**
      * The range of the component's name value
      */
-    public componentNameRange: Range;
+    public get componentNameRange(): Range {
```
TwitchBronBron, line 51: This was only here because the previous XML parser didn't support locations. I think it makes more sense to store some type of token for `componentName`, and then you can do `xmlFile.componentName.range` and `xmlFile.componentName.text`.

src/files/XmlFile.ts
```diff
@@ -84,20 +45,19 @@ export class XmlFile {
...
+        return this.parser.references.nameRange;
+    }
 
     /**
-     * The range of the component's parent name (if exist
+     * The range of the component's parent name (if exist)
      */
-    public parentNameRange: Range;
+    public get parentNameRange(): Range {
```
TwitchBronBron, line 58: Same as above comment: merge with `parentName` and store as a token.

src/files/XmlFile.ts
TwitchBronBron: Or, we might not even need `XmlFile.componentName` anymore since `parser.references` probably already stores it?

src/files/XmlFile.ts
```diff
@@ -623,51 +447,15 @@ export class XmlFile {
...
-            }
-
-            //convert .bs extensions to .brs
-            let idx = line.indexOf('.bs"');
-            if (idx > -1) {
-                line = line.substring(0, idx) + '.brs' + line.substring(idx + 3);
-            }
-
-            //convert "text/brighterscript" to "text/brightscript"
-            line = line.replace(`"text/brighterscript"`, `"text/brightscript"`);
```
TwitchBronBron, line 662: Do you apply this transformation in your new code? I couldn't find it anywhere.

src/files/XmlFile.ts
```diff
@@ -676,3 +464,67 @@ export class XmlFile {
...
+    temp.children = component.children;
+
+    //insert extra imports
+    const extraScripts = extraImports
+        .map(uri => {
+            const script = new SGScript();
+            script.uri = util.getRokuPkgPath(uri.replace(/\.bs$/, '.brs'));
+            return script;
+        });
+    if (extraScripts.length) {
```
TwitchBronBron, line 502: This should be `if(extraScripts.length > 0)`

src/parser/SGTranspileState.ts
```diff
@@ -0,0 +1,75 @@
+import type { Range } from 'vscode-languageserver';
+import { SourceNode } from 'source-map';
+import type { SGAttribute, SGToken } from './SGTypes';
+
+export class SGTranspileState {
```
TwitchBronBron, line 5: I don't really see a point in creating a separate TranspileState class just for SceneGraph. The unique features of this class could easily be moved into a utility class, and then we can share `TranspileState` between Brs and XML files.

src/parser/SGParser.ts
```diff
@@ -0,0 +1,353 @@
...
+import { SGProlog, SGChildren, SGComponent, SGField, SGFunction, SGInterface, SGNode, SGScript } from './SGTypes';
+import type { SGAst, SGTag, SGToken, SGAttribute, SGReferences } from './SGTypes';
+import { isSGComponent } from '../astUtils/xml';
+
+export default class SGParser {
+
+    /**
+     * The AST of the XML document, not including the inline scripts
+     */
+    public ast: SGAst = {};
```
TwitchBronBron, line 15: Perhaps this should be a class, just like the other SG AST elements. Then the vast majority of [XmlFile#transpileAst](https://github.com/rokucommunity/brighterscript/pull/235/files#diff-c2462ef143ff214d1683501300a6cc2960cf54d7ba36e7606f7383c95021a240R486) can be moved into this new class. 

I did something similar with the `Body` statement in the BrighterScript parser.

src/files/XmlFile.ts
```diff
@@ -676,3 +464,67 @@ export class XmlFile {
...
+        );
+    }
+
+    //sourcemap reference
+    chunks.push(`<!--//# sourceMappingURL=./${path.basename(source)}.map -->`);
+
+    return new SourceNode(null, null, source, chunks).toStringWithSourceMap();
+}
+
+function transpileAst(source: string, ast: SGAst, extraImports: string[]) {
```
TwitchBronBron, line 486: Why not call this `transpile` like the `BrsFile` equivalent? And have it operate on `this.ast` rather than having it passed in (again, like `BrsFile)

src/parser/SGTypes.ts
```diff
@@ -0,0 +1,373 @@
...
+        public tag: SGToken,
+        public attributes: SGAttribute[] = [],
+        public range?: Range
+    ) {}
+
+    getSGAttribute(name: string): SGAttribute | undefined {
+        return this.attributes.find(att => att.key.text.toLowerCase() === name);
+    }
+
+    getAttribute(name: string): string | undefined {
```
TwitchBronBron, line 37: What if we rename this to `getAttributeValue` or `getAttributeText`, and then change `getSGAttribute` to `getAttribute`?

src/parser/SGTypes.ts
```diff
@@ -0,0 +1,373 @@
...
+    }
+
+    get uri() {
+        return this.getAttribute('uri');
+    }
+    set uri(value: string) {
+        this.setAttribute('uri', value);
+    }
+
+    constructor(
```
TwitchBronBron, line 136: Can you lift all of the constructor functions to the top of these classes? I don't have a lint rule for this (is there one?), but I've tried to follow that convention in most of this project.

TwitchBronBron (changes requested):
Hey. There's a lot in this PR. Sorry it took so long to give it a thorough review. I have a bunch of suggestions and questions that I'd like you to address. Also, I need to take a little more time to look over the SceneGraph AST node classes just to get a better understanding of them.

src/Scope.ts
```diff
@@ -206,14 +207,29 @@ export class Scope {
         }
     }
 
+    /**
+     * Iterate over Brs files for callables/references
```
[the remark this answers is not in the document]
elsassph: I would like to express that we are only iterating over files relevant to code navigation / validation, as this excludes files shadowed by type def files.

src/files/XmlFile.ts
elsassph: When transpiling.

src/parser/SGTranspileState.ts
elsassph: I tried to use `TranspileState` but it was way too specific to Brs. It still feels convenient as you have less properties to pass through to all the transpile methods.

src/parser/SGTypes.ts
elsassph: I don't really agree with your pattern but ok.

TwitchBronBron:
@elsassph At this point I think we're ready to merge this. Could you please resolve the merge conflicts, and then I'll approve and merge the PR.

PR #236, opened 2020-10-31
Typedef fixes

TwitchBronBron:
Fixes issues generating class typedefs.

### DIFF
--- a/src/parser/Statement.ts
+++ b/src/parser/Statement.ts
@@ -1599,6 +1599,9 @@ export class ClassMethodStatement extends FunctionStatement {
                 ' '
             );
         }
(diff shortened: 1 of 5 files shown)


src/parser/Statement.ts
```diff
@@ -1548,7 +1548,7 @@ export class ClassMethodStatement extends FunctionStatement {
         readonly accessModifier: Token,
         name: Identifier,
         func: FunctionExpression,
-        readonly overrides: Token
+        readonly override: Token
```
TwitchBronBron, line 1551: renamed this to `override` since that's what the actual keyword is.

PR #237, opened 2020-11-01
support on-demand parse for typedef-shadowed files

TwitchBronBron:
Plugins and external consumers of `.brs`/`.bs` files should not need to know whether a file parsing has been skipped due to typedef loading. This PR addresses the issue by converting the `BrsFile.parser` into a getter, and parsing the file on-demand if the parsing was previously skipped.

### DIFF
--- a/src/files/BrsFile.spec.ts
+++ b/src/files/BrsFile.spec.ts
@@ -2134,18 +2134,19 @@ describe('BrsFile', () => {
 
     describe('type definitions', () => {
         it('only exposes defined functions even if source has more', async () => {
-            await program.addOrReplaceFile('source/main.d.bs', `
-                sub main()
-                end sub
-            `);
-
-            const file = await program.addOrReplaceFile<BrsFile>('source/main.brs', `
+            //parse the .brs file first so it doesn't know about the typedef
+            await program.addOrReplaceFile<BrsFile>('source/main.brs', `
                 sub main()
                 end sub
                 sub speak()
                 end sub
             `);
-            expect(file.parser.references.functionStatements).to.be.empty;
+
+            await program.addOrReplaceFile('source/main.d.bs', `
+                sub main()
+                end sub
+            `);
+
             const sourceScope = program.getScopeByName('source');
             const functionNames = sourceScope.getAllCallables().map(x => x.callable.name);
             expect(functionNames).to.include('main');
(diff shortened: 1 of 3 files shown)


src/files/BrsFile.spec.ts
```diff
@@ -2134,18 +2134,19 @@ describe('BrsFile', () => {
 
     describe('type definitions', () => {
         it('only exposes defined functions even if source has more', async () => {
-            await program.addOrReplaceFile('source/main.d.bs', `
```
TwitchBronBron, line 2137: Swapped the order of the file loading, so we can ensure the file was fully parsed (since it didn't know about the .d.bs file)

src/files/BrsFile.ts
```diff
@@ -138,7 +138,23 @@ export class BrsFile {
         }
     }
 
-    public parser = new Parser();
+    public get parser() {
+        if (!this._parser) {
+            //remove the typedef file (if it exists)
+            this.hasTypedef = false;
```
TwitchBronBron, line 144: these need to be removed before calling `parse`, because `parse` will stop parsing if it detects a typedef file. It's a bit of a hack, but since it's all internal logic, it works fine.

src/files/BrsFile.ts
```diff
@@ -188,13 +204,6 @@ export class BrsFile {
         //event that fires anytime a dependency changes
         this.unsubscribeFromDependencyGraph = this.program.dependencyGraph.onchange(this.dependencyGraphKey, () => {
             this.resolveTypdef();
-
-            //if there is no typedef file, and this file hasn't been parsed yet, parse it now
-            //(solves issue when typedef gets deleted and this file had skipped parsing)
-            if (!this.hasTypedef && this.wasParseSkipped) {
```
TwitchBronBron, line 194: This is no longer necessary, since the parser is a getter, it will parse the file on-demand the first time it's needed.

src/files/BrsFile.ts
```diff
@@ -208,11 +217,6 @@ export class BrsFile {
         dependencyGraph.addOrReplace(this.dependencyGraphKey, dependencies);
     }
 
-    /**
-     * Was parsing skipped because the file has a typedef?
-     */
-    private wasParseSkipped = false;
```
TwitchBronBron, line 214: not necessary now that parser is a getter.

src/files/BrsFile.ts
```diff
@@ -891,8 +893,9 @@ export class BrsFile {
 
         //consume tokens backwards until we find someting other than a dot or an identifier
         let tokens = [];
-        for (let i = this.parser.tokens.indexOf(currentToken); i >= 0; i--) {
-            currentToken = this.parser.tokens[i];
+        const parser = this.parser;
```
TwitchBronBron, line 896: lift the parser out of the loop so we only pay the getter cost once.

src/files/BrsFile.ts
```diff
@@ -932,8 +935,9 @@ export class BrsFile {
     }
 
     public getPreviousToken(token: Token) {
-        let idx = this.parser.tokens.indexOf(token);
-        return this.parser.tokens[idx - 1];
+        const parser = this.parser;
```
TwitchBronBron, line 938: lift the parser out of the loop so we only pay the getter cost once.

src/globalCallables.ts
```diff
@@ -10,12 +10,10 @@ import { InterfaceType } from './types/InterfaceType';
 import { ObjectType } from './types/ObjectType';
 import { StringType } from './types/StringType';
 import { VoidType } from './types/VoidType';
-import { Parser } from './parser';
 import util from './util';
 
 export let globalFile = new BrsFile('global', 'global', null);
-globalFile.parser = new Parser();
-globalFile.parser.parse([]);
+globalFile.parse('');
```
TwitchBronBron, line 16: simplify the global file a bit

src/files/BrsFile.ts
elsassph: Could use `_parser` no?

TwitchBronBron: No, that's not safe. This is a public method, which provides no guarantees about who will be calling it. Could be a plug-in, in which case _parser might be undefined.

PR #242, opened 2020-11-16
Load plugins relatively to the project

elsassph:
When running through VS Code extension, plugins were loaded relatively to the extension, not the project.


TwitchBronBron:
Looks good. One thing: could you add a few unit tests where you pass in a different cwd through options?

elsassph:
Done @TwitchBronBron 
Though I can't test loading a plugin as npm module yet.

PR #246, opened 2020-11-19
Class name collision

### LINKED ISSUE #245: Show diagnostic error when a variable name or function name overwrites/collides with a Class name
There is no warning/error when you declare a variable with the same name as a class. There should be :)

Example program:

```
class TestClass
  sub new()
  end sub
end class

function TestFunction() as string
  return "test"
end function

sub Main()
  ' functions already have name collision checking (brs error 1104)
  ' this will have a red line
  testFunction = TestFunction() 


  ' this will not show a red error line
  ' gives run time error:  Function Call Operator ( ) attempted on non-function
  testClass = new TestClass()
  ? TestClass
end sub
```

TwitchBronBron:
Catch when local variables and scope functions have the same name as a class.

Fixes #245

### DIFF
--- a/src/DiagnosticMessages.ts
+++ b/src/DiagnosticMessages.ts
@@ -547,7 +547,7 @@ export let DiagnosticMessages = {
         severity: DiagnosticSeverity.Error
     }),
     localVarSameNameAsClass: (className: string) => ({
-        message: `Local variable has same name as class ${className}`,
+        message: `Local variable has same name as class '${className}'`,
         code: 1106,
         severity: DiagnosticSeverity.Error
     }),
@@ -592,7 +592,7 @@ export let DiagnosticMessages = {
         severity: DiagnosticSeverity.Error
     }),
     functionCannotHaveSameNameAsClass: (className: string) => ({
-        message: `Function cannot have the same name as class '${className}'`,
+        message: `Function has same name as class '${className}'`,
         code: 1116,
         severity: DiagnosticSeverity.Error
     })
--- a/src/DiagnosticMessages.ts
+++ b/src/DiagnosticMessages.ts
@@ -590,6 +590,11 @@ export let DiagnosticMessages = {
         message: `There are multiple components with the name '${componentName}'`,
         code: 1115,
         severity: DiagnosticSeverity.Error
+    }),
+    functionCannotHaveSameNameAsClass: (className: string) => ({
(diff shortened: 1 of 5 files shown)


src/Scope.ts
```diff
@@ -292,16 +313,6 @@ export class Scope {
         return namespaceLookup;
     }
 
-    private buildClassLookup() {
```
TwitchBronBron, line 295: Moved internal to `getClassMap()`

src/DiagnosticMessages.ts
```diff
@@ -590,6 +590,11 @@ export let DiagnosticMessages = {
         message: `There are multiple components with the name '${componentName}'`,
         code: 1115,
         severity: DiagnosticSeverity.Error
+    }),
+    functionCannotHaveSameNameAsClass: (className: string) => ({
+        message: `Function cannot have the same name as class '${className}'`,
```
elsassph, line 595: Maybe more consistent with `LocalVar` message?

src/Scope.ts
```diff
@@ -431,16 +442,32 @@ export class Scope {
...
-                    ...DiagnosticMessages.scopeFunctionShadowedByBuiltInFunction(),
-                    range: func.nameRange,
-                    file: file
-                });
+            const funcName = func.getName(ParseMode.BrighterScript);
+            const lowerFuncName = funcName?.toLowerCase();
+            if (lowerFuncName) {
+
+                //find function declarations with the same name as a stdlib function
+                if (globalCallableMap.has(func.getName(ParseMode.BrighterScript))) {
```
elsassph, line 455: Use `lowerFuncName`?

TwitchBronBron: oh yeah, good catch! I fixed it and added [this](https://github.com/rokucommunity/brighterscript/pull/246/commits/8d1cdef7c0971cabe9ad3c0c655952768c957772#diff-c9178fef239b00c0b884835914c57f03a67aae8456cad2579a2531ea141d509eL300) unit test to ensure it doesn't regress.

src/DiagnosticMessages.ts
TwitchBronBron: @elsassph which LocalVar message are you referring to?

TwitchBronBron: Oh, this one: `localVarSameNameAsClass`. Yeah, good call. It's subtle, but they should be consistent.

PR #248, opened 2020-11-20
Try catch support

### LINKED ISSUE #218: Try catch keyword support
Please add try catch block support which releases in 9.4fw
Usually looks in next way:
`SUB main()
    x = 1
    try
        PRINT x.foo
    catch e
        PRINT e.number,e.message
    end try
END SUB
`

TwitchBronBron:
Add support for try/catch statements and `throw` statements.

Resolves #218

### DIFF
--- a/src/parser/Parser.Class.spec.ts
+++ b/src/parser/Parser.Class.spec.ts
@@ -73,12 +73,12 @@ describe('parser class', () => {
     it('does not allow "throw" to be defined as a local var', () => {
         const parser = Parser.parse(`
             sub main()
-                'not allowed to define throw as local var
+                'not allowed to define "throw" as local var
                 throw = true
             end sub
         `);
 
-        expect(parser.diagnostics[0]?.message).to.eql(DiagnosticMessages.cannotUseReservedWordAsIdentifier('throw').message);
+        expect(parser.diagnostics[0]?.message).to.eql(DiagnosticMessages.foundUnexpectedToken('throw').message);
     });
 
     it('does not allow function named "throw"', () => {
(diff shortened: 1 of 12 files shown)


src/parser/Parser.Class.spec.ts
```diff
@@ -73,12 +73,12 @@ describe('parser class', () => {
...
         const parser = Parser.parse(`
             sub main()
-                'not allowed to define throw as local var
+                'not allowed to define "throw" as local var
                 throw = true
             end sub
         `);
 
-        expect(parser.diagnostics[0]?.message).to.eql(DiagnosticMessages.cannotUseReservedWordAsIdentifier('throw').message);
+        expect(parser.diagnostics[0]?.message).to.eql(DiagnosticMessages.foundUnexpectedToken('=').message);
```
TwitchBronBron, line 81: Checking for `throw` token earlier in the process causes a more generic error to be thrown. It's not really worth refactoring the parser just to get a slightly better error message, and we will definitely still get an error message so it's fine to just change this test.

src/parser/tests/statement/TryCatch.spec.ts
```diff
@@ -0,0 +1,118 @@
...
+            try: print a.b.c :  catch e : print "error"
+            end try
+        `);
+
+        expectNoParseErrors(`
+            try
+            : print a.b.c : catch e : print "error" : end try
+        `);
+
+        // expectNoParseErrors(`
```
TwitchBronBron, line 86: These three outlier syntaxes aren't currently supported. I want to support them eventually, but just don't have the time right now. I left them here so we don't forget to support them sometime in the future.

PR #252, opened 2020-12-16
Xml typedef inheritance fix

TwitchBronBron:
This fixes a bug where the typedefs in parent components were not properly recognized in child component scopes.

I've introduced the concept of "own" files for scopes:
 - *all files*: every file the scope references, either in its own direct references, or from parent scopes (like xml component inheritance)
 - *own files*: only files directly referenced by the current scope (does not account for parent scope files)


TwitchBronBron:
@georgejecook can you please review this, and then also test it out with your projects (since you were the one who found this issue).

src/files/XmlFile.spec.ts
```diff
@@ -595,6 +595,58 @@ describe('XmlFile', () => {
...
+                </component>
+            `);
+            await program.addOrReplaceFile('components/ChildComponent.bs', `
+                sub init()
+                    Parent.log()
+                    Lib.log()
+                end sub
+            `);
+            await program.validate();
+            expect(program.getDiagnostics()[0]?.message).not.to.exist;
```
elsassph, line 642: This test always looks weird. You should be testing that the collection is empty.

TwitchBronBron: The reason I do it this way is because `"expected array of length 1 to be length 0"` isn't overly helpful when trying to diagnose why a test is failing. This test does _technically_ check if the collection is empty. This pattern will print the first diagnostic to the console when the test fails, which is nice to see the problem right away in the console, as well as in Github Actions because we can't debug unit tests there.

elsassph: Oh ok that's a fair reason!

TwitchBronBron:
@georgejecook thanks for testing the PR and verifying it works for you.

PR #254, opened 2020-12-17
Support integer type declaration char

### LINKED ISSUE #233: Unexpected character '%' (char code 37)
When writing any integer literal such as `0%` the error `Unexpected character '%' (char code 37)` occurs

TwitchBronBron:
According to the [BrightScript documentation](https://developer.roku.com/docs/references/brightscript/language/expressions-variables-types.md#type-declaration-characters), a `%` can be appended to an integer literal to enforce that it becomes an integer type. 

image

Currently bsc shows a syntax error on that trailing `%`, so this PR fixes that. 

Fixes #233

### DIFF
--- a/src/lexer/Lexer.spec.ts
+++ b/src/lexer/Lexer.spec.ts
@@ -792,6 +792,12 @@ describe('lexer', () => {
     });
 
     describe('long integer literals', () => {
+        it('respects \'&\' suffix', () => {
+            let f = Lexer.scan('1&').tokens[0];
+            expect(f.kind).to.equal(TokenKind.LongIntegerLiteral);
+            expect(f.text).to.eql('1&');
+        });
+
         it('supports hexadecimal literals', () => {
             let i = Lexer.scan('&hf00d&').tokens[0];
             expect(i.kind).to.equal(TokenKind.LongIntegerLiteral);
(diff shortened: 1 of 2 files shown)


src/lexer/Lexer.spec.ts
```diff
@@ -792,6 +792,12 @@ describe('lexer', () => {
     });
 
     describe('long integer literals', () => {
+        it('respects \'&\' suffix', () => {
```
TwitchBronBron, line 795: Not directly impacted by this code, but the test for this was missing so I added it.

src/lexer/Lexer.ts
```diff
@@ -655,15 +652,10 @@ export class Lexer {
             // replace the exponential marker with a JavaScript-friendly "e"
             asString = this.source.slice(this.start, this.current).replace(/[dD]/, 'e');
             this.addToken(TokenKind.DoubleLiteral);
-            return;
-        }
-
-        if (designator === '!') {
+        } else if (designator === '!') {
```
TwitchBronBron, line 655: I combined several separate if statements into one continuous if...else if...else if... block. This cleaned up the code by removing several `return` statements and making it easier to follow the flow.

src/lexer/Lexer.ts
```diff
@@ -680,21 +672,18 @@ export class Lexer {
...
-        }
-
-        if (designator === '&') {
+        } else if (designator === '&') {
             // numeric literals ending with "&" are forced to LongIntegers
-            asString = this.source.slice(this.start, this.current);
             this.advance();
             this.addToken(TokenKind.LongIntegerLiteral);
-
+        } else if (designator === '%') {
```
TwitchBronBron, line 683: This is the core of the change. we just needed a separate if block for handling `%`

elsassph (approved):

PR #257, opened 2020-12-17
Fix file extension detection logic.

### LINKED ISSUE #256: Brightscript doesn't like extending classes in files with extra periods?
If have three files:

`testclass.bs` with no issues:

```VB
class Foo
end class

class Bar extends Foo
  sub new()
    super()
  end sub
end class
```
-------------------

`testclass_no_testdot.bs` again has no issues:

```VB
class BarNoDot extends Foo
  sub new()
    super()
  end sub
end class
```
---------------------------

However, in `testclass.dot.bs`:

```VB
class BarDot extends Foo
  sub new()
    super()
  end sub
end class
```

`super()` is a error!:

```
error BS1001: Cannot find function with name 'super' when this file is included in scope 'source'
```

TwitchBronBron:
Internal brightscript logic depends on knowing the file extension. However, when we switched to capturing all the trailing file extensions to support `.d.bs`, this broke the ability to use dots in filenames (like `main.spec.bs` for instance). So instead, we now only capture the last dot and trailing text, EXCEPT when it's exactly `.d.bs`. 

Fixes #256


src/testHelpers.spec.ts
```diff
@@ -42,3 +44,17 @@ export function trim(strings: TemplateStringsArray, ...args) {
     }
     return trimLeading(text);
 }
+
+/**
+ * Test that the given object has zero diagnostics. If diagnostics are found, they are printed to the console in a pretty fashion.
+ */
+export function expectZeroDiagnostics(obj: { getDiagnostics(): BsDiagnostic[] }) {
```
TwitchBronBron, line 51: @elsassph here's a helper function to make the weird tests mentioned in [your comment here](https://github.com/rokucommunity/brighterscript/pull/252#discussion_r544958176) a little less confusing. This will print out a much nicer set of all diagnostics, their lines, the bs code, and range. While also making the usage nicer as well:

```
it('has no diagnostics', ()=>{
})```

File: src/testHelpers.spec.ts [hunk 2]
```diff
@@ -42,3 +44,17 @@ export function trim(strings: TemplateStringsArray, ...args) {
...
+    const diagnostics = obj.getDiagnostics();
+    if (diagnostics.length > 0) {
+        let message = `Expected 0 diagnostics, but instead found ${diagnostics.length}:`;
+        for (const diagnostic of diagnostics) {
+            message += `\n        • bs${diagnostic.code} "${diagnostic.message}" at ${diagnostic.file.pathAbsolute}#${diagnostic.range.start.line}:${diagnostic.range.start.character}-${diagnostic.range.end.line}:${diagnostic.range.end.character}`;
+        }
+        throw new Error(message);
```
elsassph, line 58: I think you can `assert.fail(message)`

src/util.ts
```diff
@@ -1033,7 +1033,22 @@ export class Util {
...
+    /**
+     * Get the extension for the given file path. Basically the part after the final dot, except for
+     * `d.bs` which is treated as single extension
+     */
+    public getExtension(filePath: string) {
+        filePath = filePath.toLowerCase();
+        if (filePath.endsWith('.d.bs')) {
+            return '.d.bs';
+        } else {
+            const parts = filePath.split('.');
```
elsassph, line 1047: `split` is pretty wasteful.
Maybe get `lastIndexOf('.')` or maybe just test explicitly the extensions you support, like `endsWith('.bs')`, `endsWith('.xml')` and the rest is ''.

TwitchBronBron: Sure, I can change that. `lastIndexOf` is much quicker (relatively speaking...neither would have any tangible impact on performance since ops/sec is in the billions for either implementation).
image

src/testHelpers.spec.ts
TwitchBronBron: done.

TwitchBronBron:
@elsassph I've addressed all of your comments.

PR #258, opened 2020-12-18
Remove FunctionScope and improve type tracking

TwitchBronBron:
The variable defined in `for each` statements were not showing up for completions. This PR fixes that.


src/parser/Statement.ts
```diff
@@ -852,6 +852,7 @@ export class ForEachStatement extends Statement {
 
     walk(visitor: WalkVisitor, options: WalkOptions) {
         if (options.walkMode & InternalWalkMode.walkExpressions) {
+            walk(this, 'item', visitor, options);
```
TwitchBronBron, line 855: @elsassph is this right?

src/files/BrsFile.ts
```diff
@@ -537,6 +537,20 @@ export class BrsFile {
                 });
             }
 
+            //add every loop-defined variable
+            func.body.walk(createVisitor({
```
elsassph, line 541: Mixed feeling about this walk: 
- how often is it executed?
- it should probably be collected in the references,
- but on the other hand, doing a late walking only for the function you're interested in can be good.

elsassph: As for the assignment references, they really only make sense for the function you need completion in, so it should be done on demand.

src/parser/Statement.ts
TwitchBronBron: We spoke on slack and you said it's ok, so I'm leaving it.

src/files/BrsFile.ts
TwitchBronBron: I don't see any benefits to converting this data to on-demand because `Scope.diagnosticDetectShadowedLocalVars()` references `FunctionScope.variableDeclarations`, which is run on every validation cycle. 

However, we definitely could move these into the references. I'll give that a try and see how it goes.

TwitchBronBron:
So the scope of this PR kinda spiraled out of control a bit, but I think for the better. I _did_ still fix the `for-each-var` completions issue. However, I also made a few other changes:

 - completely removed the `FunctionScope` concept, in favor of capturing variable declarations for each function at parse-time and storing them in `references`. This cut down on a lot of after-parse looping.
 -  Add `LazyType` to defer deriving the type of certain variable types until after the file has been parsed. 

This change yielded some fairly impressive performance improvements for the `validate` cycle:
image



@elsassph I'm ready for your review again.

src/parser/Parser.ts
```diff
@@ -2418,17 +2466,70 @@ export class Parser {
...
     }
 
+    private getBscTypeFromAssignment(assignment: AssignmentStatement, functionExpression: FunctionExpression): BscType {
+        try {
+            //function
+            if (isFunctionExpression(assignment.value)) {
+                let functionType = new FunctionType(assignment.value.returnType);
+                functionType.isSub = assignment.value.functionType.text === 'sub';
+                if (functionType.isSub) {
+                    functionType.returnType = new VoidType();
```
elsassph, line 2476: What about sub with explicit return type?

src/parser/Parser.ts
```diff
@@ -2449,6 +2550,15 @@ export class Parser {
                 if (!isClassMethodStatement(parent)) {
                     this._references.functionExpressions.push(expression);
                 }
+                currentFunctionExpression = expression;
```
elsassph, line 2553: With full recursive walk this is going to be wrong with nested functions:
- enter first function, `currentFunctionExpression=A`
- new var belongs to `A`
- enter sub function, `currentFunctionExpression=B`
- new var belongs to `B`
- exit sub function, `currentFunctionExpression=B`
- new var belongs to `B` though we're in `A`

elsassph: We'd need to change the walking to be non-recursive, of do this walking "late" using a non-recursive walk just when needed.

src/parser/Parser.ts
```diff
@@ -2418,17 +2466,70 @@ export class Parser {
...
+                //literal
+            } else if (isLiteralExpression(assignment.value)) {
+                return assignment.value.type;
+
+                //function call
+            } else if (isCallExpression(assignment.value)) {
+                let calleeName = ((assignment.value.callee as any).name.text as string).toLowerCase();
+                if (calleeName) {
+                    //make a lazy type which will not compute its type until the file is done parsing
+                    return new LazyType(() => {
```
elsassph, line 2496: Ew I don't know about that...
Either the name can be resolved now, or it's a scope callable (which includes other top-level functions), and that doesn't require nearly as much magic closure work.

src/parser/Parser.ts
```diff
@@ -2418,17 +2466,70 @@ export class Parser {
...
+                        //TODO this should really be scope-aware...
+                        const func = this.references.functionStatements.find(x => x.name.text.toLowerCase() === calleeName);
+                        return func?.func.returnType ?? new DynamicType();
+                    });
+                }
+            } else if (isVariableExpression(assignment.value)) {
+                //defer the type until first read, which will allow us to derive types from variables defined after this one (like from a loop perhaps)
+                return new LazyType(() => {
```
elsassph, line 2504: Same as before. I think `LazyTypes` should be more of a "scope-aware" callable resolver, or something we keep track of in references and resolve as part of scope validation (that's what I do in bslint variable tracking logic).

src/files/BrsFile.ts
TwitchBronBron: Based on my refactor, the `FunctionScope` concept has been eliminated completely so this is no longer relevant.

src/parser/Parser.ts
TwitchBronBron: Good catch. I actually removed this logic because this exact logic is already handled (correctly) inside the FunctionExpression.

src/parser/Parser.ts
TwitchBronBron: Ah, great catch! That's what I get for skipping writing unit tests for this! 

I refactored and added some tests in [this commit](https://github.com/rokucommunity/brighterscript/pull/258/commits/73c432c8485f870ddb0c10d36765269ca302d0b2). Is that better?

elsassph: Looks sensible now. Good idea to use the "growing" list of function expressions as they are discovered.

chrisdp:
Is this ready for reviewing? If so I might be interested in hopping on a call for this one due to the scope and size.

TwitchBronBron:
No, this isn't ready for reviewing yet. I still need to refactor some of typing stuff to look into scopes rather than just per-file.

src/parser/Parser.ts
TwitchBronBron: This has been resolved in recent commits. @markwpearce created a `SymbolTable` that tracks this stuff at various levels of the program (Scope, Parser, FunctionExpression, NamespaceStatement)

TwitchBronBron:
Closing in favor of the v1 PR #401

PR #260, opened 2020-12-23
CreateObject can take 4 parameters (roScreen)

TrevorSundberg:
This is to support calls like this:
```
CreateObject("roScreen", true, 720, 480)
```

See https://developer.roku.com/docs/references/brightscript/components/roscreen.md


TwitchBronBron:
Nice catch. Thanks!

PR #262, opened 2021-01-04
Allows "as <CustomType>" in parameter and return type declarations

markwpearce:
I don't think this works yet "inside" a namespace...

### DIFF
--- a/src/DiagnosticMessages.ts
+++ b/src/DiagnosticMessages.ts
@@ -231,7 +231,7 @@ export let DiagnosticMessages = {
         severity: DiagnosticSeverity.Error
     }),
     functionParameterTypeIsInvalid: (parameterName: string, typeText: string) => ({
-        message: `Function parameter '${parameterName}' is of invalid type '${parameterName}'`,
+        message: `Function parameter '${parameterName}' is of invalid type '${typeText}'`,
         code: 1044,
         severity: DiagnosticSeverity.Error
     }),
@@ -604,16 +604,6 @@ export let DiagnosticMessages = {
     missingExceptionExpressionAfterThrowKeyword: () => ({
         message: `Missing exception expression after 'throw' keyword`,
         code: 1118,
-        severity: DiagnosticSeverity.Error
-    }),
-    callableHasUnknownReturnType: (returnType: string) => ({
-        message: `Callable has unknown return type '${returnType}'`,
-        code: 1119,
-        severity: DiagnosticSeverity.Error
-    }),
-    parameterHasUnknownType: (paramType: string) => ({
-        message: `Parameter has unknown type '${paramType}'`,
-        code: 1120,
         severity: DiagnosticSeverity.Error
     })
 };
--- a/src/DiagnosticMessages.ts
+++ b/src/DiagnosticMessages.ts
@@ -605,6 +605,16 @@ export let DiagnosticMessages = {
         message: `Missing exception expression after 'throw' keyword`,
         code: 1118,
         severity: DiagnosticSeverity.Error
+    }),
+    callableHasUnknownReturnType: (returnType: string) => ({
+        message: `Callable has unknown return type '${returnType}'`,
+        code: 1119,
+        severity: DiagnosticSeverity.Error
+    }),
+    parameterHasUnknownType: (paramType: string) => ({
+        message: `Parameter has unknown type '${paramType}'`,
+        code: 1120,
+        severity: DiagnosticSeverity.Error
     })
 };
 
--- a/src/Scope.spec.ts
+++ b/src/Scope.spec.ts
@@ -510,8 +510,8 @@ describe('Scope', () => {
             `);
             await program.validate();
             expect(program.getDiagnostics().map(x => x.message)).to.eql([
-                DiagnosticMessages.callableHasUnknownReturnType('unknownType').message,
-                DiagnosticMessages.callableHasUnknownReturnType('unknownType').message
+                DiagnosticMessages.invalidFunctionReturnType('unknownType').message,
+                DiagnosticMessages.invalidFunctionReturnType('unknownType').message
             ]);
         });
 
@@ -533,8 +533,8 @@ describe('Scope', () => {
             `);
             await program.validate();
             expect(program.getDiagnostics().map(x => x.message)).to.eql([
-                DiagnosticMessages.parameterHasUnknownType('unknownType').message,
-                DiagnosticMessages.parameterHasUnknownType('unknownType').message
+                DiagnosticMessages.functionParameterTypeIsInvalid('unknownParam', 'unknownType').message,
+                DiagnosticMessages.functionParameterTypeIsInvalid('unknownParam', 'unknownType').message
             ]);
         });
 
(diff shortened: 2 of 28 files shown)


src/Scope.spec.ts
```diff
@@ -484,6 +484,60 @@ describe('Scope', () => {
...
+                    end function
+                end class
+
+                function d() as myClass
+                    return new myClass()
+                end function
+            `);
+            await program.validate();
+            expect(program.getDiagnostics().length).to.equal(2);
+            expect(program.getDiagnostics().map(x => x.message)).to.include(
```
TwitchBronBron, line 513: We should check for both diagnostic messages. You could skip the length check, and instead do this:
```
expect(program.getDiagnostics().map(x => x.message)).to.eql(
    DiagnosticMessages.callableHasUnknownReturnType('unknownType').message,
    DiagnosticMessages.callableHasUnknownReturnType('unknownType').message
);
```

src/Scope.spec.ts
```diff
@@ -484,6 +484,60 @@ describe('Scope', () => {
...
+                    end sub
+                end class
+
+                sub d(obj as myClass)
+                end sub
+            `);
+            await program.validate();
+            expect(program.getDiagnostics().length).to.equal(2);
+            expect(program.getDiagnostics().map(x => x.message)).to.include(
+                DiagnosticMessages.parameterHasUnknownType('unknownType').message
```
TwitchBronBron, line 537: We should check for both diagnostic messages. You could skip the length check, and instead do this:
```
expect(program.getDiagnostics().map(x => x.message)).to.eql(
    DiagnosticMessages.callableHasUnknownReturnType('unknownType').message,
    DiagnosticMessages.callableHasUnknownReturnType('unknownType').message
);
```

src/Scope.ts
```diff
@@ -503,6 +505,46 @@ export class Scope {
         }
     }
 
+    /**
+    * Find function parameters and function return types that are neither built-in types or known Class references
+    */
+    private diagnosticDetectInvalidFunctionExpressionTypes(file: BscFile) {
+        const classMap = this.getClassMap();
+        for (let func of file.parser.references.functionExpressions) {
+            if (func.returnType instanceof CustomType) {
```
TwitchBronBron, line 514: Don't use `instanceof` because it won't work for AST provided by a different binary version of brighterscript (i.e. v0.23.0 and v0.23.1 would return false for instanceof). Instead, add a new function in `src/astUtils/reflection.ts` like this: https://github.com/rokucommunity/brighterscript/blob/b69b86259fed32b69841d87c04ca107deb2f3dbc/src/astUtils/reflection.ts#L227-L229

src/Scope.ts
```diff
@@ -503,6 +505,46 @@ export class Scope {
...
+                            ...DiagnosticMessages.callableHasUnknownReturnType(returnTypeName),
+                            range: func.returnTypeToken.range,
+                            file: file
+                        });
+                    }
+                }
+            }
+
+            for (let param of func.parameters) {
+                if (param.type instanceof CustomType) {
```
TwitchBronBron, line 530: Don't use instanceof. See above comment: https://github.com/rokucommunity/brighterscript/pull/262/files#r551863075

src/types/CustomType.ts
```diff
@@ -0,0 +1,9 @@
+import type { BscType } from './BscType';
+import { ObjectType } from './ObjectType';
+
+export class CustomType extends ObjectType implements BscType {
+
+    constructor(public customTypeName: string) {
```
TwitchBronBron, line 6: No need for `customType` to be in the `customTypeName` property name. Just call it `name`. Similarto how the `FunctionType` has a `name` property. https://github.com/rokucommunity/brighterscript/blob/b69b86259fed32b69841d87c04ca107deb2f3dbc/src/types/FunctionType.ts#L14

src/parser/Expression.ts
```diff
@@ -216,7 +216,7 @@ export class FunctionExpression extends Expression implements TypedefProvider {
                 new SourceNode(this.asToken.range.start.line + 1, this.asToken.range.start.character, state.pathAbsolute, 'as'),
                 ' ',
                 //return type
-                new SourceNode(this.returnTypeToken.range.start.line + 1, this.returnTypeToken.range.start.character, state.pathAbsolute, this.returnTypeToken.text.toLowerCase())
+                util.bscTypeToTypeString(this.returnType)
```
TwitchBronBron, line 219: We still need to create a `SourceNode` here. The `SourceNode` tells the sourcemap generator that "here's where the original token was, and here's what the new text is". This line should look something like this:

```
new SourceNode(
    this.returnTypeToken.range.start.line + 1, 
    this.returnTypeToken.range.start.character, 
    state.pathAbsolute, 
    util.bscTypeToTypeString(this.returnType)
)
```

src/parser/Expression.ts
```diff
@@ -291,7 +291,7 @@ export class FunctionParameterExpression extends Expression {
             result.push(' ');
             result.push(new SourceNode(this.asToken.range.start.line + 1, this.asToken.range.start.character, state.pathAbsolute, 'as'));
             result.push(' ');
-            result.push(new SourceNode(this.typeToken.range.start.line + 1, this.typeToken.range.start.character, state.pathAbsolute, this.typeToken.text));
+            result.push(util.bscTypeToTypeString(this.type));
```
TwitchBronBron, line 294: Still need a `SourceNode`. See this comment: https://github.com/rokucommunity/brighterscript/pull/262/files#r551867868

src/util.ts
```diff
@@ -1032,7 +1034,20 @@ export class Util {
...
+                if (allowCustomType) {
+                    return new CustomType(token.text);
+                }
+        }
+    }
+
+    /**
+    * Convert a BscType to a type string
+    */
+    public bscTypeToTypeString(bscType: BscType) {
```
TwitchBronBron, line 1046: I feel like this should exist on the type itself. Like `BscType.toTypeString()`. Most of the bsc types could just return `toString()`.

markwpearce:
I updated the PR with your comments - I also realized typings on class fields weren't being verified.. Not sure If i'm doing that correctly or not.... Namespaces are still not correctly checked though. :(

TwitchBronBron:
@elsassph I'd appreciate your thoughts on this PR.

src/DiagnosticMessages.ts
```diff
@@ -605,6 +605,16 @@ export let DiagnosticMessages = {
         message: `Missing exception expression after 'throw' keyword`,
         code: 1118,
         severity: DiagnosticSeverity.Error
+    }),
+    callableHasUnknownReturnType: (returnType: string) => ({
```
markwpearce, line 609: I'm not sure these new diagnostics need to be added - they're a lot like other ones ("invalidFunctionReturnType" and "expectedValidTypeToFollowAsKeyword")

elsassph: If possible yes reuse existing errors if they are fitting

src/validators/ClassValidator.ts
```diff
@@ -44,10 +45,11 @@ export class BsClassValidator {
         return cls;
     }
 
+
     /**
-     * Find all "new" statements in the program,
-     * and make sure we can find a class with that name
-     */
+ * Find all "new" statements in the program,
```
elsassph, line 50: indentation?

elsassph (commented):
Looking good!

markwpearce:
This now works with namespaces!


This would be valid Brighterscript code:

Vehicles.bs:
```vb
namespace Vehicles

  class Car
  end class

  class MonsterTruck
    function driveOver(automobile as Car) as Trash.SquashedCar
    end function
  end class

end namespace
```

Trash.bs
```vb
namespace Trash

  class SquashedCar
  end class

end namespace
```

src/Scope.spec.ts
```diff
@@ -484,6 +484,148 @@ describe('Scope', () => {
...
+                function foo() as MyNamespace.UnknownType
+                end function
+            `);
+            await program.validate();
+
+            expect(program.getDiagnostics().map(x => x.message)).to.eql([
+                DiagnosticMessages.invalidFunctionReturnType('MyNamespace.UnknownType').message
+            ]);
+        });
+
```
TwitchBronBron, line 628: Could you add a few tests where the custom type exists in a different file than the function referencing it? Like a test where it finds the custom type in a different file, and another test where it doesn't include the custom type because it's in a separate scope (you'll need to add an xml file with the script reference for this one).

TwitchBronBron: Also, a test where the custom type is defined in the parent xml component, and referenced in the child xml component.

src/validators/ClassValidator.ts
```diff
@@ -44,10 +45,11 @@ export class BsClassValidator {
         return cls;
     }
 
+
     /**
-     * Find all "new" statements in the program,
-     * and make sure we can find a class with that name
-     */
+    * Find all "new" statements in the program,
+    * and make sure we can find a class with that name
```
TwitchBronBron, line 51: The jsdoc comment style requires all of the stars to be at the same indentation level. The previous code was actually correct, so can you undo this chunk?

src/validators/ClassValidator.ts
```diff
@@ -234,6 +236,37 @@ export class BsClassValidator {
...
 
+
+    /**
+     * Check the types for fields, and validate they are valid types
+     */
+    private validateFieldTypes() {
+        for (let key in this.classes) {
+            let classStatement = this.classes[key];
+            classStatement.body.filter(statement => isClassFieldStatement(statement))
+                .forEach((statement) => {
```
TwitchBronBron, line 247: Could you refactor this to use a `for(let statement of statements)` loop instead of `Array.forEach`? I know this one is a little nit-picky, but `Array.forEach` is _slightly_ less efficient than a raw for loop, and ClassValidator is a very frequently used piece of code so we want to keep it as fast as possible.

markwpearce:
@TwitchBronBron - I think everything you brought up has been addressed.

PR #263, opened 2021-01-05
Function signature and intellisense improvements

georgejecook:
Enhances intellisense, definitions and signature behaviour. In doing so, also normalizes the getSignature LSP function to pass responsiblites to program, using file and position. Refactorings include removing pre-AST functions for signature and adding ASTObject<T> to provide file context for statements, which saves expensive lookups later on. Fixes go to definition and signatures for namespace functions, classes, and callfunc calls.

*Known issues: *

**completions: **

 - I appear to be getting too many classes back for new keyword completions - this was scoped to files in scope before, and has since gone awry.. 
 - Would be lovely to prioritize completions somehow, so that property values (generic case) move below m. and class member completions
 -  Appear to be getting namespaces that are out of scope - same problem as classes, I suspect
 
 **signature help**

- I forgot to do super case.. oops.

**linting**

- one lint error I dont' understand

Putting this in now and can improve as we go along.

### DIFF
--- a/src/Program.ts
+++ b/src/Program.ts
@@ -20,10 +20,11 @@ import type { ManifestValue } from './preprocessor/Manifest';
 import { parseManifest } from './preprocessor/Manifest';
 import { URI } from 'vscode-uri';
 import PluginInterface from './PluginInterface';
-import { isBrsFile, isXmlFile, isCallExpression, isCallfuncExpression, isNewExpression, isClassStatement, isClassMethodStatement, isFunctionStatement, isFunctionExpression, isDottedGetExpression } from './astUtils/reflection';
+import { isBrsFile, isXmlFile, isCallExpression, isCallfuncExpression, isNewExpression, isClassMethodStatement, isDottedGetExpression } from './astUtils/reflection';
 import { createVisitor, WalkMode } from './astUtils/visitors';
-import type { ClassStatement, FunctionStatement, Statement } from './parser/Statement';
-import type { CallExpression, CallfuncExpression, FunctionExpression, NewExpression } from './parser/Expression';
+import type { FunctionStatement, Statement } from './parser/Statement';
+import type { CallExpression, CallfuncExpression, NewExpression } from './parser/Expression';
+import { ParseMode } from './parser';
 const startOfSourcePkgPath = `source${path.sep}`;
 
 export interface SourceObj {
@@ -43,7 +44,7 @@ export interface SignatureInfoObj {
     signature: SignatureInformation;
 }
 
-export interface ASTObj<T> {
+export interface FileLink<T> {
     item: T;
     file: BrsFile;
 }
@@ -624,8 +625,8 @@ export class Program {
         return result;
     }
 
-    public getStatementsByName(name: string, originFile: BrsFile, namespaceName?: string): ASTObj<Statement>[] {
-        let results = new Map<Statement, ASTObj<Statement>>();
+    public getStatementsByName(name: string, originFile: BrsFile, namespaceName?: string): FileLink<Statement>[] {
+        let results = new Map<Statement, FileLink<Statement>>();
         const filesSearched = new Set<BrsFile>();
         let parseMode = originFile.getParseMode();
         let lowerNamespaceName = namespaceName?.toLowerCase();
@@ -637,7 +638,7 @@ export class Program {
                 }
                 filesSearched.add(file);
 
-                const statementHandler = (statement: FunctionStatement, parent: Statement) => {
+                const statementHandler = (statement: FunctionStatement) => {
                     let parentNamespaceName = statement.namespaceName?.getName(parseMode)?.toLowerCase();
                     if (statement.name.text.toLowerCase() === name && (!parentNamespaceName || parentNamespaceName === lowerNamespaceName)) {
                         if (!results.has(statement)) {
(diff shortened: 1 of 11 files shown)


elsassph:
Can you maybe detail the goals of this refactoring because I don't think it is the right approach. The way currently the editor gets the text token at the current position is better and simpler IMHO.

Additionally I'm afraid all these statement walking are highly inefficient and redundant compared to the references we already collect in the files.

src/Program.ts
```diff
@@ -705,42 +779,73 @@ export class Program {
...
-                await file.isReady();
+        await file.isReady();
 
-                const statementHandler = (statement: FunctionStatement | ClassMethodStatement) => {
-                    if (statement.getName(file.getParseMode()).toLowerCase() === callableName) {
-                        results.push(file.getSignatureHelp(statement));
+        const results = new Map<string, SignatureInfoObj>();
+        const statementHandler = (expr: CallExpression | CallfuncExpression | NewExpression) => {
+            if (util.rangeContains(expr.range, position)) {
+                if (isCallExpression(expr)) {
```
TwitchBronBron, line 765: So little of this statementhandler is shared between the different expressions, there's not much point in sharing the same function for all three. You should just add inline handlers in the `walk` function below instead, one for each of the specific expression types.

src/Program.ts
```diff
@@ -705,42 +779,73 @@ export class Program {
...
+                        sigHelp.index = expr.call.args.findIndex((e) => util.rangeContains(e.range, position));
+                        results.set(sigHelp.key, sigHelp);
+                    }
+                }
             }
-        }
+        };
 
-        return results;
+        file.parser.ast.walk(createVisitor({
```
TwitchBronBron, line 834: Honestly, I don't think you need to do a visitor here. You can loop over `parser.references` for these three things. that'll be WAY more efficient.

georgejecook:
> Can you maybe detail the goals of this refactoring because I don't think it is the right approach. The way currently the editor gets the text token at the current position is better and simpler IMHO.
> 
> Additionally I'm afraid all these statement walking are highly inefficient and redundant compared to the references we already collect in the files.

The goals are to un-break navigatoin, definitions, intellisense, and signature help, for brighterscript, which were all working for a year, and recently got broken. I've greatly improved it in this pr.

Applicaster is a HUGE brighterscript project (undoubtedly the biggest one on earth right now) and performance is fine. Optimize late, I say. Most of these codepaths are triggered by ide navigation, and won't be expensive for non bs projects anyow.

src/Program.ts
georgejecook: was following the patterns I saw. I agree - now that I have better knoweldge of the references, that this is a faster choice.. I can fix that later this week/weekend.

src/Program.ts
georgejecook: indeed, this one started out as a good idea and ended like this - good catch. thanks.

elsassph:
> The goals are to un-break navigatoin, definitions, intellisense, and signature help, for brighterscript, which were all working for a year, and recently got broken. I've greatly improved it in this pr.

But what broke concretely? Is there an issue detailing that?

TwitchBronBron:
> > The goals are to un-break navigatoin, definitions, intellisense, and signature help, for brighterscript, which were all working for a year, and recently got broken. I've greatly improved it in this pr.
> 
> But what broke concretely? Is there an issue detailing that?

[This PR](https://github.com/rokucommunity/brighterscript/pull/191) moved most of the remaining language features into the language server. Prior to that, many of the features like symbol navigation, completions, etc,.. were run in the vscode-brightscript-language extension code itself, and were very primitive. Some of the brighterscript-specific items stopped working. You can see some of the features that this PR fixes by looking at the tests that @georgejecook wrote in `src/Program.spec.ts`.

src/Program.ts
TwitchBronBron: So `.references` only has `NewExpression`s. We don't have ` we don't actually have function calls or callfuncs. Meaning this is currently the best approach for finding these, unless we update the parser to capture more things in `references.

src/Scope.ts
```diff
@@ -850,6 +862,52 @@ export class Scope {
...
+            };
+            const fieldStatementHandler = (s: ClassFieldStatement) => {
+                if (!results.has(s.name.text)) {
+                    results.set(s.name.text, {
+                        label: s.name.text,
+                        kind: CompletionItemKind.Field
+                    });
+                }
+            };
+            file.parser.ast.walk(createVisitor({
```
elsassph, line 891: Considering you have all the classes references, it would be better to iterate over the statements of the class object directly instead of using a walker which will walk everything in depth.

src/Program.ts
elsassph: I'd like to warn that walking all the expressions of a file like that is very costly - you don't even use the cancellation pattern to stop walking when you have found a match. It might be very bad for people with large files - they already suffer quite a bit just with the effort it takes for completion to parse the file's AST.

You should try to narrow what you walk: for brs code, just find the overall function the user is in, and for a class, find the class member the position is in - they you walk this element only.

georgejecook:
@TwitchBronBron I've spotted 2 more issues, using this branch today..
1. does not narrow definition for super methods (I suggest an enhancement pr following this one)
2. I'm seeing that sig completions don't work when the position is on a , or empty space in the signature.. this is a pain when typing coz the sig appears, then as you enter args, it disappears (Which is really noticable and irritating). I suggest that is fixed before this pr progresses - I didn't have a unit test that specifically calls that out, so it's not immediately obvious.

I can look at these at the weekend.

georgejecook:
Didn't get time to get onto those 2 items at the weekend. Just getting onto them now.

georgejecook:
I updated the pr to get rid of the outrageous walk for siganature help, and also add robustness to signature help with partially typed lines. 

I've added a boatload of tests, to ensure that the siganture help does not disappear while typing.

there's probs some more things we can do to improve the narrowing; but I'd rather get this in and revisit that, or see if someone else, like perhaps @markwpearce wants to look into that, given the methods for returning completions, sigs and definitions are now all updated to be bs namespace and class aware.

I would have created a vsix to share with others to kick the tyres; but I can't compile one for some reason.

src/Scope.ts
```diff
@@ -850,6 +862,52 @@ export class Scope {
...
+                ClassFieldStatement: fieldStatementHandler
+            }), {
+                walkMode: WalkMode.visitStatements
+            });
+
+        }
+        return results;
+    }
+
+    public getClassHieararchy(className: string) {
```
markwpearce, line 902: typo - Should be `getClassHierarchy`

georgejecook: unless you're a pirate!!!.. erm. yes. okay you got me.. thanks :)

georgejecook:
@TwitchBronBron I think this is good to go.

georgejecook:
I've addressed your guys feedback, and also improved the callfunc completions and signatures based on @elsassph 's xml work, so that it correctly narrows in that case.

I get the spurious fail that we see on build https://github.com/rokucommunity/brighterscript/pull/263/checks?check_run_id=1713397052 from time to time, too.. not sure what to make of it.

Master is merged, other than the spurious fail, this pr should be good to go.

PR #269, opened 2021-01-11
disable Array.forEach via eslint

TwitchBronBron:
Disallow `Array.forEach` in this codebase because it serves no benefit, and can introduce performance problems. 
(You can still use it in `*.spec.ts` files if you wish)

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -12,7 +12,8 @@ module.exports = {
     },
     plugins: [
         '@typescript-eslint',
-        'no-only-tests'
+        'no-only-tests',
+        'github'
     ],
     extends: [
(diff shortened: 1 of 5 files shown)


.eslintrc.js
```diff
@@ -188,6 +190,7 @@ module.exports = {
             '@typescript-eslint/no-unused-vars': 'off',
             '@typescript-eslint/no-unused-vars-experimental': 'off',
             '@typescript-eslint/dot-notation': 'off',
+            'github/array-foreach': 'off',
```
TwitchBronBron, line 193: I disabled the rule for unit tests, since micro-optimizing the tests doesn't really matter...

PR #270, opened 2021-01-13
Class annotation support

georgejecook:
Adds annotation support for classes, attaching any pending annotation…s that appear before a class keyword

@elsassph inspired (i.e goaded) me to add annotation support for maestro, perhaps knowing that this feature was missing and I would therefore have to write it :D Great delegation @elsassph 👏 

Hopefully this approach is okay - I'm sure we all concur that annotations are just as usefu on classes as other statements.

### DIFF
--- a/src/parser/Parser.spec.ts
+++ b/src/parser/Parser.spec.ts
@@ -726,7 +726,7 @@ describe('parser', () => {
             expect(fn.annotations[0].call).to.be.instanceof(CallExpression);
         });
 
-        it.only('attaches annotations to a class', () => {
+        it('attaches annotations to a class', () => {
             let { statements, diagnostics } = parse(`
             @meta1
             class MyClass
--- a/src/parser/Parser.spec.ts
+++ b/src/parser/Parser.spec.ts
@@ -733,12 +733,98 @@ describe('parser', () => {
                 function main()
                     print "hello"
                 end function
-                end class
+            end class
             `, ParseMode.BrighterScript);
             expect(diagnostics[0]?.message).not.to.exist;
             let cs = statements[0] as ClassStatement;
             expect(cs.annotations?.length).to.equal(1);
             expect(cs.annotations[0]).to.be.instanceof(AnnotationExpression);
+        });
+
+        it('attaches annotations to a class constructor', () => {
+            let { statements, diagnostics } = parse(`
+            class MyClass
+                @meta1
+                function new()
+                    print "hello"
+                end function
+                function methodA()
+                    print "hello"
+                end function
+            end class
+            `, ParseMode.BrighterScript);
+            expect(diagnostics[0]?.message).not.to.exist;
+            let cs = statements[0] as ClassStatement;
+            let stat = cs.body[0];
+            expect(stat.annotations?.length).to.equal(1);
+            expect(stat.annotations[0]).to.be.instanceof(AnnotationExpression);
+        });
+
+        it('attaches annotations to a class methods', () => {
+            let { statements, diagnostics } = parse(`
+            class MyClass
+                function new()
+                    print "hello"
+                end function
+                @meta1
+                function methodA()
+                    print "hello"
+                end function
+            end class
+            `, ParseMode.BrighterScript);
+            expect(diagnostics[0]?.message).not.to.exist;
+            let cs = statements[0] as ClassStatement;
+            let stat = cs.body[1];
+            expect(stat.annotations?.length).to.equal(1);
+            expect(stat.annotations[0]).to.be.instanceof(AnnotationExpression);
+        });
+        it('attaches annotations to a class methods, fields and constructor', () => {
+            let { statements, diagnostics } = parse(`
+            @meta2
+            @meta1
+            class MyClass
+                @meta3
+                @meta4
+                function new()
+                    print "hello"
+                end function
+                @meta5
+                @meta6
+                function methodA()
+                    print "hello"
+                end function
+
+                @meta5
+                @meta6
+                public foo="bar"
+            end class
+            `, ParseMode.BrighterScript);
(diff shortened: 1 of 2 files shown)


elsassph:
Good catch - though annotations shouldn't be added from insider the statement readers normally. I need to check.

georgejecook:
I also see that we're missing it for class methods (at least that seems to be the case):
image

If so, I can add that.

If there's a more proper way to implement this, let me know - it's late, and I quickly moved over maestro and rooibos, so I wasn't expecting to jump back into bsc right now, so just did what was expedient.

src/parser/Parser.ts
```diff
@@ -460,6 +463,12 @@ export class Parser {
             parentClassName,
             this.currentNamespaceName
         );
+
+        //attach annotations to statements
+        if (classAnnotations?.length) {
```
TwitchBronBron, line 468: While this technically works due to javascripts truthy/falsey, let's check for length > 0

src/parser/Parser.spec.ts
```diff
@@ -726,6 +726,21 @@ describe('parser', () => {
             expect(fn.annotations[0].call).to.be.instanceof(CallExpression);
         });
 
+        it('attaches annotations to a class', () => {
+            let { statements, diagnostics } = parse(`
+            @meta1
```
TwitchBronBron, line 731: Can you indent this stuff properly? Everything but the `end class` needs indented once more

TwitchBronBron (changes requested):
Looks simple enough to me. I have requested two small changes, but then it's good to go..

src/parser/Parser.ts
georgejecook: I got this from other code in the project.. I didn't like it; but I thought it was idiomatic

georgejecook:
Hot a couple of questions about this. @elsassph days this is working out the box as it stands. I didn't get that behaviour with the above unit test so I'm confused. 

Secondly I've been told that this should be implemented with refactoring the parser. Don't have time to do that. 

Can we just make it work for the time being for classes and class methods and do whatever refactoring later? The gesture is half baked as it is imho.

src/parser/Parser.ts
elsassph: Test should be 
```
if (classAnnotations.length > 0) {
```
and we can merge - I'll complete some other things.

georgejecook:
I think I've covered everything here - see tests. called out a line I'd like special revision on. all tests are passing, so I think we're okay.

src/parser/Parser.ts
```diff
@@ -407,17 +411,18 @@ export class Parser {
...
 
                     body.push(methodStatement);
 
                     //fields
                 } else if (this.checkAny(TokenKind.Identifier, ...AllowedProperties)) {
-                    body.push(
-                        this.classFieldDeclaration(accessModifier)
-                    );
+
+                    const fieldStatement = this.classFieldDeclaration(accessModifier);
```
elsassph, line 423: Good stuff. Also need to reset `this.pendingAnnotations` in this case.

src/parser/Parser.ts
```diff
@@ -407,17 +411,18 @@ export class Parser {
                         funcDeclaration.func,
                         overrideKeyword
                     );
-
+                    methodStatement.annotations = fieldAnnotations;
```
elsassph, line 412: Maybe move `this.pendingAnnotations = [];` here

src/parser/Parser.ts
```diff
@@ -387,8 +387,10 @@ export class Parser {
                     overrideKeyword = this.advance();
                 }
 
+                let fieldAnnotations = this.pendingAnnotations;
```
TwitchBronBron, line 390: Why do you grab the annotations here, but then clear them inside the if statement 2 lines below? I would expect the variable and the clear to happen at the same place.

georgejecook: we need to do this or block annotations will carry over onto class members

src/parser/Parser.ts
georgejecook: I don't think so.. we can only acrue them in blocks, which is delt with in the previous branch.

georgejecook:
I think this is good to go now. I addressed both of your feedback and merged current master.

src/parser/Parser.ts
elsassph: Hmm no if you don't consume it then it will be used by the next declaration isn't it?
Also why not use the `consumePendingAnnotation` function.

georgejecook: good calll - this reads better, like @TwitchBronBron wanted, too if I do that. thanks.

elsassph: Why don't you do the same for methods?

georgejecook: same reason as classes
image

georgejecook: now I think about it - we should have cleared it out there, too..

georgejecook: pushed up a  fix for it.

elsassph: Ugh look I think this whole approach kind of works but isn't the right way to go.
The problem is that basically annotations need to be saved/restored every time a "block" or "body" is entered, instead of this adhock redundant logic.

This logic will fail as well for a field with a `sub` as initial value 

Oh and it was merged. Well... I'll need to refactor that.

georgejecook: It's quite an edge case though. I think annotations broadly working on classes except for one edge case is better than totally broken annotations on classes. 

As you say, we can always fix the broken edge case in future, and if needs be adopt a different approach at that juncture. You should file a bug for the field sub case. Hadnt even dawned on me that someone would ever do that (surely a field defined as a sub is just a function, so it's the edgiest of edge cases one could contrive - I doubt we'll ever see that bug reported).

PR #274, opened 2021-01-17
Improve/fix annotations parsing

elsassph:
Annotations should be block-restricted:
- save/restore when entering/exiting a block
- error if an annotation isn't used within a block


TwitchBronBron:
Looks good. I like the improvements, thanks for tidying this up!

PR #278, opened 2021-01-18
Remove async file adding from program

TwitchBronBron:
Since the xml parser change removed the need for any async parsing, there's no benefit to having async file adding logic at all in the Program and downstream. Notable changes:
 - remove all unnecessary async functionality from `Program`, `BrsFile`, `XmlFile`. 
 - The `ProgramBuilder` still uses async logic when loading files because it yields a lot of performance benefits.
 - Moved `fileResolver` logic out of `Program` and into `ProgramBuilder` so the language server can continue to use the in-memory `Document` stores
 - Changed `fileResolver` into an object with `readFile` and `readFileSync` functions so we can use the proper function when needed.
 - renamed `fileExists` to `pathExists` and `pathExistsSync` to line up with the ones from `fs-extra`

I broke the commits up so you can see just the raw api changes in [this commit](https://github.com/rokucommunity/brighterscript/commit/5ef936f95e75c9c309dde193945ae779884cef32), and then all the unit tests in [this commit](https://github.com/rokucommunity/brighterscript/commit/92776895fe1e5cf4751816e6a690e1e3c4f080e0).

### DIFF
--- a/src/LanguageServer.ts
+++ b/src/LanguageServer.ts
@@ -335,7 +335,7 @@ export class LanguageServer {
         //if there's a setting, we need to find the file or show error if it can't be found
         if (config?.configFile) {
             configFilePath = path.resolve(workspacePath, config.configFile);
-            if (await util.fileExists(configFilePath)) {
+            if (await util.pathExists(configFilePath)) {
                 return configFilePath;
             } else {
                 this.sendCriticalFailure(`Cannot find config file specified in user/workspace settings at '${configFilePath}'`);
@@ -344,13 +344,13 @@ export class LanguageServer {
 
         //default to config file path found in the root of the workspace
         configFilePath = path.resolve(workspacePath, 'bsconfig.json');
-        if (await util.fileExists(configFilePath)) {
+        if (await util.pathExists(configFilePath)) {
             return configFilePath;
         }
 
         //look for the deprecated `brsconfig.json` file
         configFilePath = path.resolve(workspacePath, 'brsconfig.json');
-        if (await util.fileExists(configFilePath)) {
+        if (await util.pathExists(configFilePath)) {
             return configFilePath;
         }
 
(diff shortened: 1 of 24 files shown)


src/LanguageServer.ts
```diff
@@ -816,11 +801,14 @@ export class LanguageServer {
         } else if (program.hasFile(change.pathAbsolute)) {
             //sometimes "changed" events are emitted on files that were actually deleted,
             //so determine file existance and act accordingly
-            if (await util.fileExists(change.pathAbsolute)) {
-                await program.addOrReplaceFile({
-                    src: change.pathAbsolute,
-                    dest: rokuDeploy.getDestPath(change.pathAbsolute, options.files, rootDir)
-                });
+            if (await util.pathExists(change.pathAbsolute)) {
```
elsassph, line 804: I realize: why do you test if the file physically exists instead of letting the file resolver(s) try to find it? Otherwise what's the point to the list of resolvers?

elsassph (commented):
Generally fine but added couple of questions.

src/LanguageServer.ts
```diff
@@ -371,16 +371,17 @@ export class LanguageServer {
         builder.allowConsoleClearing = false;
 
         //look for files in our in-memory cache before going to the file system
-        builder.addFileResolver((pathAbsolute) => {
-            return this.documentFileResolver(pathAbsolute);
+        builder.addFileResolver({
+            readFile: (pathAbsolute) => this.documentFileResolver(pathAbsolute),
+            readFileSync: (pathAbsolute) => this.documentFileResolver(pathAbsolute)
```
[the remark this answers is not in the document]
TwitchBronBron: The language server protocol has a concept called TextDocument. Those objects are in-memory copies of the files. Basically every time you open a file, type in the file, etc, it gets streamed to the language server automatically and exists in memory. So there's nothing async to do with those.

src/LanguageServer.ts
TwitchBronBron: The file resolver concept was created to support the language server's TextDocument design. Generally you'll only have two resolvers: the language server, and then the actual file system. We read the language server first, and if not there, go to the file system.

Since our language server has a separate process to handle not-saved transient files (I call them standalone files), we can safely assume that all other files are physically located on disk, regardless of whether they are currently opened in the editor or not, so this check works just fine.

Unless I'm missing something, I don't see any need to modify the file resolvers to support doesFileExist logic since it doesn't seem to provide any benefit. Do you?

elsassph: Ok makes sense.

elsassph (approved):

PR #280, opened 2021-01-20
Fix xml missing script type

TwitchBronBron:
Transpiling an xml file that did not have a script type would cause node runtime errors, so this fixes that. Additional changes:
 - Add `type="text/brightscript" attribute if missing from a script tag
 - Add `clone` method to `SGScript` to do a better job deep cloning


src/parser/SGTypes.ts
```diff
@@ -161,6 +161,21 @@ export class SGScript extends SGTag {
             return super.transpileBody(state);
         }
     }
+
+    public clone() {
+        return new SGScript(
+            { ...this.tag },
+            this.attributes.map(a => {
```
elsassph, line 168: You sure? It means the original tag/attributes sourcemap will be lost.
Aside from that it's cool.

TwitchBronBron: How would it lose the sourcemap? I'm cloning the range of each attribute as well, so shouldn't the sourcemap generation logic use those ranges when creating each attribute? 

Also, in `updateScript` you were returning a new SGScript without any of the original information previously...but with my clone method it should actually look more like the original script tag now.
image

elsassph: Hum oh right ok. I guess the question is whether a full clone is useful at all VS reusing objects, as it really happens during transcoding and the modified node isn't kept afterwards.
But it's up to you - it's fine.

src/parser/SGTypes.ts
```diff
@@ -419,13 +434,21 @@ export class SGAst {
     }
 
     private updateScript(script: SGScript): SGScript {
+        let clone: SGScript;
```
elsassph, line 437: So do you want to keep that or change the logic to the node's `transpile` stage?

TwitchBronBron: I'm going to change this to node's `transpile` stage. I'm implementing the override of `transpileTags` concept that we discussed in Slack.

TwitchBronBron:
@elsassph ok, I think this is ready for another review. I made a few other changes, so here's everything:
- Don't crash during transpile when script `type` is missing
- During transpile, Inject `type="brightscript"` on script tags when missing
- Move SGTranspileState creation out of the `SGAst` and into the `XmlFile`, so it's consistent with how `BrsFile` works.
- Instead of cloning the extra scripts, I inject them into the AST before transpile, and remove them afterwords. 
- Update `XmlFile` transpiling to honor the `options.sourceMap` property (i.e. if source maps are disabled, don't emit sourceMaps and don't write the comment)
- move `simpleMap` function into util, and use that within `BrsFile` too to reduce code duplication

src/files/XmlFile.ts
```diff
@@ -444,14 +446,45 @@ export class XmlFile {
...
+
+        const extraImportScripts = this.getMissingImportsForTranspile().map(uri => {
+            const script = new SGScript();
+            script.uri = util.getRokuPkgPath(uri.replace(/\.bs$/, '.brs'));
+            return script;
+        });
+
+        const extraImportIdx = this.ast.component.scripts.length - 1;
+        //temporarily add the missing imports as script tags
+        this.ast.component.scripts.push(...extraImportScripts);
```
elsassph, line 459: What about adding an internal "extraImports` property to the scripts tag and letting it transpile it?

src/parser/SGTypes.ts
```diff
@@ -161,6 +179,45 @@ export class SGScript extends SGTag {
             return super.transpileBody(state);
         }
     }
+
+    protected transpileAttributes(state: SGTranspileState): (string | SourceNode)[] {
```
elsassph, line 183: It might be better to have
```
    protected transpileAttributes(state: SGTranspileState, attributes: SGAttribute[]): (string | SourceNode)[] {
```
so the override could create a modified array in place and call the same super method.

elsassph (commented):
Nice, some suggestions

src/parser/SGTypes.ts
TwitchBronBron: Good idea. I did that.

src/files/XmlFile.ts
TwitchBronBron: I don't like this idea. It's easy enough to add and remove the scripts from the `component.scripts` array. We would end up with only 1 less line of code here if we used `extraImports`, and it seems like this specific logic doesn't really belong in the AST anyway. It doesn't help plugins either, because if more than one plugin wanted to add `extraImports`, then they'd be right back to adding/removing and tracking indexes, so might as well just keep that all in the `component.scripts` collection.

elsassph (approved):

PR #285, opened 2021-01-22
catch errors when getting signatures

georgejecook:
Ive seen a few times where I've had errors on getting signature completions. I dont' have time to work out what they are now (they're related to typing fast, with @. afaict); so this is a stop gap.

### DIFF
--- a/src/LanguageServer.ts
+++ b/src/LanguageServer.ts
@@ -963,21 +963,25 @@ export class LanguageServer {
         const filepath = util.uriToPath(params.textDocument.uri);
         await this.keyedThrottler.onIdleOnce(filepath, true);
 
-        const signatures = util.flatMap(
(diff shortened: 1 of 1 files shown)


src/LanguageServer.ts
```diff
@@ -963,21 +963,25 @@ export class LanguageServer {
...
+            let results: SignatureHelp = {
+                signatures: signatures.map((s) => s.signature),
+                activeSignature: activeSignature,
+                activeParameter: activeParameter
+            };
 
-        return results;
+            return results;
+        } catch (e) {
+            return undefined;
```
TwitchBronBron, line 983: This completely hides the error. We should at least log it. Something to the effect of:

```
 this.connection.console.log(e);
 ```

georgejecook:
@TwitchBronBron hopefully this is okay now - I might have even fixed the issue with this latest change.

Issue #287, opened 2021-01-23
annotations do not parse negative ints


georgejecook:
```
 it.only('params test with negative numbers', () => {
            program.addOrReplaceFile('source/test.spec.bs', `
                @suite
                class ATest
                    @describe("groupA")
                    @it("is test1")
                    @params(100,100)
                    @params(100,-100)
                    @params(-100,100)
                    @params(-100,-100)
                    function Test_3(a, b)
                    end function
                end class
            `);
            program.validate();
            expect(program.getDiagnostics()).to.be.empty;
            expect(plugin.session.sessionInfo.testSuitesToRun).to.not.be.empty;
        });
```

in this scenario, all th -100's turn into null, because the call expression arg is a unary, not a literal expression.

georgejecook:
I've fixed this..
I'll get a pr up some time tomorrow, with an imrpoved unit test (this came from my rooibos plugin, where Ithought the problem originated from )..

fix is this; so if you concur, please don't wate time on it - I only need a few mins to move my test over.
![image](https://user-images.githubusercontent.com/1214545/105611993-308fde80-5db9-11eb-9f61-afa2c0a248a4.png)

TwitchBronBron:
This was fixed in #288

PR #288, opened 2021-01-23
adds support for negative numbers in annotation args

georgejecook:
fix for #287


TwitchBronBron (approved):
Looks good to me, but I want the opinion of @elsassph before I merge since he wrote most of the annotation logic.

elsassph (approved):
Is that in fact a defect in parsing function calls in general?

PR #292, opened 2021-01-29
Better handling for `createStringLiteral`

TwitchBronBron:
the `createStringLiteral` function was not wrapping the value in quotes, which was unintuitive since it's not obvious that TokenKind.StringLiteral's value should be wrapped in quotes. This PR wraps the value if the quotes are missing, or leaves the value alone if at least one quotemark is found.

### DIFF
--- a/src/astUtils/creators.ts
+++ b/src/astUtils/creators.ts
@@ -46,7 +46,7 @@ export function createDottedIdentifier(path: string[], range?: Range, namespaceN
  */
 export function createStringLiteral(value: string, range?: Range) {
     //wrap the value in double quotes
-    if (value.startsWith('"') === false && value.endsWith('"') === false) {
+    if (!value.startsWith('"') && !value.endsWith('"')) {
         value = '"' + value + '"';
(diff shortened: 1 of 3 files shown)


src/astUtils/creators.ts
```diff
@@ -39,7 +39,16 @@ export function createDottedIdentifier(path: string[], range?: Range, namespaceN
...
 }
 
+/**
+ * Create a StringLiteralExpression. The TokenKind.StringLiteral token value includes the leading and trailing doublequote during lexing.
+ * Since brightscript doesn't support strings with quotes in them, we can safely auto-detect and wrap the value in quotes in this function.
+ * @param value - the string value. (value will be wrapped in quotes if they are missing)
+ */
 export function createStringLiteral(value: string, range?: Range) {
+    //wrap the value in double quotes
+    if (value.startsWith('"') === false && value.endsWith('"') === false) {
```
elsassph, line 49: `!value.startsWith('"')` should do :D

PR #297, opened 2021-02-02
Fix validate crash

TwitchBronBron:
Found another situation that causes the language server to show errors.

### DIFF
--- a/src/LanguageServer.ts
+++ b/src/LanguageServer.ts
@@ -903,8 +903,8 @@ export class LanguageServer {
 
             await this.sendDiagnostics();
         } catch (e) {
-            this.connection.tracer.log(e);
-            this.sendCriticalFailure(`Critical error validating workspace: ${e.message}`);
+            this.connection.console.error(e);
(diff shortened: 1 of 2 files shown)


src/LanguageServer.ts
```diff
@@ -903,8 +903,8 @@ export class LanguageServer {
 
             await this.sendDiagnostics();
         } catch (e) {
-            this.connection.tracer.log(e);
-            this.sendCriticalFailure(`Critical error validating workspace: ${e.message}`);
+            this.connection.console.error(e);
```
TwitchBronBron, line 906: The tracer log wasn't showing up in the output panel, but `this.connection.console.error` does. Not relevant to the fix, but useful to have overall.

elsassph (approved):
Nice one... Parsers with recovery can really produce wonky AST 😬

src/LanguageServer.ts
TwitchBronBron: > Nice one... Parsers with recovery can really produce wonky AST 😬

Yeah definitely. I think I introduced this bug when I added the for loop item to completions.

PR #298, opened 2021-02-02
Add codeAction support

TwitchBronBron:
This adds the ability to send code actions from the language server to the editor. 

I created a helper (`util.createCodeAction`) to simplify creating code actions (because they're kind of annoying to create.

**TODO**
- [x]  support scope-based code actions


src/Program.ts
```diff
@@ -733,6 +733,26 @@ export class Program {
...
+        this.plugins.emit('beforeGetCodeActions', file, range, codeActions);
+
+        //get code actions from the file
+        codeActions.push(
+            ...file.getCodeActions(range)
+        );
+
+        //TODO get code actions from each scope
+
+        this.plugins.emit('beforeGetCodeActions', file, range, codeActions);
```
elsassph, line 752: "after"?

TwitchBronBron: Oops. Yep! Copypasta.

src/DiagnosticMessages.ts
```diff
@@ -17,6 +17,7 @@ export let DiagnosticMessages = {
     callToUnknownFunction: (name: string, scopeName: string) => ({
         message: `Cannot find function with name '${name}' when this file is included in scope '${scopeName}'`,
         code: 1001,
+        functionName: name,
```
TwitchBronBron, line 20: This feels a little like a hack, but it's either this or I scan the diagnostic message with regex to find the function name.

src/DiagnosticMessages.ts
```diff
@@ -608,13 +609,13 @@ export let DiagnosticMessages = {
     })
 };
 
-let allCodes = [] as number[];
+export const DiagnosticCodeMap = {} as Record<keyof (typeof DiagnosticMessages), number>;
```
TwitchBronBron, line 612: An easier way to look up diagnostic codes without having to call the DiagnosticMessages function

src/XmlScope.ts
```diff
@@ -153,6 +153,46 @@ export class XmlScope extends Scope {
...
+    public getCodeActions(file: BscFile, range: Range) {
+        const result = [] as CodeAction[];
+        for (const diagnostic of this.diagnostics) {
+            //skip diagnostics that don't occur on this line
+            if (diagnostic.range?.start.line !== range.start.line) {
+                continue;
+            }
+            if (diagnostic.code === DiagnosticCodeMap.callToUnknownFunction) {
+                //functionName is stored on this specific diagnostic
+                const lowerFunctionName = (diagnostic as any).functionName.toLowerCase();
```
TwitchBronBron, line 165: I feel like these will get quite large pretty quickly, so it might be useful to extract it into some type of codeAction-specific helper file. Thoughts?

src/util.ts
```diff
@@ -1175,6 +1176,28 @@ export class Util {
             range: attr.range
         } as SGAttribute;
     }
+
+    public createCodeAction(obj: CodeActionShorthand) {
```
TwitchBronBron, line 1180: Creating codeActions that actually work are actually kind of hard. I couldn't get the interface items to work, so I had to call `TextEdit.insert`, `TextEdit.replace`, `CodeAction.create`, `URI.file()`, etc. So this function simplifies that process by accepting a more simple structured object. The plugins can return their own `CodeAction` objects if they want, but they might as well just use this function.

src/DiagnosticMessages.ts
elsassph: If we want to add a payload to diagnostics we need something more generic.

src/Program.ts
```diff
@@ -733,6 +733,36 @@ export class Program {
...
+
+        //get code actions from the file
+        codeActions.push(
+            ...file.getCodeActions(range)
+        );
+
+        //get code actions from every scope this file is a member of
+        for (let key in this.scopes) {
```
elsassph, line 751: `for of` maybe?

src/XmlScope.ts
elsassph: I definitely don't approve the pattern of adding code-actions logic inside Scope/BrsFile/XmlFile. If anything we need less logic (including validation) in those files.

Logic should be entirely decoupled from these classes, and possibly even moved into an internal plugin.

src/XmlScope.ts
```diff
@@ -153,6 +153,46 @@ export class XmlScope extends Scope {
         });
     }
 
+    public getCodeActions(file: BscFile, range: Range) {
+        const result = [] as CodeAction[];
+        for (const diagnostic of this.diagnostics) {
```
elsassph, line 158: This filtering of the diagnostics should be done once at the reception of the code actions, otherwise you'll end up with duplicated logic of looking up for diagnostics matching the range.

src/files/XmlFile.ts
```diff
@@ -369,6 +370,38 @@ export class XmlFile {
...
+        for (const diagnostic of this.diagnostics) {
+            //skip diagnostics that don't occur on this line
+            if (diagnostic.range?.start.line !== range.start.line) {
+                continue;
+            }
+            if (diagnostic.code === DiagnosticMessages.xmlComponentMissingExtendsAttribute().code) {
+                //add the attribute at the end of the first attribute, or after the `<component` if no attributes
+                const pos = (this.parser.ast.component.attributes[0] ?? this.parser.ast.component.tag).range.end;
+                //add this extends attribute right after the closing `>` token
+                range.end.character -= 1;
```
elsassph, line 385: Don't mutate input data.

src/parser/SGTypes.ts
```diff
@@ -359,6 +360,7 @@ export class SGInterface extends SGTag {
 
 export class SGComponent extends SGTag {
     constructor(
+        public ast: ElementCtx,
```
elsassph, line 363: Don't leak the internal XML parser ast; what information are you missing and how can it be defined in a generically useful way? Is it maybe the full range, including the open slash?

src/util.ts
elsassph: I like it.

elsassph (changes requested):

src/Program.ts
TwitchBronBron: No, `this.scopes` is `Record<string, Scope>` so we need `for in` to iterate over the keys.

src/files/XmlFile.ts
TwitchBronBron: good catch. I wasn't actually even using this anymore. Removed.

src/XmlScope.ts
TwitchBronBron: Done. moved into `bscPlugin/BscPlugin.ts`

src/XmlScope.ts
TwitchBronBron: Done.

src/XmlScope.ts
georgejecook: I know you 2 have a different opinion; but I'm of the opposite school of thought. I think Roku app are small and this language will only ever be for Roku apps so my opinion is augment all the things early and thoroughly so as to decorate the way with as much info as possible to facilitate its purpose, which surely is transpiling be and supporting the ide. 

I believe Bron has some clever ideas about how to do this that won't incur the tax of supporting a shadow-ast or the cost (perf and conceptionally) of synchronizing the ide specific metadata with the ast. 

But I'm still staunchly of the opposite opinion.  What do we want? Ide features and diagnostics? When do we want them? Now!! :) Ie get as much as we can as fast as we can. 

Sharing this for your benefit @elsassph as Bron knows my views on this already. I'm very solid in the idea that we should never forget what we are building. These apps will never ever be more than a few mb big. That's my ocams razor right there. You 2 will ignore me, of course; but at least you know where I'm coming from.

elsassph: Not sure what you are referring to @georgejecook. Is it about my suggestion to move this logic into an internal plugin?

georgejecook: >I definitely don't approve the pattern of adding code-actions logic inside Scope/BrsFile/XmlFile

georgejecook: I personally say, if it's clean and manageable, put it wherever gives us the most info, without the need for more parses, or complex data structure sync.

elsassph: Using an internal plugin won't require more data or ast duplication and stuff.

elsassph: On the other hand the amount of code in BrsFile is getting our of control.

src/Program.ts
```diff
@@ -69,6 +70,10 @@ export class Program {
         this.logger = logger || new Logger(options.logLevel as LogLevel);
         this.plugins = plugins || new PluginInterface([], undefined);
 
+        //inject the bsc plugin as the first plugin in the stack.
+        //eslint-disable-next-line @typescript-eslint/dot-notation
+        this.plugins['plugins'].unshift(new BscPlugin());
```
elsassph, line 75: Shouldn't we have some "add first" method in `PluginInterface` instead?

src/bscPlugin/codeActions/XmlFileCodeActionsProcessor.ts
```diff
@@ -0,0 +1,44 @@
+import type { CodeAction, Range } from 'vscode-languageserver';
+import { CodeActionKind } from 'vscode-languageserver';
+import { DiagnosticCodeMap } from '../../DiagnosticMessages';
+import type { XmlFile } from '../../files/XmlFile';
+import type { BsDiagnostic } from '../../interfaces';
+import util from '../../util';
+
+export class XmlFileCodeActionsProcessor {
```
elsassph, line 8: You really do like classes ;)

src/bscPlugin/codeActions/XmlFileCodeActionsProcessor.ts
```diff
@@ -0,0 +1,44 @@
...
+                this.addMissingExtends();
+            }
+        }
+    }
+
+    private addMissingExtends() {
+        //add the attribute at the end of the first attribute, or after the `<component` if no attributes
+        const pos = (this.file.parser.ast.component.attributes[0] ?? this.file.parser.ast.component.tag).range.end;
+        this.codeActions.push(
+            util.createCodeAction({
```
elsassph, line 30: Would be nice to offer a few options like `Group`, `Task` and `ContentNode`.

src/bscPlugin/codeActions/XmlFileCodeActionsProcessor.ts
```diff
@@ -0,0 +1,44 @@
...
+        for (const diagnostic of this.diagnostics) {
+            if (diagnostic.code === DiagnosticCodeMap.xmlComponentMissingExtendsAttribute) {
+                this.addMissingExtends();
+            }
+        }
+    }
+
+    private addMissingExtends() {
+        //add the attribute at the end of the first attribute, or after the `<component` if no attributes
+        const pos = (this.file.parser.ast.component.attributes[0] ?? this.file.parser.ast.component.tag).range.end;
```
elsassph, line 28: I'm not sure the position is correct for a tag without any attribute. You probably should just always insert at the end of the tag instead of trying to find the first attribute.

Also maybe have some `const { component } = this.file.parser.ast;` for readability.

src/bscPlugin/codeActions/XmlScopeCodeActionsProcessor.ts
```diff
@@ -0,0 +1,58 @@
...
+        //functionName is stored on this specific diagnostic
+        const lowerFunctionName = (diagnostic as any).functionName.toLowerCase();
+
+        //find every file with this function defined
+        for (const key in this.scope.program.files) {
+            const file = this.scope.program.files[key];
+            if (isBrsFile(file)) {
+                //TODO handle namespace-relative function calls
+                const stmt = file.parser.references.functionStatementLookup.get(lowerFunctionName);
+                const slashOpenToken = this.scope.xmlFile.parser.ast.component?.ast.SLASH_OPEN?.[0];
```
elsassph, line 38: I'm still opposed to peeking at the internal AST :)

We should have a generic "add script" codeaction generator (might start a codeaction utils), which would try to append a script after the last script tags, then try to add it after the interface, and then try to add it at the beginning of the component's body

src/bscPlugin/codeActions/XmlScopeCodeActionsProcessor.ts
```diff
@@ -0,0 +1,58 @@
...
+                if (stmt && slashOpenToken) {
+                    const pkgPath = util.getRokuPkgPath(file.pkgPath);
+                    this.codeActions.push(
+                        util.createCodeAction({
+                            title: `Import "${pkgPath}" into component "${this.scope.xmlFile.componentName.text ?? this.scope.name}"`,
+                            // diagnostics: [diagnostic]
+                            changes: [{
+                                filePath: this.scope.xmlFile.pathAbsolute,
+                                newText: `  <script type="text/brightscript" uri="${pkgPath}" />\n`,
```
elsassph, line 47: Does it honor the file's indentation? Surely that should be possible to get he IDE do automatically.
Would be worth having an `insertLine` action where you only provide the line number.

src/interfaces.ts
```diff
@@ -199,6 +199,13 @@ export interface CompilerPlugin {
     afterFileTranspile?: (entry: TranspileObj) => void;
     beforeFileDispose?: (file: BscFile) => void;
     afterFileDispose?: (file: BscFile) => void;
+
```
elsassph, line 202: Events are sorted by target, program events, scope events, file events

src/XmlScope.ts
```diff
@@ -153,6 +153,11 @@ export class XmlScope extends Scope {
         });
     }
 
+    public getCodeActions(file: BscFile, range: Range, codeActions: CodeAction[]) {
+        const relevantDiagnostics = this.diagnostics.filter(x => x.range?.start.line === range.start.line);
+        this.program.plugins.emit('onScopeGetCodeActions', this, file, range, relevantDiagnostics, codeActions);
```
elsassph, line 158: Should we avoid emitting the scope event if there is no diagnostic match?

src/Program.ts
TwitchBronBron: I want the BscPlugin to ALWAYS be first in the chain, an `addFirst` method would make it easier for external users to mess with the order of things. I don't see a reason for an `addFirst` method _except_ for what we are doing here...and since I'm using the index access pattern we will get tsc errors if the internals of `PluginInterface` ever change.

src/bscPlugin/codeActions/XmlFileCodeActionsProcessor.ts
TwitchBronBron: Yep! :D Methods on direct module exports are much more annoying to mock/stub in unit tests. Also, for this specific set of classes, it vastly simplifies the individual method parameters because the constructor receives and maintains the state.

src/bscPlugin/codeActions/XmlFileCodeActionsProcessor.ts
TwitchBronBron: Nice suggestion. Added!

src/bscPlugin/codeActions/XmlScopeCodeActionsProcessor.ts
TwitchBronBron: I did it this way so we don't occasionally insert two consecutive whitespaces. It's easier to reason about the whitespace after an attribute. Although I do agree that we should add this to the very end of the tag, so I adjusted this to find the FINAL attribute first, and then after the `<component` object second if there are no attributes

src/bscPlugin/codeActions/XmlFileCodeActionsProcessor.ts
TwitchBronBron: I did it this way so we don't occasionally insert two consecutive whitespaces. It's easier to reason about the whitespace after an attribute. Although I do agree that we should add this to the very end of the tag, so I adjusted this to find the FINAL attribute first, and then after the `<component` object second if there are no attributes

src/XmlScope.ts
TwitchBronBron: I think we should emit the event no matter what. If we are sticking with "there can be code actions that are not attached to diagnostics", then skipping scopes with no diagnostics would be detrimental to plugins that want to follow that pattern. 

We are already only emitting the event for scopes who have the associated file, so I think it should be fine to emit the event for all scopes for that file. 

For example, assume there's some MVVM plugin that does the following:
```
    <Button on:click="vm.newValue = 1 + 2 + 3" />
```

There could be a code action `Refactor to view model method` for within the `on:click`, which transforms to something like this:
```
<button on:click="vm.clickButton()"
<script>
    vm.clickButton = function()
         vm.newValue = 1 + 2 + 3
    end function
</script>
```

Now, an argument can still be made that there should be a "hint" diagnostic there to begin with, but that's what we need to decide. **Should brighterscript require codeActions to be tied to a diagnostic?**

src/bscPlugin/codeActions/XmlScopeCodeActionsProcessor.ts
TwitchBronBron: No, it doesn't currently honor the file's indentation. I agree, this needs some more work. I'll work on it.

src/interfaces.ts
TwitchBronBron: Ok. I added comments in the interface explaining that, and moved the events to the proper locations.

src/parser/SGTypes.ts
TwitchBronBron: With the brs parser AST, I have location information for almost every token. I want that same information available with the XML AST. It seems wasteful to pipe all of the tokens from the underlying parser into this, when we already have the ctx available. 

Ideally I would have the following info:
 - range for opening `<`
 - range for opening tag tag name
 - range for full attribute (i.e. `|someAttr="someval"|`)
 - range for attribute value (i.e. `someAttr="|someval|")
 - range for closing ">" in the opening tag
 - range for opening `<` in closing tag
 - range for closing tag name
 - range for closing `>` in closing tag

What practical benefit is there to hiding the internal XML parser AST? Is this to guard against the possible future change to a different xml parser?

elsassph: I really think the internal AST should not be used. It could be replaced, and it's unreliable.

We could have a few more ranges for XML tags, like the open and close tags, including angle brakets, they could even be late-resolution based on the internal AST.

src/Program.ts
elsassph: A user plugin could use addFirst, and the compiler is always the last to add plugins.

src/bscPlugin/codeActions/XmlFileCodeActionsProcessor.ts
elsassph: Adding an attribute or a tag should be done with a helper function BTW, so it's testable and reusable.

src/XmlScope.ts
elsassph: Some code actions surely shouldn't be a "compilation cost" and be late evaluated if possible.

BTW the compiler should have a mode so that for pure compilation there would be no work done to feed the metadata for codeactions.

src/parser/SGTypes.ts
georgejecook: What do we propose to replace it with? Is it going to be more efficient that what we have now?

src/bscPlugin/codeActions/XmlFileCodeActionsProcessor.ts
georgejecook: Please: more classes!! It's hard enough to anything in a performant way as it is. There's a massive learning curve in getting up to speed as it is. More ooad will help mitigate that and get more contributions.

src/XmlScope.ts
georgejecook: Again, are we optmizing early here? What is the cost of these guards (coneptionally, and in terms of implementation)? It can't be overstated: Apps can never be bigger than 3mb. What else are we using the compiler for? Is it worth all the extra logic for a few 50ms on a non-ide bsc transpile? 

I think we're better off solving the speed and stability issues, and improving the api than imposing more constraints and adding more logic.

TwitchBronBron: > Some code actions surely shouldn't be a "compilation cost" and be late evaluated if possible.

This would definitely be a justification for NOT mandating a diagnostic for every code action (and therefore we must emit the event for all scopes for the file).  In that way, we don't need a special compilation mode then either.

if a plug-in wants to generate tons of code actions, they can just do it on demand when the ide asks for that specific range. That costs no extra compile time which IS something we should be guarding. 

So I'm inclined to agree with George's statement that all of this discussion around it to prevent 
few eventss from firing might be a bit of premature optimization. 

I will do a little testing to see what the performance impact of many scopes for a file would.hsve on code actions, and we can make a more educated decision based on that.

src/parser/SGTypes.ts
TwitchBronBron: @georgejecook we aren't proposing anything new at the moment. We are just designing our custom SceneGraph ast so it is resilient to these changes in the future. I agree with the idea. Once bsc lands at v1.0.0, we will be locked in to certain design decisions for a while, so I do want to have the conversations sooner rather than later about these types of things. 

@elsassph I like the idea of keeping the internal AST around, and exposing gettters. For on-demand access to the ranges I want. That's reasonable: it's efficient because we don't do unnecessary copying during parse, and it still gives us access to all the data from the original parser that's we might nee for any future getters that we don't expose right away. 

I'll go this route. Possibly in a separate PR since it's directly related to this PR.

src/Program.ts
TwitchBronBron: Fair enough. If a dev starts messing with the PluginInterface, it's their own fault.

src/bscPlugin/codeActions/XmlFileCodeActionsProcessor.ts
TwitchBronBron: Good point. I'll do that.

src/bscPlugin/codeActions/XmlScopeCodeActionsProcessor.ts
TwitchBronBron: Great idea. I'll work on that.

src/DiagnosticMessages.ts
TwitchBronBron: Yeah, good point. I'll implement that.

src/interfaces.ts
```diff
@@ -187,18 +188,24 @@ export interface CompilerPlugin {
...
     afterProgramTranspile?: (program: Program, entries: TranspileObj[]) => void;
+    beforeProgramGetCodeActions?: (program: Program, file: BscFile, range: Range, codeActions: CodeAction[]) => void;
+    afterProgramGetCodeActions?: (program: Program, file: BscFile, range: Range, codeActions: CodeAction[]) => void;
+    //scope events
     afterScopeCreate?: (scope: Scope) => void;
     beforeScopeDispose?: (scope: Scope) => void;
     afterScopeDispose?: (scope: Scope) => void;
     beforeScopeValidate?: ValidateHandler;
     afterScopeValidate?: ValidateHandler;
+    onScopeGetCodeActions?: (scope: Scope, file: BscFile, range: Range, diagnostics: BsDiagnostic[], codeActions: CodeAction[]) => void;
```
TwitchBronBron, line 199: @elsassph what are your thoughts on the name of these new events? I didn't see any benefit in having a `before` and `after`, so i figured `on` would be the pattern for this type of event.

src/XmlScope.ts
elsassph: I think I said that I'm fine with dispatching the events. Keep cool 😅

TwitchBronBron: All cool! 😁 Just thinking out loud. I appreciate all the feedback. You know me, I just want to find the best solution.

src/interfaces.ts
elsassph: Maybe without `Get` like `onScopeCodeActions`?
There definitely are a few `before/after` which feel artificial in general.

src/DiagnosticMessages.ts
```diff
@@ -623,15 +626,22 @@ export let DiagnosticMessages = {
...
     code: number;
     severity: DiagnosticSeverity;
 }
+
+/**
+ * Provides easy type support for the return value of any DiagnosticMessage function.
+ * The second type parameter is optional, but allows plugins to pass in their own
+ * DiagnosticMessages-like object in order to get the same type support
+ */
+export type DiagnosticMessageType<K extends keyof D, D extends Record<string, (...args: any) => any> = typeof DiagnosticMessages> = ReturnType<D[K]>;
```
TwitchBronBron, line 647: @elsassph I think you'll find this quite useful. It allows you to do this:
```
const diagnostic = someDiagnostic as DiagnosticMessageType<'callToUnknownFunction'>;
```
which will look at DiagnosticMessage.callToUnknownFunction(), and enforce type checking based on that return value. This is useful when paired with the new `BsDiagnostic.data` `any` object, so we can enforce stricter types when using those.

This also supports an optional second type parameter, which allows plugins to use the same functionality, assuming they follow the same DiagnosticMessages model in their own plugins. 
```
export const BsLintDiagnosticMessages = {
    someDiagnostic: ()=>({
        message: 'Some diagnostic',
        code: 12345,
        data: {
            isTrue: true
        }
    })
};

const typedDiagnostic= program.getDiagnostics()[0] as DiagnosticMessageTypes<'someDiagnostic', BsDiagnosticMessages>;

Comment (TwitchBronBron) [reply src/DiagnosticMessages.ts:647]: You can see it used [here](https://github.com/rokucommunity/brighterscript/pull/298/files#diff-1412e36a3e7918cf699241fb44ab227a45ce0b7d92f609b4d543021861252d93R28)

File: src/Program.ts (hunk 10 above)
Comment (TwitchBronBron) [reply src/Program.ts]: Done.

File: src/interfaces.ts (hunk 18 above)
Comment (TwitchBronBron) [reply src/interfaces.ts:203]: I kinda like the `get`. all of our other events have a verb in them, it feels out of place to have `onScopeCodeActions` without a verb. 

A weak second argument is that `onScopeCodeActions` is kind of ambiguous as to whether it means "get code actions" or "execute code actions". Not that there is an event for "execute", but I like how "get" makes it more explicit.

**Comment by TwitchBronBron** [comment]
@elsassph I really want to get this feature merged, but I do agree that we should not leak internal xml ast data. So, I have ripped out that specific codeAction. So please give this PR a final review, and approve if you have no further concerns. I'll continue working on those other comments in separate PRs.

File: src/LanguageServer.ts [hunk 20]
```diff
@@ -543,6 +550,22 @@ export class LanguageServer {
         return item;
     }
 
+    private async onCodeAction(params: CodeActionParams) {
+        //ensure programs are initialized
+        await this.waitAllProgramFirstRuns();
```
elsassph, line 555: These awaits will potentially accumulate code action requests - should there instead be a mechanism to ignore codeactions when the file isn't ready? We may "miss" some actions but just moving the cursor around will produce then again.

elsassph (approved):
Just a little concern about accumulating code action requests

src/DiagnosticMessages.ts
TwitchBronBron: I added a generic `data` property to diagnostic that can contain anything.

src/parser/SGTypes.ts
TwitchBronBron: I removed this codeAction for now so the PR can merge. I'm still working on this in another branch and will submit a future PR when it's ready.

src/bscPlugin/codeActions/XmlScopeCodeActionsProcessor.ts
TwitchBronBron: I'll do this in a separate PR.

src/bscPlugin/codeActions/XmlScopeCodeActionsProcessor.ts
TwitchBronBron: I'll handle this in a separate PR, since this code is no longer included in this PR

src/LanguageServer.ts
TwitchBronBron: The language server protocol might already account for this and prevent additional requests if the first one hasn't arrived yet. So I'll test that first, and if we DO get multiple requests at the same time, then I'll add some code to ignore code actions that arrive before the programs are finished with their first runs.

TwitchBronBron: I just tested, and I can't get these to queue up. There's a `waitAllProgramFirstRuns()` in the onInitialized, so the vscode doesn't even send codeAction events until that's resolved, so this line is really a "just in case" measure.

PR #300, opened 2021-02-03
demonstrates efficient namespace validation. [NEEDS TESTS]

georgejecook:
also has placeholders that could easily facilitate enum validation and subsitution.

I suggest that on some expressions we have a knownType?: Statement property, which we can populate at validate time, and then the transpiler can use as a hint (e.g. we could set the knownType to an EnumValueStatement, which the tranpsiler could substitute as transpile time)..

I don't have time to drive this home with unit tests right now; but I'm already using this to great effect, and will gladly get this feature in, if you guys concur with the approach.


TwitchBronBron:
Thanks for the POC. I think we're probably going to try implementing this without a full AST walk, but we'll know more once more of the type tracking features land in v1.

PR #304, opened 2021-02-05
fixes super.methodCall() not transpiling in all cases

### LINKED ISSUE #303: super method calls missing in transpile output if nested inside a sub block
e.g..

image

vs
image

here's a test case: excuse the lack of expected test - tried various formats of output but none match (we should add trim to that compare util imo)

```
        it.only('transpiles super when not the first statement', () => {
            testTranspile(`
                class Creature
                    sub new(name as string)
                    end sub
                    function sayHello(text)
                    ? text
                    end function
                end class

                class Duck extends Creature
                    override function sayHello(text)
                        text = "The duck says " + text
                        if text <> invalid
                            super.sayHello(text)
                        end if
                    end function
                end class
            `, ``, undefined, 'source/main.bs');
        });`
``

### PR DESCRIPTION
**Pull request by georgejecook**
fixes #303

### DIFF
--- a/src/files/BrsFile.Class.spec.ts
+++ b/src/files/BrsFile.Class.spec.ts
@@ -417,7 +417,7 @@ describe('BrsFile BrighterScript classes', () => {
             `, undefined, 'source/main.bs');
         });
 
-        it('transpiles super when not the first statement', () => {
+        it('transpiles super in nested blocks', () => {
             testTranspile(`
                 class Creature
                     sub new(name as string)
--- a/src/files/BrsFile.Class.spec.ts
+++ b/src/files/BrsFile.Class.spec.ts
@@ -417,6 +417,61 @@ describe('BrsFile BrighterScript classes', () => {
             `, undefined, 'source/main.bs');
         });
 
+        it('transpiles super when not the first statement', () => {
+            testTranspile(`
+                class Creature
+                    sub new(name as string)
+                    end sub
+                    function sayHello(text)
+                    ? text
+                    end function
+                end class
+
+                class Duck extends Creature
+                    override function sayHello(text)
+                        text = "The duck says " + text
+                        if text <> invalid
+                            super.sayHello(text)
+                        end if
+                    end function
+                end class
+            `, `
+        function __Creature_builder()
+            instance = {}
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: src/files/BrsFile.Class.spec.ts [hunk 1]
```diff
@@ -417,6 +417,61 @@ describe('BrsFile BrighterScript classes', () => {
...
+                class Duck extends Creature
+                    override function sayHello(text)
+                        text = "The duck says " + text
+                        if text <> invalid
+                            super.sayHello(text)
+                        end if
+                    end function
+                end class
+            `, `
+        function __Creature_builder()
```
TwitchBronBron, line 439: Fix the indentation for this entire string template so it matches the one above. Then pass 'trim,' instead of undefined as the third parameter to testTranspile

src/parser/Statement.ts
```diff
@@ -1604,7 +1604,7 @@ export class ClassMethodStatement extends FunctionStatement {
                 }
             }
         });
-        const walkOptions: WalkOptions = { walkMode: WalkMode.visitExpressions };
```
elsassph, line 1607: @TwitchBronBron I think the problem is that `visitExpressions` should also include the `walkStatements` flag.

TwitchBronBron: Ah, yeah I think you're right. The recursive one already does exactly that.

@georgejecook can you please undo this line, and instead change the definition of `visitExpressions` to include `InternalWalkOptions.walkStatements`

https://github.com/rokucommunity/brighterscript/blob/1a84bb55f786ceb8bfc218011556e669227a72fe/src/astUtils/visitors.ts#L183

TwitchBronBron: @elsassph I made the fix we discussed (adding walkStatements to visitExpressions). That fixed George's issue, however one of the visitors tests was failing. I'm not overly familiar with that test to know how to properly validate it, so could you give it a look and see why it's failing?

elsassph: uh odd

elsassph: Ok that's because this test expects `visitExpressions` to NOT walk statements :D
I'll fix that.

PR #310, opened 2021-02-11
Fix watcher bug on windows devices.

TwitchBronBron:
running the bsc cli in watch mode on a windows device is crashing on any file change. The issue was that certain terminals will force the drive letter to lower case. this fixes that.

### DIFF
--- a/src/ProgramBuilder.ts
+++ b/src/ProgramBuilder.ts
@@ -155,10 +155,19 @@ export class ProgramBuilder {
         }
     }
 
+    /**
+     * A handle for the watch mode interval that keeps the process alive.
+     * We need this so we can clear it if the builder is disposed
+     */
+    private watchInterval: NodeJS.Timer;
+
     public enableWatchMode() {
         this.watcher = new Watcher(this.options);
+        if (this.watchInterval) {
+            clearInterval(this.watchInterval);
+        }
         //keep the process alive indefinitely by setting an interval that runs once every 12 days
-        setInterval(() => { }, 1073741824);
+        this.watchInterval = setInterval(() => { }, 1073741824);
 
         //clear the console
         this.clearConsole();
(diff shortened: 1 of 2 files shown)


src/ProgramBuilder.ts
```diff
@@ -186,11 +194,15 @@ export class ProgramBuilder {
             if (event === 'add' || event === 'change') {
                 const fileObj = {
                     src: thePath,
-                    dest: rokuDeploy.getDestPath(thePath, this.program.options.files, this.rootDir)
+                    dest: rokuDeploy.getDestPath(
+                        thePath,
+                        this.program.options.files,
+                        //some shells will toTowerCase the drive letter, so do it to rootDir for consistency
+                        util.driveLetterToLower(this.rootDir)
```
TwitchBronBron, line 201: This is the core of this PR. The other stuff was just a few small tweaks to make the CLi a little more friendly.

src/ProgramBuilder.ts
```diff
@@ -176,8 +185,7 @@ export class ProgramBuilder {
         let debouncedRunOnce = debounce(async () => {
             this.logger.log('File change detected. Starting incremental compilation...');
             await this.runOnce();
-            let errorCount = this.getDiagnostics().length;
-            this.logger.log(`Found ${errorCount} errors. Watching for file changes.`);
+            this.logger.log(`Watching for file changes.`);
```
TwitchBronBron, line 188: The error count will be reported in the `runOnce` method so this was a duplicated message.

src/ProgramBuilder.ts
```diff
@@ -498,5 +510,8 @@ export class ProgramBuilder {
         if (this.program) {
             this.program.dispose?.();
         }
+        if (this.watchInterval) {
+            clearInterval(this.watchInterval);
```
TwitchBronBron, line 514: Calling dispose while running a watcher-enabled builder will now allow the process terminate successfully.

elsassph (approved):

PR #311, opened 2021-02-11
Don't mangle xml scripts during transpile

TwitchBronBron:
Fix bug with multiple-in-a-row transpiles for xml files that hadn't changed, would cause scripts to get mangled.

### DIFF
--- a/src/files/XmlFile.ts
+++ b/src/files/XmlFile.ts
@@ -454,16 +454,20 @@ export class XmlFile {
             return script;
         });
 
-        const extraImportIdx = this.ast.component.scripts.length - 1;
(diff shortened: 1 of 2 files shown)


src/files/XmlFile.ts
```diff
@@ -454,16 +454,20 @@ export class XmlFile {
             return script;
         });
 
-        const extraImportIdx = this.ast.component.scripts.length - 1;
```
TwitchBronBron, line 457: Rather than mess with array indexes...now we just push a new array, and restore the old one after the fact.

elsassph (approved):

Issue #179, opened 2020-08-26
Avoid requiring 'bs:disable-line for Roku_Ads() if library is imported


iObject:
We should avoid requiring 'bs:disable-line for Roku_Ads() if library is imported.

TwitchBronBron:
I started investigating this like 6 months ago, but kept getting pulled into other things. We still need to add support for detecting the library import in the manifest, and then adding those functions into the global function scope if imported. 

@iObject is this something you would have any interest in implementing? We can have a quick chat on Slack if you want to work out the implementation details. I don't expect it to be all that complicated.

TwitchBronBron:
I did a some on-device investigation, just to get a broader understanding of the edge cases for this feature:

1. in `source` scope, the `Library "Roku_Ads.brs"` statement can appear in any file. It does not need to be in the same file as the call to `Roku_Ads()`, and does not need to be in the file with the `main()` function.
2. in a `Task` component scope, the `Library "Roku_Ads.brs"` statement can appear in any of the referenced script files. it does not need to be in the same file as the call to `Roku_Ads()`. 
2. Declaring the `Library` statement in `source` scope, but calling `Roku_Ads` in a component scope results in `function not found` error
3. Declaring the `Library` statement in a component scope, and in the same file calling `Roku_Ads()` fails with `[RAF.err] RAF call in wrong environment (RENDER thread?). RAF only works from task or main() thread`


So, to summarize: 
1. the `Library "Roku_Ads.brs"` statement must be declared in a script in the same scope as the call to `Roku_Ads()`
2. Component scopes will crash when calling `Roku_Ads()` even though they support the `Library "Roku_Ads.brs"` statement (it's useless)

elliot-nelson:
Copying some ideas from discussion in Slack:

In your own Roku project, you could create a `.d.bs` file that describes the function you are using, for example:

```brighterscript
' src/source/Roku_Ads.d.bs

function Roku_Ads()
end function
```

Now you can call this `Roku_Ads` function from your main thread:

```brighterscript
Library "Roku_Ads.brs"

sub main()
  Roku_Ads()
end sub
```

You can also call it from a task, or component, by importing the `.d.bs` file:

```brighterscript
import "../source/Roku_Ads.d.bs"

sub init()
  Roku_Ads()
end sub
```

This allows unblocking any libraries used in your own project.

[...]

PR #312, opened 2021-02-13
Intermediary fix for Roku_Ads error

TwitchBronBron:
This is a temporary fix for the fact that `Roku_Ads` is _sometimes_ available globally (dependent on a `Library` import. 

Long-term we want to add formal `Library` handling. However, There's a lot of other more important things to work on right now, so this will hold us over until then.

Provides temporary relief for #179

### DIFF
--- a/src/globalCallables.spec.ts
+++ b/src/globalCallables.spec.ts
@@ -1,6 +1,7 @@
 import { standardizePath as s } from './util';
 import { Program } from './Program';
 import { expect } from 'chai';
+import { expectZeroDiagnostics } from './testHelpers.spec';
 
 let tmpPath = s`${process.cwd()}/.tmp`;
 let rootDir = s`${tmpPath}/rootDir`;
@@ -16,6 +17,18 @@ describe('globalCallables', () => {
     });
     afterEach(() => {
         program.dispose();
+    });
+
+    describe('Roku_ads', () => {
+        it('exists', () => {
+            program.addOrReplaceFile('source/main.brs', `
+                sub main()
+                    adIface = Roku_Ads()
+                end sub
+            `);
+            program.validate();
+            expectZeroDiagnostics(program);
+        });
     });
 
     describe('val', () => {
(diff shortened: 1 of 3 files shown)


src/testHelpers.spec.ts
```diff
@@ -57,10 +57,10 @@ export function expectZeroDiagnostics(arg: { getDiagnostics(): Array<Diagnostic>
     let diagnostics: BsDiagnostic[];
     if (Array.isArray(arg)) {
         diagnostics = arg as BsDiagnostic[];
-    } else if ((arg as any).diagnostics) {
-        diagnostics = (arg as any).diagnostics;
     } else if ((arg as any).getDiagnostics) {
         diagnostics = (arg as any).getDiagnostics();
+    } else if ((arg as any).diagnostics) {
```
TwitchBronBron, line 62: This is just a test helper function. The `Program` object has `diagnostics` that are program-specific, but also has a `getDiagnostics()` function that rolls up its own diagnostics along with all scopes and files. So we should check for `getDiagnostics()` first since that one will be more complete.

src/globalCallables.spec.ts
```diff
@@ -18,6 +19,18 @@ describe('globalCallables', () => {
         program.dispose();
     });
 
+    describe('Roku_ads', () => {
+        it('exists', () => {
```
TwitchBronBron, line 23: We'll update this test once we have formal library support.

rrirower:
Has there been any action on this?  I'm trying to use 'Roku_Ads()' within a script file.  I've included the Library call at the top.  When I call the function, I get: `Could not load source 'roku_ads_lib:/libsource/Roku_Ads_Beacons.brs'.`

rrirower:
Is there a workaround for this issue?  When trying to use Roku_Ads() in the same source as the Library statement, I get:

```
Found 1 compile error
--- Error loading file. (compile error &hb9) in pkg:/components/tasks/PlayerTask.brs(4) 'Roku_Ads.brs'
*** ERROR compiling PlayerTask:
```

rrirower:
Disregard my earlier post.  I was able to include Roku_Ads() after reading some of your other posts.

PR #315, opened 2021-02-15
do not revalidate on changes that are outside of any workspace

georgejecook:
I spotted an issue where vscode would constantly update itself. There's some config file that get's updated whenever a user does pretty much anything, and sometimes in the background.

This meant that whole workspace revalidations would occur every time the user does anything in the ide, such as go to a symbol, show message tool tips, degbug, I suppost just about anything. In my case it was causing catastrophic performance issues culminating in whirring fans.


TwitchBronBron (approved):
Looks great. Awesome find!

PR #317, opened 2021-02-17
Standardize sourcenode

TwitchBronBron:
Centralizes all of the brs transpiling SourceNode line and column tracking into a single function.
 - the previous `state.sourceNode` implementation did not convert 0-based Range line to 1-based SourceNode line. This fixes that problem and should improve debug precision
 - remove use of `new SourceNode` from `Expressions.ts` and `Statements.ts`
 - for token transpiling, use `state.tokenToSourceNode` which should have a slight performance bump since the token text is passed in by reference on the `token` reference rather than by value as the second parameter to `state.sourceNode`
 - fixed a few small bugs in transpile

### DIFF
--- a/src/parser/TranspileState.ts
+++ b/src/parser/TranspileState.ts
@@ -1,6 +1,7 @@
 import { SourceNode } from 'source-map';
 import type { Range } from 'vscode-languageserver';
 import type { BrsFile } from '../files/BrsFile';
+import type { Token } from '../lexer/Token';
 import type { ClassStatement } from './Statement';
 
 /**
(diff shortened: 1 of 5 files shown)


src/parser/TranspileState.ts
```diff
@@ -68,13 +69,30 @@ export class TranspileState {
     /**
      * Shorthand for creating a new source node
      */
-    public sourceNode(locatable: { range: Range }, code: string) {
-        let result = new SourceNode(
-            locatable.range.start.line,
+    public sourceNode(locatable: { range: Range }, code: string | SourceNode | Array<string | SourceNode>): SourceNode | undefined {
+        return new SourceNode(
+            //convert 0-based range line to 1-based SourceNode line
+            locatable.range.start.line + 1,
```
TwitchBronBron, line 75: This is a bug that probably caused all sorts of issues when stepping through transpiled code in the debugger.

elsassph (approved):
It's great to remove the redundant logic, though I don't think it can have an impact on performance.
Did you try the impact on the line index change during debugging?

TwitchBronBron:
The benchmarks show these changes don't have an affect on transpiling.
image

Running a single file transpile (just to show some higher numbers):
image

In terms of actually seeing the bugs during debugging, I didn't have any specific tests, I just know that sometimes I would see it debugging the wrong line. This is sometimes related to roku reporting the wrong line, but the rest would be related to these SourceNode bugs.

PR #319, opened 2021-02-17
include annotations in type files

georgejecook:
- include annotations in type definition output

### DIFF
--- a/src/Program.spec.ts
+++ b/src/Program.spec.ts
@@ -1746,6 +1746,54 @@ describe('Program', () => {
                 expect(fsExtra.pathExistsSync(s`${stagingFolderPath}/source/Duck.d.brs`)).to.be.false;
             });
 
+            it('includes annotations in type defs for .bs files', async () => {
+                program.addOrReplaceFile<BrsFile>('source/Duck.bs', `
+                    namespace test
+                    @an
+                    @anFunc("value")
+                    function getDuck()
+                    end function
+                        @anClass
+                        class Duck
+                            @anMember
+                            @anMember("field")
+                            private thing
+
+                            @anMember
+                            @anMember("func")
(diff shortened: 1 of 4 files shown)


src/Program.spec.ts
```diff
@@ -1746,6 +1746,54 @@ describe('Program', () => {
...
+                    end namespace`);
+                program.options.emitDefinitions = true;
+                program.validate();
+                await program.transpile([], stagingFolderPath);
+
+                expect(fsExtra.pathExistsSync(s`${stagingFolderPath}/source/Duck.brs`)).to.be.true;
+                expect(fsExtra.pathExistsSync(s`${stagingFolderPath}/source/Duck.d.bs`)).to.be.true;
+                expect(fsExtra.pathExistsSync(s`${stagingFolderPath}/source/Duck.d.brs`)).to.be.false;
+                let dFileText = fsExtra.readFileSync(s`${stagingFolderPath}/source/Duck.d.bs`).toString();
+                expect(dFileText).to.equal(trim`namespace test
```
TwitchBronBron, line 1776: please use `trimLeading` and properly indent this test code.

src/parser/Statement.ts
```diff
@@ -377,8 +377,16 @@ export class FunctionStatement extends Statement implements TypedefProvider {
         return this.func.transpile(state, nameToken);
     }
 
-    getTypedef(state: TranspileState) {
-        return this.func.getTypedef(state, this.name);
+    getTypedef(state: TranspileState, isClassMethod = false) {
```
TwitchBronBron, line 380: This doesn't belong here. This breaks the point of inheritance. ClassMethod should handle its own transpiling of annotations (or not), and same with FunctionStatement (because it wraps a FunctionExpression).

PR #327, opened 2021-02-20
fixes template strings introducing errant plus when there are no items

georgejecook:


src/parser/tests/expression/TemplateStringExpression.spec.ts
```diff
@@ -74,6 +74,13 @@ describe('TemplateStringExpression', () => {
             program.dispose();
         });
 
+        it('properly transpiles simple template string with no leading text', () => {
+            testTranspile(
+                `a = \`\${one},\${two}\``,
```
TwitchBronBron, line 79: You can use single quotes for this string so that you don't need to escape the template string stuff. 

```
'a = `${one},${two}`',

TwitchBronBron: I fixed this.

TwitchBronBron (approved):
Looks great! I tweaked the input test for your call to `testTranspile` so we didn't have to escape so many characters. Thanks for this!

PR #329, opened 2021-02-22
Misc file tweaks

TwitchBronBron:
Here are some internal changes that I've been wanting to make for a while. These are breaking changes, but I'd rather get them in now before more projects start to depend on brighterscript and would have a more wide-reaching impact.

Notable changes:
- rename all instances of `pathAbsolute` to `srcPath` for brevity and clarity
- `pkgPath` strings now contain an actual pkg path (i.e. `"source/main.brs"` is now `"pkg:/source/main.brs"`)
- `pkgPath` originally stored the platform-specific path separator (`\` on windows, `/` on everything else). Now `pkgPath` always stores the unix separator because that's what is used on the roku device. 
- combine some of the `program` file functions to auto-detect the path type and use the correct file collection (`this.files` or `this.pkgMap`) accordingly.
- `program.files` and `program.pkgMap` now store their keys in all lower case, which removes the need to iterate the collection to find files 
- Add a `normalizePath: boolean` parameter to `program.getFile` to support skipping `util.standardizePath` when the caller knows the path has already been normalized


src/files/XmlFile.spec.ts
```diff
@@ -377,12 +375,11 @@ describe('XmlFile', () => {
 
     describe('getCompletions', () => {
         it('formats completion paths with proper slashes', () => {
-            let scriptPath = s`C:/app/components/component1/component1.brs`;
-            program.files[scriptPath] = new BrsFile(scriptPath, s`components/component1/component1.brs`, program);
+            program.addOrReplaceFile('pkg:/components/component1/component1.brs', '');
 
-            let xmlFile = new XmlFile(s`${rootDir}/components/component1/component1.xml`, s`components/component1/component1.xml`, <any>program);
+            let xmlFile = new XmlFile(s`${rootDir}/components/component1/component1.xml`, 'pkg:/components/component1/component1.xml', <any>program);
```
elsassph, line 380: Should `s` be also removed for the `srcPath` argument, as it is normalised by the file class as well?

src/util.ts
```diff
@@ -370,32 +372,27 @@ export class Util {
...
             return null;
         }
+        //if the target starts with 'pkg:', return as-is
+        if (targetPath.startsWith('pkg:/')) {
+            return targetPath;
+        }
 
         //remove the filename
-        let containingFolder = path.normalize(path.dirname(containingFilePathAbsolute));
+        let containingFolder = path.posix.normalize(path.dirname(sourcePkgPath));
```
elsassph, line 390: It looks like `path.posix.normalize` does NOT replace `\` with `/`

src/files/XmlFile.spec.ts
TwitchBronBron: Yeah, we can remove it. I figured this PR was huge already, and this is in a spec file, so it wasn't overly worthwhile do to that right now. I plan on doing a large refactoring of the test suite sometime after this lands. However, if you really want, I can do all of that refactoring in here too.

src/util.ts
TwitchBronBron: Oh, yeah good call. I didn't realize that. I still want to use normalize so we can fix any `../` or double slashes, but I refactored this to use a regex split on either path separator just to be safe.

only pkgPaths should be passed to this, but better safe than sorry i guess?

src/util.ts
```diff
@@ -1175,6 +1173,23 @@ export class Util {
             range: attr.range
         } as SGAttribute;
     }
+
+    /**
+     * Given a pkgPath, remove the `pkg:/` portion of the path.
+     */
+    public removePkgProtocol(pkgPath: string) {
+        if (pkgPath.startsWith('pkg:/')) {
```
elsassph, line 1181: Maybe add `libpkg:`?

src/util.ts
```diff
@@ -70,11 +70,13 @@ export class Util {
     /**
      * Given a pkg path of any kind, transform it to a roku-specific pkg path (i.e. "pkg:/some/path.brs")
      */
-    public getRokuPkgPath(pkgPath: string) {
+    public sanitizePkgPath(pkgPath: string) {
         pkgPath = pkgPath.replace(/\\/g, '/');
-        return 'pkg:/' + pkgPath;
+        if (!pkgPath.startsWith('pkg:/')) {
```
elsassph, line 75: Maybe add `libpkg:`

src/util.ts
```diff
@@ -370,32 +372,26 @@ export class Util {
...
-            if (targetPath === '') {
-                return null;
-            } else {
-                return path.normalize(targetPath);
-            }
-        }
-        if (targetPath === 'pkg:') {
+    public getPkgPathFromTarget(sourcePkgPath: string, targetPath: string) {
+        //handle some edge cases
+        if (targetPath === 'pkg:' || targetPath === 'pkg:/') {
```
elsassph, line 381: Maybe add `libpkg:`

TwitchBronBron:
Closing in favor of #399

src/util.ts
TwitchBronBron: Handled in 50a155549b9b4a9fbacbef8ecc769109d6abc66e

src/util.ts
TwitchBronBron: Handled in 50a155549b9b4a9fbacbef8ecc769109d6abc66e

src/util.ts
TwitchBronBron: Handled in 50a155549b9b4a9fbacbef8ecc769109d6abc66e

PR #330, opened 2021-02-23
Refactor plugin events to emit an event object

TwitchBronBron:
Refactored plugin events so that they all emit a single event object instead of ordered parameters. This will allow us to add new properties to the events in the future without complicating the parameter order, and will allow us to more easily deprecate parameters in the future. This also solves the problem with ordered parameters where we try to ensure parameter types across all events are in the same order

This should not be merged until https://github.com/rokucommunity/brighterscript/pull/329 is ready, just because we want to release all of the breaking changes at the same time.


src/Program.ts
```diff
@@ -533,11 +536,50 @@ export class Program {
...
+
+                //for every unvalidated file, validate it
+                if (!file.isValidated) {
+                    this.plugins.emit('beforeFileValidate', {
+                        program: this,
+                        file: file
+                    });
+
+                    //call file.validate() IF the file has that function defined
+                    file.validate?.();
```
elsassph, line 548: Why would that happen?

elsassph (approved):

src/Program.ts
TwitchBronBron: I'm planning on eventually adding the ability for plugins to register their own file types. So this is safety for when a custom file doesn't implement a validate function.

TwitchBronBron:
Closing this in favor of #397 which targets the v1 branch

PR #334, opened 2021-02-25
bslib ropm support

TwitchBronBron:
auto-detect if bslib is installed as a ropm module, and don't copy our version of bslib during transpile if true (in favor of using the version from ropm).


src/Scope.ts
```diff
@@ -835,6 +836,10 @@ export class Scope {
             let referencedFile = this.getFileByRelativePath(scriptImport.pkgPath);
             //if we can't find the file
             if (!referencedFile) {
+                //skip the default bslib file, it will exist at transpile time but should not show up in the program during validation cycle
+                if (s`${scriptImport.pkgPath}` === `source${path.sep}bslib.brs`) {
```
elsassph, line 840: Do you need `path.sep`? Does `scriptImport.pkgPath` need to be normalized?

src/util.ts
```diff
@@ -1175,6 +1175,31 @@ export class Util {
...
+     * Copy the version of bslib from local node_modules to the staging folder
+     */
+    public async copyBslibToStaging(stagingDir: string) {
+        //copy bslib to the output directory
+        await fsExtra.ensureDir(standardizePath(`${stagingDir}/source`));
+        // eslint-disable-next-line
+        const bslib = require('bslib');
+        let source = bslib.source as string;
+
+        //apply the `bslib_` prefixe to the functions
```
elsassph, line 1189: `prefix`

README.md
```diff
@@ -255,6 +255,22 @@ end sub
...
+
+To encourage reducing code duplication, BrighterScript has built-in support for loading `bslib` from [ropm](https://github.com/rokucommunity/ropm). Here's how it works:
+1. if your program does not use ropm, or _does_ use ropm but does not directly reference bslib, then the BrighterScript compiler will copy bslib to `"pkg:/source/bslib.brs"` at transpile-time.
+2. if your program uses ropm and has installed `bslib` as a dependency, then the BrighterScript compiler will _not_ emit a copy of bslib at `pkg:/source/bslib.brs`, and will instead use the path to the version from ropm `pkg:/source/roku_modules/bslib/bslib.brs`.
+
+
+
+### Installing bslib in your ropm-enabled project
+bslib is actually published to npm under the name [@RokuCommunity/bslib](https://npmjs.com/package/@rokucommunity/bslib). However, to keep the bslib function names short, BrighterScript requires that you install @rokucommunity/bslib with the `bslib` alias. Here's the command to do that using the ropm CLI.
```
elsassph, line 269: Lowercase `@RokuCommunity`?

src/Scope.ts
TwitchBronBron: Yeah we need path.sep right now. Until #329 merges, pkgPath uses windows separators on windows and unix slashes everywhere else. 

I'll check on whether scriptImport is already normalized elsewhere, perhaps we can skip this normalization.

README.md
TwitchBronBron: resolved

src/util.ts
TwitchBronBron: resolved

src/Scope.ts
TwitchBronBron: yeah, scriptImport has already been normalized, so this is redundant. i removed it. good catch!

TwitchBronBron:
@elsassph i believe ci have addressed all of your concerns.

elsassph (approved):

PR #340, opened 2021-02-26
Completion for pkg path inside strings

### LINKED ISSUE #228: completions should not show when inside a string
Completions should not show when inside a string. 
image

One exception to this should be when doing completions inside strings within known callback function parameter positions (like ObserveField, etc...).

### LINKED ISSUE #333: File path completion for any string starting with pkg: or libpkg:
image
In situations like the above it's likely safe to assume the developer is trying to type a file path in the codebase. We should also offer this type of completion out side of imports.

There are two main file protocols we need to look for at the start of the string. The main one being `pkg:` and the second being `libpkg:`. The second is used in some cases by component libraries.

TwitchBronBron:
**Notable Changes:**
 - Show completions for file paths inside strings when they start with `pkg:` or `libpkg:`. 
 - don't show completions inside of a string (expect for the pkg: and libpkg: changes mentioned above)

I couldn't get this to auto-show the completions while typing. so you have to manually activate the intellisense menu using a keyboard shortcut or the command pallet. 

6e9c0583-4288-489e-85ba-5668a3e57c45

fixes #333 
fixes #228


chrisdp (approved):

src/LanguageServer.ts
```diff
@@ -194,7 +194,7 @@ export class LanguageServer {
                 completionProvider: {
                     resolveProvider: true,
                     //anytime the user types a period, auto-show the completion results
-                    triggerCharacters: ['.'],
+                    triggerCharacters: ['.', ':'],
```
TwitchBronBron, line 197: @chrisdp  @elsassph @georgejecook  I added a colon to the trigger characters. This means anytime you type a colon _anywhere_ in your code, it will trigger completions. The typescript language server doesn't do this, but it was the only way to get `"pkg:` inside a string to trigger these completions automatically. 

The side-effect of this is that it triggers the global completions inside of associative arrays too (after typing the colon). I'm not sure that's actually a bad thing, but I just wanted to get your feedback on whether that would be annoying or acceptable.

chrisdp: Hummmm my gut tells me that could be very annoying to encounter. I would need to likely test it in practice to get a feel but there are quite a few times where I have completions I have to dismiss. Knowing when to stop suggesting is something that could still be improved so I think this will just add to that. 

I would likely rather be ok with just needing to hit the completion hotkey for now.

elsassph: I normally would like that behaviour, but for AAs it may be annoying to have it pop, especially because the completion list will disappear as soon as you press a space for instance, which means it's going to pop for no useful reason.

src/files/BrsFile.ts
```diff
@@ -768,8 +768,37 @@ export class BrsFile {
...
+        const tokenKind = currentToken?.kind;
+        if (tokenKind === TokenKind.Comment) {
             return [];
+        } else if (tokenKind === TokenKind.StringLiteral || tokenKind === TokenKind.TemplateStringQuasi) {
+            const match = /^("?)(pkg|libpkg):/.exec(currentToken.text);
+            if (match) {
+                const [, openingQuote, fileProtocol] = match;
+                //include every absolute file path from this scope
+                for (const file of scope.getAllFiles()) {
+                    const pkgPath = `${fileProtocol}:/${file.pkgPath.replace(/\\/g, '/')}`;
```
elsassph, line 780: `pkgPath` is supposed to be normalised already no? Seems like a waste.

TwitchBronBron: Until #329 merges, pkgPath uses windows separators on windows, and unix separators everywhere else. So this is absolutely necessary right now.

src/LanguageServer.ts
TwitchBronBron: Yeah, i agree. it would be nice to have the completions show for this, but it's not worth the side-effects. I removed the colon trigger character.

PR #343, opened 2021-02-27
Iife parse

TwitchBronBron:
Fixes a critical bug in BrsFile that would crash entire file parsing if the file contained an immediately-invoked-function-expression. 
Also supports non-grouped IIFEs (gross, but technically valid syntax...so we should support it).

### DIFF
--- a/src/files/BrsFile.ts
+++ b/src/files/BrsFile.ts
@@ -649,12 +649,13 @@ export class BrsFile {
         for (let func of this._parser.references.functionExpressions) {
             //for all function calls in this function
             for (let expression of func.callExpressions) {
-
                 if (
                     //filter out dotted function invocations (i.e. object.doSomething()) (not currently supported. TODO support it)
                     (expression.callee as any).obj ||
                     //filter out method calls on method calls for now (i.e. getSomething().getSomethingElse())
-                    (expression.callee as any).callee
+                    (expression.callee as any).callee ||
+                    //filter out callees without a name (immediately-invoked function expressions)
+                    !(expression.callee as any).name
                 ) {
                     continue;
                 }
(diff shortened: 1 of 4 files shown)


src/files/BrsFile.ts
```diff
@@ -649,12 +649,13 @@ export class BrsFile {
...
             for (let expression of func.callExpressions) {
-
                 if (
                     //filter out dotted function invocations (i.e. object.doSomething()) (not currently supported. TODO support it)
                     (expression.callee as any).obj ||
                     //filter out method calls on method calls for now (i.e. getSomething().getSomethingElse())
-                    (expression.callee as any).callee
+                    (expression.callee as any).callee ||
+                    //filter out callees without a name (immediately-invoked function expressions)
+                    !(expression.callee as any).name
```
TwitchBronBron, line 658: This was the fix for of the crash. IIFEs don't have a name, but lower in this function we expected the `expression.callee` to have a name token. So now we guard against any callee without a name.

src/parser/Parser.ts
```diff
@@ -1937,7 +1937,13 @@ export class Parser {
 
     private anonymousFunction(): Expression {
         if (this.checkAny(TokenKind.Sub, TokenKind.Function)) {
-            return this.functionDeclaration(true);
+            const func = this.functionDeclaration(true);
+            //if there's an open paren after this, this is an IIFE
+            if (this.check(TokenKind.LeftParen)) {
+                return this.finishCall(this.advance(), func);
```
TwitchBronBron, line 1943: Roku supports non-grouped IIFE. example:
```
print function()
    return true
end function()
```

This adds parser support for that pattern.

elsassph (approved):

PR #345, opened 2021-03-01
Fix - print statement parsing

### LINKED ISSUE #344: single-line if statement with else regression
The `IfStatement` refactoring from [v0.23.0](https://github.com/rokucommunity/brighterscript/blob/master/CHANGELOG.md#0230---2020-12-18) introduced a regression where single-line if statements with an `else` portion now cause parse errors. We weren't validating this syntax in the test suite which is probably why this got missed.

This spec shows the issue:
```
it.only('parses single line if statement with else', () => {
    const parser = Parser.parse(`
        if true print 1 else print 1
        if true then print 1 else print 1
    `);
    expectZeroDiagnostics(parser);
});
```

elsassph:
Print statement parsing needs to stop when it reaches an `else`

Fixes #344

### DIFF
--- a/src/parser/Parser.ts
+++ b/src/parser/Parser.ts
@@ -1782,7 +1782,7 @@ export class Parser {
             return new ReturnStatement(tokens);
         }
 
-        let toReturn = this.expression();
+        let toReturn = this.check(TokenKind.Else) ? undefined : this.expression();
         return new ReturnStatement(tokens, toReturn);
     }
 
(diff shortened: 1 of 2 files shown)


TwitchBronBron:
Thanks @elsassph . What about return statements? Are there other statement types that might cause this same issue?

```
if childCount = 0 then return else print 2
```
image

elsassph:
@TwitchBronBron thanks, `return` without value is one problem too. Can't think of others for now.

PR #346, opened 2021-03-01
Bslib fixes

TwitchBronBron:
Adds:
 - support for both npm aliased bslib AND non-aliased:
     - `ropm install bslib@npm:@rokucommunity/bslib`
     - `ropm install @rokucommunity/bslib`

Fixes:
 - bug with npm alias of bslib for node version 10
 - infinite loop when rewriting the bslib file during transpile.

### DIFF
--- a/README.md
+++ b/README.md
@@ -271,7 +271,7 @@ bslib is actually published to npm under the name [@rokucommunity/bslib](https:/
 ropm install bslib@npm:@rokucommunity/bslib
 ```
 
-If you're using a version of NodeJS that doesn't support aliases, then you can also simply install @rokucommunity/bslib directly without the alias, and BrighterScript will recognize this as well.
+If you're using a version of NodeJS that doesn't support npm aliases, then you can also simply install @rokucommunity/bslib directly without the alias, and BrighterScript will recognize this as well.
 ```bash
 ropm install @rokucommunity/bslib
 ```
--- a/README.md
+++ b/README.md
@@ -271,6 +271,10 @@ bslib is actually published to npm under the name [@rokucommunity/bslib](https:/
 ropm install bslib@npm:@rokucommunity/bslib
 ```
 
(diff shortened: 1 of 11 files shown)


README.md
```diff
@@ -271,6 +271,10 @@ bslib is actually published to npm under the name [@rokucommunity/bslib](https:/
 ropm install bslib@npm:@rokucommunity/bslib
 ```
 
+If you're using a version of NodeJS that doesn't support npm aliases, then you can also simply install @rokucommunity/bslib directly without the alias, and BrighterScript will recognize this as well.
```
elsassph, line 274: Why even recommending the alias-based approach if the simple install works as well?

elsassph: Oh ok, then maybe you should say that "If you use Node 12+" it is "recommended" to install using the alias.

elsassph: PS: Generally whitespace in the readme is a bit random.
Sometimes there is a blank line around titles, sometimes not.
My OCD, it is hurt.

elsassph (approved):

README.md
TwitchBronBron: @elsassph I tweaked the readme to be a little more clear. Do you like that any better? 

I really like using `bslib_` as a prefix. `rokucommunity_bslib_` is really verbose, and kind of gross to look at (even though yes, I know the transpiled code isn't exactly pretty anyway...). I could be swayed to remove support for the `bslib` alias entirely if @georgejecook and @chrisdp  and @triwav also agree it's not a big deal to keep the `bslib_` shorter alias. 

I also added some whitespace before the headers. I've been sloppy with that, sorry!

triwav: I'd prefer `bslib_` as well. If I'm understanding correctly they have to install it themselves vs having ropm do it for them @TwitchBronBron?

georgejecook: I say keep the short alias

TwitchBronBron: Ok cool. I'll keep the logic as-is, (look for the short alias first, but support the full version if present to support older versions of node). Thanks!

PR #347, opened 2021-03-02
Add codeAction for bs imports

TwitchBronBron:
Support for brighterscript code action to suggest file imports for missing functions.


src/CodeActionUtil.ts
```diff
@@ -25,6 +25,64 @@ export class CodeActionUtil {
...
     }
+
+    private rangeToKey(range: Range) {
+        return `${range?.start?.line}:${range?.start?.character},${range?.end?.line}:${range?.end?.character}`;
+    }
+
+    /**
+     * Generate a unique key for each code action, and only keep one of each action
+     */
+    public dedupe(codeActions: CodeAction[]) {
```
TwitchBronBron, line 36: This function generates a unique key for each code action, and then only returns one of each unique code action. 

The reason I needed to do this was the `OnScopeGetCodeAction` event would produce a code action for each scope for a missing function, but the action was the same "import "somefile.brs".

TwitchBronBron: vscode does not dedupe the code actions, so this would result in many identical code actions showing in the UI.

src/util.ts
```diff
@@ -1200,6 +1200,23 @@ export class Util {
         }
         await fsExtra.writeFile(`${stagingDir}/source/bslib.brs`, source);
     }
+
+    /**
+     * Given a Diagnostic or BsDiagnostic, return a copy of the diagnostic
+     */
+    public toDiagnostic(diagnostic: Diagnostic | BsDiagnostic) {
```
TwitchBronBron, line 1207: the `BsDiagnostic` objects contain a `file` reference which has a `program`, which has an array of files (including the one we just referenced). This causes a recursive error during JSON.stringify when the language server sends the code actions or diagnostics to the client. This function will clone a diagnostic, but only keep the officially supported properties. Thus preventing the JSON.stringify crash.

TwitchBronBron: any time we send diagnostics to the client, we should run them through this function.

src/parser/Parser.ts
```diff
@@ -2632,30 +2632,29 @@ export interface ParseOptions {
...
+export class References {
+    public assignmentStatements = [] as AssignmentStatement[];
+    public classStatements = [] as ClassStatement[];
+    public functionExpressions = [] as FunctionExpression[];
+    public functionStatements = [] as FunctionStatement[];
+    /**
+     * A map of function statements, indexed by fully-namespaced lower function name.
+     */
+    public get functionStatementLookup() {
+        if (!this._functionStatementLookup) {
```
TwitchBronBron, line 2644: Compute this on-demand, so we don't pay the price until we need it. (currently only used during code actions)

src/bscPlugin/codeActions/XmlFileCodeActionsProcessor.ts
```diff
@@ -18,19 +19,19 @@ export class XmlFileCodeActionsProcessor {
...
-    private addMissingExtends() {
+    private addMissingExtends(diagnostic: DiagnosticMessageType<'xmlComponentMissingExtendsAttribute'>) {
         const { component } = this.file.parser.ast;
         //inject new attribute after the final attribute, or after the `<component` if there are no attributes
         const pos = (component.attributes[component.attributes.length - 1] ?? component.tag).range.end;
         this.codeActions.push(
             codeActionUtil.createCodeAction({
                 title: `Extend "Group"`,
-                // diagnostics: [diagnostic],
+                diagnostics: [diagnostic],
```
TwitchBronBron, line 34: We can send diagnostics for these actions now that we are sanitizing them before sending them in the language server.

src/bscPlugin/codeActions/ScopeBrsFileCodeActionsProcessor.ts
```diff
@@ -0,0 +1,57 @@
...
+            //certain brighterscript-specific code actions
+            if (isBsFile) {
+                if (diagnostic.code === DiagnosticCodeMap.callToUnknownFunction) {
+                    this.suggestImports(diagnostic as any);
+                }
+            }
+        }
+    }
+
+    private suggestImports(diagnostic: DiagnosticMessageType<'callToUnknownFunction'>) {
```
TwitchBronBron, line 34: So, the the only "issue" with this function is that it will generate a code action for every scope, which are all identical. I don't have a great way of preventing that on this side, so @elsassph I'm up for any suggestions you might have. 

I could handle this in the onProgramGetDiagnostics, and look for the FIRST occurance of this diagnostic on any scope, but that seems like excess diagnostic filtering and scope looping. but then again, this might be more expensive since there could be many scopes with this problem.

codeActions are also done on demand, so combined with the diagnostic deduping, perhaps it's not worth working around right now. it's still really fast...

elsassph: Ahem, reminds me about our discussion about whether there should be scope-based actions at all.

src/bscPlugin/codeActions/XmlFileCodeActionsProcessor.ts
elsassph: What does the diagnostic give?

src/CodeActionUtil.ts
elsassph: That looks fantastically expensive to compute, if the goal is to just to avoid creating multiple quick-fixes.
I'd suggest just deduplicating on the title and making that clear somehow in the API/doc.

src/bscPlugin/codeActions/XmlFileCodeActionsProcessor.ts
TwitchBronBron: I don't actually know. I looked through the docs and they don't explain it at all.

src/bscPlugin/codeActions/ScopeBrsFileCodeActionsProcessor.ts
TwitchBronBron: Yep, you won me over. :) I refactored the code to only have a single program event now.

TwitchBronBron:
@elsassph I've finished refactoring the codeActions plugin functionality. Now there's only a single `onProgramGetCodeActions` event. It emits all diagnostics for the entire program, filtered by only those that touch the range from the event. If we want scope-related information, then BsDiagnostic.data can have a `scope` object added (we'll do this on an as-needed basis as we find diagnostics that need that). 

This is technically a breaking change for the codeActions plugin api, but it was only recently introduced, and the entire plugin system is still in beta so we are still at liberty to break it as needed. This is the first step towards object-based plugin events (the rest will land in #330). 

It's ready for another review.

src/bscPlugin/codeActions/CodeActionsProcessor.ts
```diff
@@ -0,0 +1,108 @@
...
+    public constructor(
+        public event: OnGetCodeActionsEvent
+    ) {
+
+    }
+
+    public process() {
+        for (const diagnostic of this.event.diagnostics) {
+            if (diagnostic.code === DiagnosticCodeMap.callToUnknownFunction) {
+                this.suggestImports(diagnostic as any);
```
elsassph, line 20: Can do better than `as any`?

src/bscPlugin/codeActions/CodeActionsProcessor.ts
```diff
@@ -0,0 +1,108 @@
...
+
+        //skip generating duplicate suggestions if we've already done this one
+        if (this.suggestedImports.has(lowerFunctionName)) {
+            return;
+        }
+
+        this.suggestedImports.add(lowerFunctionName);
+
+        //find the position of the first import statement, or the top of the file if there is none
+        const insertPosition = (this.event.file as BrsFile).parser.references.importStatements[0]?.importToken.range?.start ?? util.createPosition(0, 0);
```
elsassph, line 40: Traditionally prefered to add as last import, though could be a problem if imports can be anywhere.

src/CodeActionUtil.ts
TwitchBronBron: this was removed.

src/bscPlugin/codeActions/CodeActionsProcessor.ts
TwitchBronBron: yeah, good point. i was just being lazy. I could easily do `importStatements[statements.length-1]` to get the last one. If they have imports anywhere, that's a bsc error, and it's their own fault. :D

src/bscPlugin/codeActions/CodeActionsProcessor.ts
TwitchBronBron: We can be more specific, but the `suggestImports` function already typed the diagnostic parameter. It just seemed needlessly verbose to cast as the same type when the entire loop is clearly related to diagnostics. 

```typescript
private suggestImports(diagnostic: DiagnosticMessageType<'callToUnknownFunction'>) {
```

src/bscPlugin/codeActions/CodeActionsProcessor.ts
TwitchBronBron: I fixed this.

TwitchBronBron:
@elsassph i addressed one of your concerns, but I didn't address https://github.com/rokucommunity/brighterscript/pull/347#discussion_r589020793 because I wanted to verify that you thought it was actually useful to duplicate the typecast there (I don't see much value)

PR #351, opened 2021-03-05
Fixes #348 - Use 'toTypeString' method on transpilation of function return types

### LINKED ISSUE #348: Transpilation Bug: CustomTypes used as return values do not get transpiled to "object"
This is occurring in v0.34.2

For example, this file:

```vb

class SomeKlass
end class

function foo() as SomeKlass
 return new SomeKlass()
end function
```

Gets transpiled to (notice the `foo() as SomeKlass`):

```vb
function __SomeKlass_builder()
    instance = {}
    instance.new = sub()
    end sub
    return instance
end function
function SomeKlass()
    instance = __SomeKlass_builder()
    instance.new()
    return instance
end function

function foo() as SomeKlass
    return SomeKlass()
end function'//# sourceMappingURL=./test.bs.map
```

markwpearce:
Also added a test with transpiling with custom types.

### DIFF
--- a/src/Program.spec.ts
+++ b/src/Program.spec.ts
@@ -1750,7 +1750,6 @@ describe('Program', () => {
                 s`${sourceRoot}/source/main.bs`
             );
         });
-
     });
 
     describe('typedef', () => {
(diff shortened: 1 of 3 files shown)


TwitchBronBron:
Good catch! Thanks for the fix.

PR #353, opened 2021-03-09
Fix irrecoverable compiler state on duplicate components

TwitchBronBron:
This fixes a bug that has been in the compiler for a long time. The issue turned out to be that having multiple components defined with the same name would cause the component scopes to get confused, and it was hard to recover. 

The fix was:
 - change the Program.components map to store an ARRAY of components for each unique name. The program assumes the first one is the primary. Store these components sorted by pkgPath, which provides consistency and prevents race conditions relating to which component "wins".
 - disconnect and re-attach the dependencyGraph any time a component's position in its own named array is changed (only really applies to duplicate components)
 - dynamically generate the dependencyGraphKey for xml scopes and components, and force the non-primary duplicate components to use a unique dependency graph key (i.e. `component:SomeComponent[1]`)


chrisdp (approved):

TwitchBronBron:
There's something wrong with one of my tests. it runs in an infinite loop. Which is strange, because it passes just fine locally.

chrisdp:
Sounds good. This is going to be such a nice fix once it lands lol.

TwitchBronBron:
The latest master merge broken a few unit tests, which uncovered what I believe to be the root cause of the infinite loop in the CI process. I'm going to trigger the builds a few times to ensure everything is working properly.

If all the builds succeed, then This PR is ready for its final review.

PR #361, opened 2021-03-13
Fix template string transpile bug.

### LINKED ISSUE #355: template strings break when mulitple expressions are back to back
this: 
```brighterscript
return `${m.global.appSettings.imageProcessorEndpoint}${transform}/${uri}`
```

erroneously tranpsiles to :
```brighterscript
return bslib_toString(m.global.appSettings.imageProcessorEndpoint)bslib_toString(transform) + "/" + bslib_toString(uri)
```

TwitchBronBron:
Fixes a bug where transpiling a template string with two side-by-side inline expressions were missing a plus to separate them.
Fixes #355

### DIFF
--- a/src/parser/Expression.ts
+++ b/src/parser/Expression.ts
@@ -1065,7 +1065,10 @@ export class TemplateStringExpression extends Expression {
                     ...items
                 );
             }
-            plus = items.length > 0 ? ' + ' : '';
+            //set the plus after the first occurance of a nonzero length set of items
+            if (plus === '' && items.length > 0) {
+                plus = ' + ';
+            }
         }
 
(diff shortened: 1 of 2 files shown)


src/parser/Expression.ts
```diff
@@ -1065,7 +1065,10 @@ export class TemplateStringExpression extends Expression {
                     ...items
                 );
             }
-            plus = items.length > 0 ? ' + ' : '';
```
TwitchBronBron, line 1068: @georgejecook this was the issue. The plus was being set _back_ to an empty string for the next zero length item. And since the expressions were side-by-side, the next quasi was zero length.

chrisdp (approved):

PR #373, opened 2021-03-25
Retain print separators during transpile

### LINKED ISSUE #321: Missing parts of print statement support
### Issue 1:
Currently when using `bsc` to transpile in `print` statements it always converts the `,` separator to a `;`. 

For example: 
```
	a$="string"
	print a$,a$,a$
```
would become:
```
	a$="string"
	print a$;a$;a$
```

This is what Roku says about the two separators:
`If commas are used, the cursor automatically advances to the next print zone before printing the next item. If semi-colons are used, no space is inserted between the items printed.`

For example this is the difference when printing with the two above syntaxes:
Original: 
```
string          string          string
```
Transpiled: 
```
stringstringstring
```

### Issue 2:
Also currently we are stripping any trailing semi-colons and this should not be the case as per the following:
`A trailing semi-colon over-rides the cursor-return so that the next PRINT begins where the last one left off. If no trailing punctuation is used with print, the cursor drops down to the beginning of the next line.`

Original:
```
	a$="string"
	print a$;a$;a$;
	print a$;a$;a$
```
this results in the following output:
```
stringstringstringstringstringstring
```

Transpiled:
```
	a$="string"
	print a$;a$;a$
	print a$;a$;a$
```
this results in the following output:
```
stringstringstring
stringstringstring
```

### Issue 3:
Lastly it looks like we are injecting unnecessary semi-colons. `Semi-colons can be dropped in some cases. For example, this is legal:`
Example: 
```
Print "this is a five " 5 "!!"
```
result: `this is a five  5!!`

Transpiled:
```
print "this is a five " ; 5 ; "!!"
```
The output in this example is not changed but is unnecessary changes to the code.

Here is the docs on more complex `print` use cases: https://developer.roku.com/docs/references/brightscript/language/program-statements.md#print-item-list

TwitchBronBron:
Fixes #321

### DIFF
--- a/src/files/BrsFile.spec.ts
+++ b/src/files/BrsFile.spec.ts
@@ -2149,7 +2149,7 @@ describe('BrsFile', () => {
                         returnValue = 12 'comment
                         return returnValue 'comment
                     end function 'comment
-                    print "a" ; "b" ; 3 'comment
+                    print "a"; "b"; 3 'comment
                     a(1, 2, 3) 'comment
                     person.functionCall(1, 2, 3) 'comment
                     if true then 'comment
--- a/src/parser/Statement.ts
+++ b/src/parser/Statement.ts
@@ -578,19 +578,22 @@ export class PrintStatement extends Statement {
             ' '
         ];
         for (let i = 0; i < this.expressions.length; i++) {
-            let expression: any = this.expressions[i];
-            if (expression.transpile) {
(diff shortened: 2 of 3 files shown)


chrisdp (approved):

TwitchBronBron:
Finally got the CI passing again.

PR #375, opened 2021-03-28
Unify TranspileState for xml and brs files

TwitchBronBron:
The TranspileState for XmlFile and BrsFile share many similarities. This PR combines most of the functionality into a single class, and then adds the brs-specific items to a new BrsTranspileState used by BrsFiles. 

**Notable changes:**
- `TranspileState.tokenToSourceNode` usage in brs/bs transpile has been replaced with `TranspileState.transpileToken`, which provides more stability for plugin-generated AST nodes with missing ranges.
- `SGTranspileState` has been removed in favor of using the base `TranspileState` (there's nothing unique about XmlFile transpiling)
- Added new `BrsTranspileState` used by Bright(er)Script transpiling, which includes a few small brightscript/brighterscript specific transpile properties.

Running the benchmarks show no noticeable performance difference between this PR and the latest release (v0.37.0).
image


src/parser/TranspileState.ts
```diff
@@ -91,14 +70,46 @@ export class TranspileState {
...
-            this.pathAbsolute,
+            this.srcPath,
             token.text
         );
     }
+
+    /**
+     * Create a SourceNode from a token, accounting for missing range and multi-line text
+     */
+    public transpileToken(token: { range?: Range; text: string }) {
```
TwitchBronBron, line 87: This was mostly copied from SGTranspileState, except I converted the `map()` call to a for loop and call `this.SourceNode` to reuse existing sourceNode logic

src/parser/TranspileState.ts
```diff
@@ -1,87 +1,66 @@
...
 export class TranspileState {
     constructor(
         /**
-         * The BrsFile that is currently being transpiled
+         * The absolute path to the source location of this file. If sourceRoot is specified,
+         * this path will be full path to the file in sourceRoot instead of rootDir.
+         * If the file resides outside of rootDir, then no changes will be made to this path.
          */
-        public file: BrsFile
+        public srcPath: string,
```
TwitchBronBron, line 15: Use name `srcPath` to align with the path variable renames in the upcoming PR #329

PR #376, opened 2021-03-30
Xml ast improvements

TwitchBronBron:
This is part of the large set of breaking changes PR series (including #329). 
Depends on #375 

**Notable changes:**
- Add location tracking (`range` property) to all SceneGraph/XML nodes.
- Remove redundant SG `transpile()` functions in favor of the function from the base class. 
  - This has the added benefit of including all valid XML from the source in the transpiled output. Prior to this PR, only known xml attributes and elements were included, but then had the potential to break any beta features that Roku might be testing.
- Refactored `SGComponent` `scripts`, `children`, and `interface` properties into getter functions which are driven by the underlying XML AST. This means as you modify the AST, those getter functions will automatically reflect the changes. 
- Refactored `SGInterface` properties `fields` and `functions` to getter functions which are driven by the underlying XML AST (same benefits as previous point). 
- Moved `SGParser` module functions onto the SGParser class itself to match the OO patterns in the rest of the project.

The additional location tracking does result in a performance slowdown for XML parsing. However, considering XML files are generally pretty small (compared to brs files) and also considering that xml parsing is still over 200% faster than brs parsing, the XML location tracking is well worth the performance hit. However, I will be running a node profiler to see if I can squeeze out any additional performance. 

image


TwitchBronBron:
I made a few small performance improvements to the SGParser which increased the parse-xml from `84 ops/se`c to `99 ops/sec`, at the expense of a slightly slower transpile speed (from `124 ops/sec` to `100 ops/sec`). So then I modified the `TranspileState.transpileToken` to be monomorphic (i.e. return a `SourceNode` every time instead of sometimes a `string`) and that gave back the lost transpile performance due to v8 optimizations. 

I refactored the logic to find the XML node's bounding range to be computed on-demand instead of during every parse. 

image

elsassph:
We need to ensure also that we support "unknown" tags and attributes everywhere.

src/parser/SGParser.ts
```diff
@@ -129,225 +129,271 @@ export default class SGParser {
...
-            }
-            const cdata = getCdata(content);
-            return new SGScript(name, attributes, cdata, range);
-        case 'children':
-            const childrenContent = mapNodes(content);
-            return new SGChildren(name, childrenContent, range);
-        default:
-            const nodeContent = mapNodes(content);
-            return new SGNode(name, attributes, nodeContent, range);
+    mapElement({ children }: ElementCstNode): SGTag {
```
elsassph, line 168: I don't like seeing all these pure functions moved into the classe but well...

elsassph (approved):

TwitchBronBron:
Closed in favor of #400

PR #385, opened 2021-04-09
Code action callfunc operator refactor

### LINKED ISSUE #240: provide codeAction in .bs when callfunc can be refactored to callfunc operator
would be great if the parser could identify when brs syntax could be upgraded to bs syntax.

for example, consider:

`node.callFunc("funcName", invalid)`

this could be :
`node@.funcName()`

TwitchBronBron:
Adds code action in brighterscript files to suggest refactoring a normal callfunc call into using a callfunc operator. 

I don't produce any diagnostics for this, as it's not the compiler's job to suggest such a preference-based thing. However, since it's easy enough to provide hints when you click on it, i think that's a fair compromise. 

Resolves #240 

**Notable changes:**
 - add codeAction to refactor callfunc statement to use the callfunc operator.
 - fixed bug in code action utils that wasn't supporting update. Also added delete support.

### DIFF
--- a/src/bscPlugin/codeActions/CodeActionsProcessor.spec.ts
+++ b/src/bscPlugin/codeActions/CodeActionsProcessor.spec.ts
@@ -225,6 +225,51 @@ describe('CodeActionsProcessor', () => {
                 `import "pkg:/source/Animals.bs"`
             ]);
         });
+
+        it('suggests callfunc operator refactor', () => {
+            //define the function in two files
+            const file = program.addOrReplaceFile('components/main.bs', `
+                sub doSomething()
+                    someComponent.callfunc("doThing")
+                    someComponent.callfunc("doOtherThing", 1, 2)
+                end sub
+            `);
+            program.validate();
+
+            expectCodeActions(() => {
+                program.getCodeActions(
+                    file.pathAbsolute,
+                    util.createRange(2, 39, 2, 39)
+                );
+            }, [{
+                title: `Refactor to use callfunc operator`,
+                isPreferred: false,
+                kind: 'quickfix',
+                changes: [{
+                    filePath: file.pathAbsolute,
+                    newText: '@.doThing(',
+                    type: 'replace',
+                    range: util.createRange(2, 33, 2, 52)
+                }]
+            }]);
+
+            expectCodeActions(() => {
+                program.getCodeActions(
+                    file.pathAbsolute,
+                    util.createRange(3, 39, 3, 39)
+                );
+            }, [{
+                title: `Refactor to use callfunc operator`,
+                isPreferred: false,
+                kind: 'quickfix',
+                changes: [{
+                    filePath: file.pathAbsolute,
+                    newText: '@.doOtherThing(',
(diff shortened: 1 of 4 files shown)


src/bscPlugin/codeActions/CodeActionsProcessor.spec.ts
```diff
@@ -225,6 +225,51 @@ describe('CodeActionsProcessor', () => {
                 `import "pkg:/source/Animals.bs"`
             ]);
         });
+
+        it('suggests callfunc operator refactor', () => {
+            //define the function in two files
+            const file = program.addOrReplaceFile('components/main.bs', `
+                sub doSomething()
+                    someComponent.callfunc("doThing")
```
georgejecook, line 233: nobody would actually have this code, coz it would not work (it woudl do nothing at run time).

It would have to be `someComponent.callfunc("doThing", invalid)`, which (imho) should become `someComponent@.doThing()` (which is semantically equivalent).

georgejecook (changes requested):
left a comment re: unit test, which I think needs a tweak.

src/bscPlugin/codeActions/CodeActionsProcessor.spec.ts
TwitchBronBron: While I understand the "standard", I've seen callfunc in the wild with no second parameter. I believe @elsassph also had mentioned they use it like that in their code successfully. 

However, I agree, we should refactor `someComponent.callfunc("doThing", invalid)` to `someComponent@.doThing()` as well, so I'll update this PR to support that as well.

TwitchBronBron:
I've implemented your requested changes. Please review again.

src/bscPlugin/codeActions/CodeActionsProcessor.spec.ts
elsassph: `callfunc` without argument should be a warning, but I have seen it.

georgejecook: yeah, guess you guys are right, if there's a default param on the function, then in that case callfunc without `invalid` will work (I wasn't aware of this or it changed since I got this notion > 2 years ago). Clearly, we can never know if this is the case, so 🤷 I'm not sure if even have a warning here..

georgejecook (approved):
Don't know what the implications are for not needing to pass an invalid argument are, or whether to have a warning or not, so I say just go ahead and merge.

TwitchBronBron:
It's probably okay to merge. I wanted to think about the API structure a bit, but it's probably fine. 🫠

TwitchBronBron:
Oh, i remember why. I wanted to think about that callfunc with the second parameter thing. We should chat about this on Tuesday to flush out the details.

chrisdp:
Sure. I think the TLDR from me is any change to how we handle the second param/lack there of should be in the V1 release. Also need to check what logging we get if it is missing and can the developers easily self diagnose when there is a miss match. (Mostly notes for my self for when we chat)

PR #386, opened 2021-04-17
Upgrade chokidar to support M1 macs

TwitchBronBron:
I heard that the bsc watch mode doesn't work with M1 macs. I don't have a mac, but I did find [this GitHub issue](https://github.com/paulmillr/chokidar/issues/1048) stating that a recent chokidar release supports mac M1, so i'm upgrading BSc to use that version. It shouldn't cause any issues? I'll need someone with an M1 mac to test this build out.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -536,13 +536,12 @@
             "dev": true
         },
         "@types/chokidar": {
-            "version": "1.7.5",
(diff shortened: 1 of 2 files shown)


chrisdp (approved):
Tested on my M1 MacBook Pro with both watchers and without and looks good. No issues found.

PR #387, opened 2021-04-20
File api

TwitchBronBron:
Moves file handling into plugin-land so plugins can contribute custom file types. This PR depends on #330 .

Notable changes:
- added plugin event `onProvideFile`. If a plugin returns a file, that file is used. Otherwise, move on to next plugin until we run out. The program will handle all non-plugin-handled files.
- every single file found in the `files` array that actually exists on disk will be included in the program. (this will help support the custom tree view we wanted to add in vscode but couldn't because non xml and brs files were missing)
- make almost all `BscFile` functions and properties optional so plugin-contributed files only need to write what they need.
- adds several file contribution points for both xml and brs file types, so any file could theoretically contribute all of that information (work-in-progress)


TwitchBronBron:
Closed in favor of #408

PR #393, opened 2021-04-20
Plugin support for hover

TwitchBronBron:
Moves the `getHover` functionality into the plugin API so plugins can participate in providing hover support. 

**Notable changes:**
 - add new `beforeProvideHover`, `provideHover`, `afterProvideHover` plugin events. 
 - Plugins can push 0 or more Hovers to the `event.hovers` collection. 
 - All hovers are deduped, and then merged into a single markdown string each separated by a single newline. This allows multiple plugins to provide hovers for a given position. If plugins don't want to step on each other, they can first check that the `event.hovers` is empty.
 - ~~modify the plugin event pipleline to short-circuit if the handler returns `false`. This is useful for `onGetHover` because only a single plugin may contribute a hover, so the first one that contributes needs to indicate they did so.~~
 - move `BrsFile` `getHover` functionality into the `BscPlugin`. I didn't change the core logic, I just copied the code and made it work within the new object.


src/PluginInterface.ts
```diff
@@ -21,9 +21,14 @@ export default class PluginInterface<T extends CompilerPlugin = CompilerPlugin>
...
             if ((plugin as any)[event]) {
                 try {
-                    this.logger.time(LogLevel.debug, [plugin.name, event], () => {
-                        (plugin as any)[event](...args);
+                    const returnValue = this.logger.time(LogLevel.debug, [plugin.name, event], () => {
+                        return (plugin as any)[event](...args);
                     });
+
+                    //plugins can short-circuit the event by returning `false`
+                    if (returnValue === false) {
```
elsassph, line 29: Should that be the opposite - return `true` if captured?

src/interfaces.ts
```diff
@@ -182,33 +182,34 @@ export type CompilerPluginFactory = () => CompilerPlugin;
...
-    afterProgramTranspile?: (program: Program, entries: TranspileObj[]) => void;
+    beforeProgramCreate?: (builder: ProgramBuilder) => void | boolean;
+    beforePrepublish?: (builder: ProgramBuilder, files: FileObj[]) => void | boolean;
+    afterPrepublish?: (builder: ProgramBuilder, files: FileObj[]) => void | boolean;
+    beforePublish?: (builder: ProgramBuilder, files: FileObj[]) => void | boolean;
+    afterPublish?: (builder: ProgramBuilder, files: FileObj[]) => void | boolean;
+    afterProgramCreate?: (program: Program) => void | boolean;
+    beforeProgramValidate?: (program: Program) => void | boolean;
+    afterProgramValidate?: (program: Program) => void | boolean;
+    beforeProgramTranspile?: (program: Program, entries: TranspileObj[]) => void | boolean;
```
elsassph, line 193: I don't think all the events should be cancellable...

src/PluginInterface.ts
TwitchBronBron: I following the design of the browser event model, where returning false cancels the entire chain. https://stackoverflow.com/a/30473685/1633757
I think of it as `"shouldContinue?: true|false"`. 

I'm open to switching it still if you feel strongly about it, but `false` makes more sense to me.

src/interfaces.ts
TwitchBronBron: fair enough. I can remove that from these interfaces that aren't applicable. However, we don't have a way to limit this in the actual PluginInterface right now until #330 lands (because we can add a third `isCancellable` parameter to `emit` once the second param is an event object).

TwitchBronBron: Reverted all except the `onGetHover` one.

src/PluginInterface.ts
TwitchBronBron: @georgejecook @triwav @chrisdp care to weigh in on this?

georgejecook: This is what I'd expect, too (false to short-circuit), for this kind of mechanism; but what are we cancelling here? It looks like we are preventing any other plugins getting events here, right?

src/interfaces.ts
elsassph: There could also be an `emitCancellable` in addition to `emit`

elsassph: I think for `onHover` it's not about cancelling but confirming that it's handled, as when we confirm a key event is handled.

src/PluginInterface.ts
elsassph: I think for `onHover` it's not about cancelling but handling, as when we confirm a key event is handled.

elsassph: That said ultimately I don't have a strong opinion on either options.

georgejecook: when I read this, it looks like if I return false, then no other plugin will be notified of this event, as we exit the function, so I'm not reading this as opt-in to handle event; but return false, and no more plugins events.

image

imo: If it's how I read it, it _should_ be false, if it's opt-in, like @elsassph says, it should be true

georgejecook: I'm not a fan of this though; as I could potentiallly have to consdier other people's plugins, when debugging my own. what's the use case?

src/interfaces.ts
georgejecook: is it not more common to have an cancellation token?

src/PluginInterface.ts
TwitchBronBron: The use case is in the description of this pr: 
>This is useful for onGetHover because only a single plugin may contribute a hover, so the first one that contributes needs to indicate they did so.

For `onGetCompletions`, every plugin may contribute (add to) the result...it's cumulative. However, for `onGetHover`, the language server protocol only supports a single `hover` result. Which means we should stop looking for a hover result the moment we find one. that's what this is for. the first plugin who responds to this, wins.

triwav: I think most of my thinking comes from the fact that you return true on Roku to say you handled something and it shouldn't continue up the node tree. One other idea. This would take more resources but if instead of exiting out on the first you let all the plugins go. Each one can return a priority on it and the highest one wins. Not huge fan of it as if we ever have community plugins there'd need to be an agreed upon priority scale for it.

elsassph: Yes @triwav that's also what I'm thinking here - it's not a "cancellation" pattern.
We want plugins to say, for some events, "I have handled the event, stop propagating" like Roku key events.
And to avoid weird issues I'd also suggest having a separte `emitUntilHandled` method.

georgejecook: Given that, I agree @elsassph, I'd expect to return true for this.

TwitchBronBron: I entirely refactored this logic so that plugins can all contribute one or more hovers, and we merge all the text together into a single response to the client. In this way, we no longer need to short-circuit at all. A plugin can decide to contribute or not based on whether the `event.hovers` array is empty.

chrisdp (approved):

PR #394, opened 2021-04-21
Allow class field overrides, disallow type changes

### LINKED ISSUE #391: should be able to override class fields
Currently classes show a diagnostic whenever a parent and child class have the same property name. I intentially designed the classes this way, but now I'm not seeing much benefit in doing that. 

[here's the code](https://github.com/rokucommunity/brighterscript/blob/d94bfc86b3e2106be2856df5919468770ba828a6/src/validators/ClassValidator.ts#L198). 

We should definitely show diagnostic errors whenever the property types are different between child and parent. However, if the properties are the same type, a child should absolutely be able to re-declare the same property name. 

For example:
```brighterscript
class Parent
    public name = "Mommy"
    public age = 30
end class
class Child
    public name = "Baby" 'this currently produces a diagnostic, but should not
    public age = "0.2" ' this should show a diagnostic because the types aren't the same (don't think it does currently...)
end class
```

Thoughts?

TwitchBronBron:
Fixes #391.

**Notable changes:**
- allow class fields to be redeclared in child classes as long as the types are the same
- flag type differences between parent and child class fields with same name
- allow custom types for class field types

### DIFF
--- a/src/DiagnosticMessages.ts
+++ b/src/DiagnosticMessages.ts
@@ -511,8 +511,8 @@ export let DiagnosticMessages = {
         code: 1097,
         severity: DiagnosticSeverity.Error
     }),
-    memberAlreadyExistsInParentClass: (memberType: string, parentClassName: string) => ({
-        message: `A ${memberType} with this name already exists in inherited class '${parentClassName}'`,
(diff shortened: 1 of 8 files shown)


TwitchBronBron:
@markwpearce I'd appreciate your feedback on this PR as well.

markwpearce:
I like this change. Makes it cleaner to set properties in class declaration instead of in constructor.

I don't know the code well enough to  review it though!

PR #395, opened 2021-04-21
Feature/scope tracked symbol tables

markwpearce:
Adds scope based and namespace based symbol tables

Adds validation on arguments in function calls


markwpearce:
Sorry about all the whitespace changes in src/Scope.spec.ts ... for some reason they got all screwed up, and I did my best to change them back

src/Scope.ts
```diff
@@ -459,6 +470,33 @@ export class Scope {
...
+        if (this._symbolTable) {
+            for (const childTable of this._symbolTable.children) {
+                childTable.setParent(null);
+            }
+        }
+        this._symbolTable = null;
+    }
+
+    /**
+     * Gets ths current symbol table for the scope, by merging the tables for all the files in this scope.
```
TwitchBronBron, line 486: spelling: `ths`

src/Scope.ts
```diff
@@ -343,8 +347,10 @@ export class Scope {
                         namespaces: {},
                         classStatements: {},
                         functionStatements: {},
-                        statements: []
+                        statements: [],
+                        symbolTable: new SymbolTable(parentSymbolTable)
                     };
+                    parentSymbolTable = namespaceLookup[lowerLoopName].symbolTable;
```
TwitchBronBron, line 353: Could you explain this a little bit? I _think_ this is setting every namespace part to the previous part. 
i.e. if you have `namespace a.b.c`, then `a.b.c`'s parent would be `a.b`, and `a.b`'s parent would be `a`. Is that right? If so, what benefit does that provide?

src/SymbolTable.ts
```diff
@@ -65,15 +89,15 @@ export class SymbolTable {
...
             //Check if each time it was set, it was set to the same type
             // TODO handle union types
-            let sameType = true;
-            const firstType = symbols[0].type.toString();
+            let sameImpliedType = true;
+            let impliedType = symbols[0].type;
             for (const symbol of symbols) {
-                sameType = firstType === symbol.type.toString();
-                if (!sameType) {
+                sameImpliedType = (impliedType.toString() === symbol.type.toString());
```
TwitchBronBron, line 95: You don't have to address this right now, but we should probably have a `.equals()` method on each `Type` so we can compare them more precisely.

src/parser/Expression.ts
```diff
@@ -124,6 +125,7 @@ export class FunctionExpression extends Expression implements TypedefProvider {
     readonly symbolTable: SymbolTable;
 
     constructor(
+        name: Identifier,
```
TwitchBronBron, line 128: I don't think `name` belongs here. If a function has a name, the `FunctionStatement` has that name. Otherwise, it's an anonymous function expression and doesn't actually have a name. So in either case, `FunctionExpression` shouldn't have a name.

If you must have the name, then we should pass in the `FunctionStatement` that owns this expression.

TwitchBronBron (changes requested):
Looks really great! I left a few comments.

src/Scope.ts
markwpearce: Yep.. this is an error. all of these new symbol tables should be parented to the scope

src/Scope.spec.ts
```diff
@@ -791,7 +920,7 @@ describe('Scope', () => {
 
     describe('getDefinition', () => {
         it('returns empty list when there are no files', () => {
-            let file = program.addOrReplaceFile({ src: `${rootDir}/source/main.brs`, dest: 'source/main.brs' }, '');
+            let file = program.addOrReplaceFile({ src: `${rootDir} /source/main.brs`, dest: 'source/main.brs' }, '');
```
TwitchBronBron, line 923: This added a space in the path, which is completely bogus. Do you think you could go through and undo these additional unnecessary whitespace changes in this file?

src/SymbolTable.ts
```diff
@@ -12,6 +12,8 @@ export class SymbolTable {
 
     private symbols = new Map<string, BscSymbol[]>();
 
+    public readonly children: SymbolTable[] = [];
```
TwitchBronBron, line 15: I don't think I like this `children` concept. I know it makes things slightly easier during cleanup. However, every a user types in a .brs file, the previous `BrsFile` is destroyed and a brand new `BrsFile` is created. Which means that we could potentially hold on to those old references for a while in these scopes. 

I think instead, the scope should wire itself up to all children at the start of `validate()` and then unwire at the end of `validate()`.  I believe currently the "children" would persist after the validate is finished.

src/util.ts
```diff
@@ -1139,6 +1139,27 @@ export class Util {
             source: 'brs'
         };
     }
+
+    /**
+     * Finds the array of callables from a container map, taking into account the function from which it was called
+     * If the callable was called in a function in a namespace, functions in that namespace are preferred
+     * @return an array with callable containers - could be empty if nothing was found
+     */
+    public getCallableContainersFromContainerMapByFunctionCall(callablesByLowerName: CallableContainerMap, expCall: FunctionCall): CallableContainer[] {
```
TwitchBronBron, line 1148: This name is crazy! Is there any way we could find a shorter name without lots of abbreviations?

TwitchBronBron (changes requested):
I added some more comments.

src/Scope.spec.ts
```diff
@@ -806,4 +935,122 @@ describe('Scope', () => {
             expect(completions.filter(x => !!x.documentation)).to.have.length.greaterThan(0);
         });
     });
+
+    describe('scope symbol tables', () => {
+        it('adds symbols for the complete scope', () => {
+            program.addOrReplaceFile('source/file.brs', `
+            function funcInt() as integer
```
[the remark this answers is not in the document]
markwpearce: Ugh... I probably missed a bracket somehere when I saved and autoformat whent crazy.

src/util.ts
markwpearce: No... It;s impossible

TwitchBronBron:
Looks great! Thanks for all the changes.

PR #396, opened 2021-04-22
Plugin semantic highlighting

TwitchBronBron:
Adds plugin support for semantic highlighting. Also adds semantic highlighting for in-scope class names for `new` expressions and CustomType parameter types.

ac359b92-f478-4ba4-8a0e-ba50d225e4a6


src/SemanticTokenUtils.ts
```diff
@@ -0,0 +1,72 @@
...
+    util.sortByRange(tokens);
+    const builder = new SemanticTokensBuilder();
+    for (const token of tokens) {
+        builder.push(
+            token.range.start.line,
+            token.range.start.character,
+            token.range.end.character - token.range.start.character,
+            semanticTokensLegend.tokenTypes.indexOf(token.tokenType),
+            //modifier bit flags (TODO implement)
+            0
```
TwitchBronBron, line 68: Still need to implement this before we can merge.

TwitchBronBron: This has been implemented.

PR #400, opened 2021-04-23
Xml ast improvements

TwitchBronBron:
Migrated from #376 

**Notable changes:**
- Add location tracking (`range` property) to all SceneGraph/XML nodes.
- Remove redundant SG `transpile()` functions in favor of the function from the base class. 
  - This has the added benefit of including all valid XML from the source in the transpiled output. Prior to this PR, only known xml attributes and elements were included, but then had the potential to break any beta features that Roku might be testing.
- Refactored `SGComponent` `scripts`, `children`, and `interface` properties into getter functions which are driven by the underlying XML AST. This means as you modify the AST, those getter functions will automatically reflect the changes. 
- Refactored `SGInterface` properties `fields` and `functions` to getter functions which are driven by the underlying XML AST (same benefits as previous point). 
- Moved `SGParser` module functions onto the SGParser class itself to match the OO patterns in the rest of the project.


TwitchBronBron:
@elsassph I know you approved the PR against master (#376), but I've made a few additional changes to this PR so I'd appreciate another review when you get a chance. Changes since your last review:

 - moved all xml tokens into `.tokens` object on SGNode or SGAttribute. 
 - renamed `SGNode.children` array to `SGNode.childNodes` so it doesn't conflict with `SGComponent.children` which points to a `<children>` element.
 - added  `fields`,`functions`, and `members` properties to `SGInterface` and made them all read-only because SceneGraph supports having multiple `<interface>` elements in the file and merges their fields and functions together. 
- added `SGNode.removeField()` function

src/XmlScope.ts
```diff
@@ -52,48 +52,52 @@ export class XmlScope extends Scope {
     }
 
     private diagnosticValidateInterface(callableContainerMap: CallableContainerMap) {
-        if (!this.xmlFile.parser.ast?.component?.api) {
+        const component = this.xmlFile.parser.ast?.component;
+        //roku supports multiple <interface> elements, so check that there's at least one
+        if (!component || component.getChildNodesByTagName('interface').length === 0) {
```
elsassph, line 57: Why not:
```
const interfaceMembers = component?.interfaceMembers || [];
if (interfaceMembers.length === 0) { return; }
```

src/files/XmlFile.spec.ts
```diff
@@ -58,13 +58,13 @@ describe('XmlFile', () => {
             program.plugins.add({
                 name: 'allows modifying the parsed XML model',
                 afterFileParse: () => {
-                    let child = file.parser.ast.component.children.children[0];
+                    let child = file.parser.ast.component.children.childNodes[0];
                     expect(child.attributes).to.have.lengthOf(4);
                     child.setAttribute('text', undefined);
-                    expect(child.getAttribute('id').value.text).to.equal('one');
+                    expect(child.getAttribute('id').tokens.value.text).to.equal('one');
```
elsassph, line 64: `child.getAttribute('id').value` maybe? so you also test the getter.

src/parser/SGTypes.ts
```diff
@@ -11,19 +10,152 @@ export interface SGToken {
...
+     * The array of direct children AST elements of this AST node
+     */
+    public childNodes = [] as SGTag[];
+
+    public get range() {
+        if (!this._range) {
+            this._range = util.createBoundingRange(
+                this.tokens.startTagOpen,
+                this.tokens.startTagName,
+                this.attributes?.[this.attributes?.length - 1],
```
elsassph, line 141: Could we simplify by assuming `attributes` and `childNodes` should never be null?

src/parser/SGTypes.ts
```diff
@@ -32,120 +164,128 @@ export class SGTag {
...
+    }
+
     getAttribute(name: string): SGAttribute | undefined {
-        return this.attributes.find(att => att.key.text.toLowerCase() === name);
+        return this.attributes.find(att => att.tokens.key.text.toLowerCase() === name);
     }
 
     getAttributeValue(name: string): string | undefined {
-        return this.getAttribute(name)?.value?.text;
+        return this.getAttribute(name)?.tokens.value?.text;
```
elsassph, line 213: Why not using the getter here?

src/parser/SGTypes.ts
```diff
@@ -161,34 +301,35 @@ export class SGScript extends SGTag {
...
-    protected transpileAttributes(state: TranspileState, attributes: SGAttribute[]): (string | SourceNode)[] {
+    protected transpileAttributes(state: TranspileState, attributes: SGAttribute[]) {
         const modifiedAttributes = [] as SGAttribute[];
         let foundType = false;
         const bsExtensionRegexp = /\.bs$/i;
 
         for (const attr of attributes) {
-            const lowerKey = attr.key.text.toLowerCase();
-            if (lowerKey === 'uri' && bsExtensionRegexp.exec(attr.value.text)) {
+            const lowerKey = attr.tokens.key.text.toLowerCase();
```
elsassph, line 325: This code assumes tokens are all valid.

src/parser/SGTypes.ts
```diff
@@ -323,61 +451,72 @@ export class SGInterface extends SGTag {
...
-                    this.children = tag;
-                } else {
-                    throw new Error(`Unexpected tag ${tag.tag.text}`);
-                }
-            }
+
+    /**
+     * Get all the <Field> and <Function> elements across all <Interface> nodes in this component
+     */
+    public get interfaceMembers() {
```
elsassph, line 498: API to find an `interface` node seem inconvenient (I assume it requires to use `getChildNodesByTagName<SGInterface>('interface')`). Adding fields/functions is going to be a major use case for the XML AST manipulations.

src/parser/SGParser.ts
```diff
@@ -49,28 +49,28 @@ export default class SGParser {
...
 
         const { component } = this.ast;
         if (!component) {
             return;
         }
 
         const nameAttr = component.getAttribute('name');
-        if (nameAttr?.value) {
-            this._references.name = nameAttr.value;
+        if (nameAttr?.tokens.value) {
```
elsassph, line 60: Pattern could probably be simplified into `this._references.name = nameAttr?.value;` without the `if` guard.

src/parser/SGTypes.ts
TwitchBronBron: I'll make a few tweaks. Good catch!

src/parser/SGTypes.ts
TwitchBronBron: Getters are technically function calls in javascript, so I figured this would be slightly more efficient.

src/parser/SGTypes.ts
TwitchBronBron: Yeah, good point. I worry about plugin authors creating bad AST Nodes. However, I'll update the constructor function to set these to empty array if passed in as null, so that should be safe enough.

src/files/XmlFile.spec.ts
TwitchBronBron: Will do.

src/XmlScope.ts
TwitchBronBron: Yeah, I have no idea why I did it this way! (perhaps I was just resolving tsc errors and didn't think through optimizing the code). This has been updated.

src/parser/SGTypes.ts
TwitchBronBron: The SGInterface element has these functions: `setField`, `removeField`, `setFunction`, `removeFunction`. However, you are right that authors would need to call `getChildNodesByTagName<SGInterface>('interface')` and use the first interface that they find. So what do you propose?  `setInterfaceField`, `removeInterfaceField`, `setInterfaceFunction`, `removeInterfaceFunction` functions on `SGComponent` that auto-picks the first `SGInterface` it finds, or creates one if missing?

src/parser/SGParser.ts
TwitchBronBron: Done.

TwitchBronBron:
@elsassph I believe I've addressed all of your concerns except we need to discuss [this comment](https://github.com/rokucommunity/brighterscript/pull/400#discussion_r620144022).

src/parser/SGTypes.ts
TwitchBronBron: @elsassph  I went ahead and added a bunch of interface-related methods to SGComponent. [Here's the commit](https://github.com/rokucommunity/brighterscript/commit/52cc44ef3c06bc13a9c10c633a28f38f545dd5e1). Do these suit your needs?

PR #402, opened 2021-04-26
adds warning for mismatched class member accessibility

georgejecook:


src/validators/ClassValidator.ts
```diff
@@ -223,6 +224,24 @@ export class BsClassValidator {
                                 range: member.range
                             });
                         }
+
+                        //child member has different visiblity
+                        if (
+                            //is a method
+                            isClassMethodStatement(member) &&
```
TwitchBronBron, line 231: Why only class methods? Fields should also be included in this. TypeScript applies this restriction to both methods and fields.

TwitchBronBron (changes requested):
Looks good. I asked for a few small changes.

Also, the github actions build is failing right now because coveralls (our coverage tool) has a server-side error. I'll keep an eye on that.

src/validators/ClassValidator.ts
georgejecook: I did this as we don't support overriding field yet, do we?

TwitchBronBron:
Looks good. Thanks!

PR #407, opened 2021-05-07
Move token-specific functions into `Parser.ts`

TwitchBronBron:
Move all of the token-specific functions out of `BrsFile.ts` and into `Parser.ts`


markwpearce:
Nice! I support this change 👍

PR #409, opened 2021-05-07
Class Symbol Tables

markwpearce:
**Features**

Validations:
- Type checking in function arguments includes class types
- Type checking in class method invocations and super() calls
- Argument count mismatch validations in class method invocations and super() calls
- Type checking for class types includes looking for ancestors

Hover/Intellisense:
- Declarations have hovers: Class, class methods, class fields
- Completion on member fields that are classes
- Return type on class methods is known

Probably a few others...


markwpearce:
More work needs to be done on this:
- Class/object symbol tables need to be built at link-time (instead of parse time)
- Namespace hovers and other namespace related issues (validation of assignment to namespaced classes, for example)
- Add symbols tables to SG components 
- Add symbol tables to AssociativeArrays
- Unified symbol type lookup across objects/classes/components/interfaces

markwpearce:
Still to do:

- Class/object symbol tables need to be built at link-time (instead of parse time)
- Handling function chains being called on primitive types. Eg. `myStr.trim().toInt().toStr()`
- Handling token chains with array referencing. Eg. `myIntArray[0].toStr()`

Future work (for another PR):
- Add symbols tables to SG components
- Add symbol tables to AssociativeArrays
- Unified symbol type lookup across objects/classes/components/interfaces

Other Future Work (for yet another PR):
- Add all of the data from Roku Developer Reference website as types

markwpearce:
I actually think this is ready for review?

markwpearce:
Apart from the test failure (easy fix),
I loaded set this branch as the active Brighterscript version in the vs code extension in our main production code base, and there was a new error that crashed the language server. 

There seems to be an error resolving LazyTypes with circular references. 

I have a fix in mind, which would also add type caching. I should be able to get that in sometime next week, although if we want to merge this as is, with that known issue, it’s cool with me. 

I am sorry (not sorry) that this is such a huge PR! So much needed to be touched to get types everywhere.

src/parser/Expression.ts
```diff
@@ -661,18 +663,33 @@ export class AAMemberExpression extends Expression {
...
 
+    readonly symbolTable: SymbolTable = new SymbolTable();
+    readonly memberTable: SymbolTable = new SymbolTable();
+
     public readonly range: Range;
 
+    public buildSymbolTable() {
+        this.symbolTable.clear();
+        this.symbolTable.addSymbol('m', { start: this.open.range.start, end: this.close.range.end }, new ObjectType(this.memberTable));
+        for (const element of this.elements) {
```
elsassph, line 686: Should/could it add the `m.x* = ...` found in its function values?

src/types/helpers.ts
```diff
@@ -0,0 +1,107 @@
...
+        if (isFunctionType(currentKnownType)) {
+            return currentKnownType.returnType;
+        }
+        if (currentKnownType) {
+            // this will probably only happen if a functionName has been assigned to something else previously?
+            return currentKnownType;
+        }
+        return new LazyType((context?: TypeContext) => {
+            // Give best guess if there is no file context
+            let futureType = functionExpression.symbolTable.getSymbolType(calleeName.text);
```
elsassph, line 32: Should that be resolved if the next `if` doesn't resolve a type?

src/parser/Parser.ts
```diff
@@ -1785,13 +1789,15 @@ export class Parser {
...
             } else if (isDottedGetExpression(left)) {
-                return new DottedSetStatement(
+                const dottedSetStmt = new DottedSetStatement(
                     left.obj,
                     left.name,
                     operator.kind === TokenKind.Equal
                         ? right
                         : new BinaryExpression(left, operator, right)
                 );
+                this._references.dottedSetStatements.push(dottedSetStmt);
```
elsassph, line 1799: I think you'll collect dottedSets in anonymous functions attached potentially to AA objects and not to the top `m`. I think this risky and instead you should have a post-parsing walk of the statements of the top-level functions.

src/parser/Parser.ts
```diff
@@ -2718,6 +2727,70 @@ export class Parser {
...
+    /**
+     * Finds the previous identifier in a chain (e.g. 'm.obj.func(someFunc()).value'), skipping over any arguments of function calls
+     * If this function was called with the token at 'value' above, the previous identifier in the chain is 'func'
+     * @param currentToken token to start from
+     * @param allowCurrent can the current token be the token that's the identifier?
+     * @returns the previous identifer
+     */
+    public getPreviousIdentifierInChain(currentToken: Token, allowCurrent = false): Token {
+        if (!allowCurrent) {
+            currentToken = this.getPreviousToken(currentToken);
```
elsassph, line 2739: `getPreviousToken` does a `parser.tokens.indexOf(token)` so this function may be very inefficient in very long scripts. You should maybe get the index of the token, then walk back from it?

elsassph (commented):
Did one pass at reviewing the code. Consider that a fairly superficial review :D for now.

src/files/BrsFile.Class.spec.ts
```diff
@@ -1178,4 +1186,340 @@ describe('BrsFile BrighterScript classes', () => {
...
+            expect(hover).to.exist;
+            expect(hover.contents).to.equal('m as Dog');
+        });
+
+        it('gets the correct text for m.field', () => {
+            let animalCode = program.setFile('source/animal.bs', animalClassCode);
+            program.validate();
+            let hover = animalCode.getHover(Position.create(6, 19));
+            expect(hover).to.exist;
+            expect(hover.contents).to.equal('Animal.kind as string');
```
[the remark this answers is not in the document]
markwpearce: I was just following Typescript's example - it doesn't differentiate between static and not in hovers

src/files/BrsFile.ts
```diff
@@ -679,17 +702,239 @@ export class BrsFile {
...
-        if (previousToken?.kind === TokenKind.Identifier && previousToken?.text.toLowerCase() === 'm' && isClassMethodStatement(functionExpression.functionStatement)) {
-            return { item: this.parser.references.classStatements.find((cs) => util.rangeContains(cs.range, position)), file: this };
+    /**
+     * Gets the class (if any) of a given token based on the scope
+     * @param currentToken token in question
+     * @param functionExpression current functionExpression
+     * @param scope the current scope
+     * @returns A fileLink of the ClassStatement, if it is a class, otherwise undefined
+     */
+    public getClassFromToken(currentToken: Token, functionExpression: FunctionExpression, scope: Scope): FileLink<ClassStatement> | undefined {
```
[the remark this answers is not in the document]
markwpearce: Agree in principle, however, there may be different completions that are necessary for different types (e.g. private methods, etc)

src/parser/Expression.ts
markwpearce: yeah ... we need to provide the correct symbol table context for those functions, so when they are parsed they add to "m" automatically... hmm...

src/parser/Parser.ts
markwpearce: Yeah, good Idea. I did not think of that

src/Program.ts
```diff
@@ -846,7 +846,12 @@ export class Program {
 
         //get the completions from all scopes for this file
         let allCompletions = util.flatMap(
-            scopes.map(ctx => file.getCompletions(position, ctx)),
+            scopes.map(ctx => {
+                ctx.linkSymbolTable();
```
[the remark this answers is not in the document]
markwpearce: The "most important" symbol checking function is `getSymbolTypeFromToken` which is currently in `BscFile` ... I'm thinking I should actually move that to `Scope`, and letting scope decide if it needs to link/unlink any symbol tables. (perhaps through some static reference to the currently linked scope)

Also, I think I should cache the result of `getSymbolTypeFromToken` in `Scope`, so that if there is subsequent call while the symbolTables are still linked with the same token, then it could just return the cached version.

What do you think?

src/Scope.ts
```diff
@@ -503,15 +549,32 @@ export class Scope {
...
 
                 for (const namespace of file.parser.references.namespaceStatements) {
                     const namespaceNameLower = namespace.nameExpression.getName(ParseMode.BrighterScript).toLowerCase();
                     const namespaceSymbolTable = this.namespaceLookup[namespaceNameLower].symbolTable;
                     namespace.symbolTable.setParent(namespaceSymbolTable);
                 }
+                //TODO: build symbol tables for dotted set assignments?
+                // eg. m.key = "value"
+
+                for (const dotSetStmt of file.parser.references.dottedSetStatements) {
```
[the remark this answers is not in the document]
markwpearce: `m` exists in the source scope too.

See below for my idea for the caching strategy

src/Scope.ts
```diff
@@ -636,36 +702,98 @@ export class Scope {
...
+                    for (let index = 0; index < funcType.params.length; index++) {
+                        const param = funcType.params[index];
+                        const arg = expCall.args[index];
+                        if (!arg) {
+                            // not enough args
+                            break;
+                        }
+                        let argType = arg.type ?? new UninitializedType();
+                        let assignable = false;
+                        const typeContext = { file: file, scope: this };
```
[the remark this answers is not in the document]
markwpearce: This iteration is just checking the types of the arguments vs parameters in this file/scope context. Other scopes using this file will ALSO check against their context.
The file/scope context of a function call/function declaration do not change on a arg-to-arg basis

src/XmlScope.ts
```diff
@@ -39,6 +41,38 @@ export class XmlScope extends Scope {
...
+            if (isXmlScope(parentScope)) {
+                this._topTable.setParent(parentScope.topTable);
+            }
+        }
+        return this._topTable;
+    }
+
+    public get memberTable() {
+        if (!this._memberTable) {
+            this._memberTable = new SymbolTable(this.getParentScope()?.memberTable);
```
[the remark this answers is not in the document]
markwpearce: I forgot to make it delete _memberTable on nscope invalidation. Thanks!

markwpearce:
As of the latest applying this version of the parser/validator to our current huge production codebase, there are three "false positive" validation issues:

A member function with same name as global Callable is called after a bracket-ed expression:

```vb
twoStr = (1+1).toStr()
```

`for-each` loops that loop over an array that is returned from a function:

```vb
 for each child in  m.navGroup.getChildren(-1, 0) 
   ...
 end for
```

Using multiple square brackets in chain:

```
m.obj["property"]["subProp"].funcCall()
```

src/SymbolTable.ts
```diff
@@ -26,6 +26,17 @@ export class SymbolTable {
         return [].concat(...this.symbolMap.values());
     }
 
+    /**
+     * Get list of all symbols declared in this SymbolTable (includes parent SymbolTable).
+     */
+    public get allSymbols(): BscSymbol[] {
```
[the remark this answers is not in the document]
markwpearce: I was just doing the same as `get ownSymbols` ... I'm not against making these functions instead of getters

src/SymbolTable.ts
```diff
@@ -90,30 +101,32 @@ export class SymbolTable {
...
+    getSymbolType(name: string, searchParent = true, context?: TypeContext): BscType {
         const key = name.toLowerCase();
         const symbols = this.symbolMap.get(key);
         if (symbols?.length > 1) {
             //Check if each time it was set, it was set to the same type
             // TODO handle union types
             let sameImpliedType = true;
-            let impliedType = symbols[0].type;
+            const impliedType = getTypeFromContext(symbols[0].type, context);
             for (const symbol of symbols) {
```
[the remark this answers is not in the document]
markwpearce: Yep... I also changed this around so that it only looks at "previous" assignments so that it better knows what the type should be

markwpearce:
Ok... Running this on our production code base and my game engine project only show VALID errors now :) 

Also, I think it runs pretty fast. I don't think performance is an issue.

src/Program.ts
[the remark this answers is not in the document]
TwitchBronBron: Since v1 is still in alpha, I'm fine with landing this as-is for now. We can address performance issues and caching strategies in future PRs (this one is big enough already...)

src/SymbolTable.ts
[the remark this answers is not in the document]
TwitchBronBron: Let's make `get ownSymbols` and `get allSymbols` both functions. While not overly expensive, it's probably more involved than what a getter typically should be doing.

src/SymbolTable.ts
```diff
@@ -90,30 +101,38 @@ export class SymbolTable {
...
+    getSymbolType(name: string, searchParent = true, context?: TypeContext): BscType {
         const key = name.toLowerCase();
         const symbols = this.symbolMap.get(key);
         if (symbols?.length > 1) {
             //Check if each time it was set, it was set to the same type
-            // TODO handle union types
+            // TODO TYPES handle union types
             let sameImpliedType = true;
-            let impliedType = symbols[0].type;
+            let impliedType: BscType;
```
TwitchBronBron, line 114: I think this may be better named `inferredType`.

src/files/BrsFile.Class.spec.ts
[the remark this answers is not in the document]
TwitchBronBron: image
image

You're right, typescript doesn't differentiate. However, if we do eventually add static properties, we would probably do something like:
```
(static property) Animal.kind
```
So I'm fine with this as-is for now. This type of language feature can be easily changed in the future since it's the presentation layer.

src/files/BrsFile.ts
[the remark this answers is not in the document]
TwitchBronBron: This is fine for now, but I agree with @elsassph . Consider this:
```
class Person
    name as string
end class

interface Android
    osVersion as string
end interface

sub main()
    person = new Human()
    droid = {} as Android
end sub
```

We would want to be able to look up `as Android` in the interfaces registry. 

However, classes are the only types we currently support, so we can handle that case in a future PR once we start integrating interfaces into more of the language.

src/files/BrsFile.ts
```diff
@@ -679,17 +702,239 @@ export class BrsFile {
...
+        let namespaceContainer: NamespaceContainer;
+        while (tokenChain[0]) {
+            const namespaceNameToCheck = `${startsWithNamespace}${startsWithNamespace.length > 0 ? '.' : ''}${tokenChain[0].text}`.toLowerCase();
+            const foundNamespace = scope.namespaceLookup[namespaceNameToCheck];
+
+            if (foundNamespace) {
+                namespaceContainer = foundNamespace;
+                namespaceTokens.push(tokenChain[0]);
+                startsWithNamespace = namespaceTokens.map(token => token.text).join('.');
+                tokenChain.shift();
```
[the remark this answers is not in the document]
TwitchBronBron: I agree. Make a copy of `tokenChain` at the top of the function, then you can mutate to your hearts content. :D 

```typescript
tokenChain = [...tokenChain]
```

src/files/BrsFile.ts
```diff
@@ -1369,7 +1631,7 @@ export class BrsFile {
     /**
      * Convert the brightscript/brighterscript source code into valid brightscript
      */
-    public transpile(): CodeWithSourceMap {
+    public transpile(scope?: Scope): CodeWithSourceMap {
```
TwitchBronBron, line 1634: You're not using scope in this function at all, so does the parameter need to be there?

src/parser/Parser.Class.spec.ts
```diff
@@ -27,7 +28,8 @@ describe('parser class', () => {
             TokenKind.Public,
             TokenKind.Protected,
             TokenKind.Private,
-            TokenKind.Override
+            TokenKind.Override,
```
TwitchBronBron, line 31: I'm not sure I agree with preventing these as class properties. Is there a technical reason why we need to prevent them from being properties?

src/parser/Parser.ts
```diff
@@ -2839,6 +2985,19 @@ export class References {
     public namespaceStatements = [] as NamespaceStatement[];
     public newExpressions = [] as NewExpression[];
     public propertyHints = {} as Record<string, string>;
+
+    public getContainingClass(currentToken: Token): ClassStatement {
```
TwitchBronBron, line 2989: These don't really seem to belong on the `References` class. Instead, we should add these as methods to the `Parser` class.

src/parser/Parser.ts
```diff
@@ -2847,6 +3006,17 @@ export interface LocalVarEntry {
     type: BscType;
 }
 
+export interface TokenWithIndex {
+    token: Token;
+    index: number;
```
TwitchBronBron, line 3011: Can you add some comments explaining what `index` is?

src/parser/Parser.ts
```diff
@@ -2847,6 +3006,17 @@ export interface LocalVarEntry {
...
+    isUnknown?: boolean;
```
TwitchBronBron, line 3012: What does this mean? is unknown token? index is unknown? If index is unknown, how about we just set index to undefined instead.

src/parser/Parser.ts
```diff
@@ -2718,6 +2730,138 @@ export class Parser {
...
+            currentTokenIndex = previousTokenResult?.index;
+            includesUnknown = !!previousTokenResult?.isUnknown;
+            while (!includesUnknown && this.isAcceptableChainToken(currentToken, lastTokenHasWhitespace)) {
+                tokenChain.push(currentToken);
+                lastTokenHasWhitespace = currentToken?.leadingWhitespace.length > 0;
+                if (!lastTokenHasWhitespace) {
+                    previousTokenResult = this.getPreviousTokenInChain(currentTokenIndex);
+                    currentToken = previousTokenResult?.token;
+                    currentTokenIndex = previousTokenResult?.index;
+                    includesUnknown = previousTokenResult?.isUnknown;
```
TwitchBronBron, line 2857: It seems like this gets reset on each token. Shouldn't you be doing something like this to flag if the chain has at least one unknown?
```typescript
includesUnknown = includesUnkown || previousTokenResult?.isUnknown;
```

markwpearce: As soon as it includes an unknown token, then the loop exits....

src/parser/Expression.ts
markwpearce: Maybe we could at least check for `m.*` in sub init() ? That's a common pattern (in my experience) to initialize a bunch of values in `init()` ...

PR #412, opened 2021-05-11
Add flag to remove parameter types on transpile

TwitchBronBron:
There are runtime performance implications for using function parameter types. This PR adds a compiler flag to remove param types during transpile. 

**Notable changes:**
 - add `options.transpileOptions.removeParameterTypes` bsc option that, when enabled, removes all function parameter types during transpile. (i.e. `param as object` becomes `param`).

### DIFF
--- a/src/BsConfig.ts
+++ b/src/BsConfig.ts
@@ -91,7 +91,7 @@ export interface BsConfig {
     /**
      * A list of error codes the compiler should NOT emit, even if encountered.
      */
-    ignoreErrorCodes?: (number|string)[];
+    ignoreErrorCodes?: (number | string)[];
 
     /**
      * Emit full paths to files when printing diagnostics to the console. Defaults to false
@@ -107,7 +107,7 @@ export interface BsConfig {
     /**
      * A list of filters used to exclude diagnostics from the output
      */
-    diagnosticFilters?: Array<number | string | { src: string; codes: (number|string)[] } | { src: string } | { codes: (number|string)[] }>;
+    diagnosticFilters?: Array<number | string | { src: string; codes: (number | string)[] } | { src: string } | { codes: (number | string)[] }>;
 
     /**
      * Specify what diagnostic types should be printed to the console. Defaults to 'warn'
@@ -147,4 +147,14 @@ export interface BsConfig {
      * @default true
      */
     sourceMap?: boolean;
+    /**
+     * A set of options that only apply during transpile
+     */
+    transpileOptions?: {
+        /**
+         * Remove the types from function parameters during transpile. This can sometimes provide runtime performance improvements and reduce the risk of runtime crashes
+         * due to type mismatches on-device
+         */
+        removeParameterTypes?: boolean;
+    };
 }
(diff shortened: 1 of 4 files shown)


TwitchBronBron:
One question I have: 
1. do we like the `transpileOptions` sub-config object? Will there be future transpile options? Should the config option just be `options.removeParameterTypesOnTranspile`? Alternatively, should we move the sourcemap config options into `transpileOptions` as part of v1?

TwitchBronBron:
> Only function parameters right? No return value type.

Oh right..good point. I forgot about those. @triwav do return types have the same performance impacts as parameter types?

elsassph:
I mean removing function return types could break store automated validation - they shouldn't be removed.

chrisdp:
> I mean removing function return types could break store automated validation - they shouldn't be removed.

Do you have an example of what you mean or are you talking about how onKeyEvent logs warnings if you don't have the as Boolean type? 

I think this is the only one right now and we could likely white list it.

elsassph:
@chrisdp if you have a function `as void` with a `return` without value, then if `as void` is removed automatically then it will break.

chrisdp:
@elsassph Well this also crash it it's typed as dynamic￼?

Was thinking that rather than actually removing types if all types of subs and functions and returns were converted to dynamic this would probably solve the issue for both subs functions and return that are as voids￼￼.

markwpearce:
TODO: validate return type matches function declaration...

TwitchBronBron:
Wow, I didn't realize there was so much involved in this.....

chrisdp:
@TwitchBronBron i think that's the same for a lot of us haha. Didn't really spot the scope till we started really thinking about the edge cases.

triwav:
@TwitchBronBron late to the party but while in benchmarking there is a tiny difference in performance with types, turning it off in our apps showed no performance improvement so we have left them in

georgejecook:
First, to this question:

> transpileOptions as part of v1 (sourceRoot, sourceMap)?

imo, yes.

I never considered return types on functions when we had this conversation on slack, personally, I'd not be fussed if they were unchanged.

I also think setting them all to dynamic is perfectly reasonable, too.

triwav:
@georgejecook definitely can't set them all to dynamic as that would break something like
```
function a() as Void
return
end function
```
What @elsassph had mentioned above. @chrisdp it will crash they are not the same.

georgejecook:
Great point. I missed that. I'm not bothered about the return values to be honest. The issue I have is that you can't pass invalid into a param as a string, and I don't want default params; but I do want a string which might also be invalid. I think the use case is obvious; but can elaborate further if unclear.  Stripping the param types would solve this. So my vote would be don't touch the return values, which afaict, solves nothing and introduces issues.

elsassph:
Same as George.

georgejecook:
@TwitchBronBron is this going to make it in, any time soon? got a crasher right now, which would be resolved by this.

TwitchBronBron:
After giving this some more thought, we aren't going to include this in core brighterscript. A plugin can easily add a `beforeFileTranspile` hook to change the parameter type to `dynamic` which would result in this same behavior. 

Closing.

PR #415, opened 2021-05-13
concatanates source literal file paths so that the transpiled output does not cause issues with roku's static analysis tool

georgejecook:


chrisdp (approved):

TwitchBronBron (approved):
Looks great. Thanks for remembering to update the docs!

PR #416, opened 2021-05-13
Fix/bs config schema allows string diagnostic codes

markwpearce:
Allows setting the bsconfig to have string types for the codes array of diagnostic filters:

bsLint has diagnostic codes like `"LINT2001"`

e.g.

```
{
  "diagnosticFilters": [
   {
      src: "./someDir/**",
     codes: ["LINT3011"]
   }
 ]
}


TwitchBronBron:
Nice. thanks!

PR #417, opened 2021-05-14
Unifies global callables and function type param definitions

markwpearce:
- Changes `FunctionType` to use the `CallableParam` format used by Callables.
- Removes extra code that basically duplicated construction of FunctionType Params
- Fixes issue with `Val` - there are two global functions with that name, with different param counts (and return types) - https://developer.roku.com/en-ca/docs/references/brightscript/language/global-string-functions.md#vals-as-string-as-float


TwitchBronBron:
Great work! Thanks for this.

PR #418, opened 2021-05-14
ParseJson allows two parameters as of Roku OS 9.4

markwpearce:
https://developer.roku.com/en-ca/docs/references/brightscript/language/global-utility-functions.md#parsejsonjsonstring-as-string-flags---as-string-as-object


TwitchBronBron:
Nice!

PR #421, opened 2021-05-21
Fix code actions construction for "replace" tasks

elsassph:
Also add missing information in the action.


TwitchBronBron:
Looks great. Thanks!

PR #424, opened 2021-05-25
Fix incorrect Block range for inline if/then branch

elsassph:


TwitchBronBron:
Looks great! Can you please add a test to ensure we don't regress on this in the future?

src/lexer/Token.ts
```diff
@@ -10,13 +10,13 @@ export interface Token {
...
     /** True if this token's `text` is a reserved word, otherwise `false`. */
-    isReserved: boolean;
+    isReserved?: boolean;
     /** Where the token was found. */
     range: Range;
     /**
      * Any leading whitespace found prior to this token. Excludes newline characters.
      */
-    leadingWhitespace: string;
+    leadingWhitespace?: string;
```
philippe-elsass-deltatre, line 19: Not sure about making this one optional, but a lot of tests need to be fixed to add this field otherwise...

markwpearce (approved):

PR #426, opened 2021-05-29
Adds language support for Interface statements

TwitchBronBron:
Adds `interface` statements to the language. This PR doesn't do much with them other than allow them to be parsed and included in the AST. Future PRs will probably do more with then.

Is there anything else we need to address in this initial PR?

### DIFF
--- a/src/DiagnosticMessages.ts
+++ b/src/DiagnosticMessages.ts
@@ -175,8 +175,8 @@ export let DiagnosticMessages = {
         code: 1031,
         severity: DiagnosticSeverity.Error
     }),
-    expectedClassKeyword: () => ({
-        message: `Expected 'class' keyword`,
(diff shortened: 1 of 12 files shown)


elsassph:
Are interface functions appearing in the AST as empty functions?

TwitchBronBron:
> Are interface functions appearing in the AST as empty functions?

Yes? Here's the logic in the parser that creates the new `InterfaceMethodStatement`. 

https://github.com/rokucommunity/brighterscript/pull/426/files#diff-0d27a1b5098cc3c9f464ebe55376e540e5ffd0779dd682227eb284abf03e6478R384

markwpearce:
In my “class symbol tables” PR, I added a symbol table to all BscTypes, which is very similar to this idea! 👍
We just need to refactor this PR to match.

PR #435, opened 2021-07-15
Create object validation

TwitchBronBron:
This is the first attempt at validating `CreateObject` calls. I've built a script that scans all of the roku web documentation and builds a JSON file with that information in a useable format. 

I'm opening the PR early so people can see the output and comment as I work through the process.

TODO: 

- [x] link events with the components/elements that emit them '
- [x] scrape the node docs (SceneGraph components)
- [x] implement the CreateObject validation based on the generated docs
- [x] sort the results alphabetically to reduce confusion when diffing in future PRs


markwpearce:
It would be great to get this into `v1` !

TwitchBronBron:
Agreed! I don't think I want to store that data in json long-term, as I'd prefer to auto-generate the actual typescript code so we pay less startup cost. However, it is probably fine to leave as-is in the short term. I'll look into merging this in the next week or so, as time allows.

markwpearce:
- [ ] Capture Node `extends`

markwpearce:
- [ ] Capture deprecation status of nodes/components

TwitchBronBron:
@markwpearce I just found this issue in our Fubo app. 

image


We probably need to add some logic to allow us to freehand some additional items that might not be documented. Something like this:

```
public async run() {
        const outPath = s`${__dirname}/../src/roku-types/data.json`;
        fsExtra.removeSync(outPath);
        loadCache();
        //load the base level roku docs data
        await this.loadReferences();

        //build the various types of docs
        await this.buildComponents();
        await this.buildInterfaces();
        await this.buildEvents();
        await this.buildNodes();

        //certain references are missing hyperlinks. this attempts to repair them
        this.repairReferences();
        this.linkMissingImplementers();

        //NEW PROPOSED CODE:
        this.mergeOverridesOntoResult();

        //store the output
        fsExtra.outputFileSync(outPath, JSON.stringify(this.result, null, 4));
    }
    mergeOverridesOntoResult(){
        deepMerge(this.result, {
             nodes: {
                 "RSGNode": { /* ... */}
             }
        });
```

markwpearce:
@TwitchBronBron Added deprecation status stuff: 

- Scapes docs for deprecation status
- adds it to data.json
- adds a validation for CreateObject(<deprecated>)

TwitchBronBron:
Long-term, I don't want to include the json file in the npm package. However, to get away from that, we need things like fully generated typedef files, method overloading, a way to define platform-included components, user-defined components, etc. That's still a ways out, so might as well land this as-is for now.

Just a note to any plugin authors: do not depend on `roku-types/data.json` or `roku-types/index.ts`. These are only here temporarily until the previously mentioned features are added.

PR #452, opened 2021-09-24
Add regex literal support

TwitchBronBron:
Adds the ability to declare regular expression literals rather than dealing with strings. Example:
```brightscript
regexp = /[a-z]+/i
```
transpiles to
```
regexp = CreateObject("roRegex", "[a-z]+", "i")
```

Tasks: 
- [x] Lexer functionality for capturing regex literals as a single token
- [x] parser functionality
- [x] handle escape characters e.g. `/capture parens \( \)/`
- [x] transpile functionality
- [x] user docs

### DIFF
--- a/src/lexer/Lexer.spec.ts
+++ b/src/lexer/Lexer.spec.ts
@@ -1242,7 +1242,9 @@ describe('lexer', () => {
                 /with spaces/s,
                 /with(parens)and[squarebraces]/,
                 //lots of special characters
-                /.*()^$@/
+                /.*()^$@/,
+                //captures quote char
+                /"/
             );
         });
 
--- a/src/lexer/Lexer.spec.ts
+++ b/src/lexer/Lexer.spec.ts
@@ -1222,4 +1222,26 @@ describe('lexer', () => {
             TokenKind.Eof
         ]);
     });
+
+    describe('regular expression literals', () => {
+        function testRegex(...regexps) {
+            const results = [] as string[];
+            for (const regexp of regexps) {
+                const { tokens } = Lexer.scan(regexp);
+                results.push(tokens[0].text);
+            }
+            expect(results).to.eql(regexps);
+        }
+
+        it('recognizes regex literals', () => {
+            testRegex(
+                '/simple/',
+                '/SimpleWithValidFlags/imsx',
+                '/UnknownFlags/VUI',
+                '/with spaces/andflags',
+                '/with(parens)and[squarebraces]/',
+                '/*()^$@/'
+            );
+        });
+    });
 });
--- a/src/lexer/Lexer.spec.ts
+++ b/src/lexer/Lexer.spec.ts
@@ -1224,10 +1224,11 @@ describe('lexer', () => {
     });
 
     describe('regular expression literals', () => {
-        function testRegex(...regexps) {
+        function testRegex(...regexps: Array<string | RegExp>) {
+            regexps = regexps.map(x => x.toString());
             const results = [] as string[];
             for (const regexp of regexps) {
-                const { tokens } = Lexer.scan(regexp);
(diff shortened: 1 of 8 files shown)


src/lexer/Lexer.spec.ts
```diff
@@ -1222,4 +1222,40 @@ describe('lexer', () => {
...
+                results.push(tokens[0].text);
+            }
+            expect(results).to.eql(regexps);
+        }
+
+        it('recognizes regex literals', () => {
+            testRegex(
+                /simple/,
+                /SimpleWithValidFlags/g,
+                /UnknownFlags/gi,
```
philippe-elsass-deltatre, line 1241: One case should probably have a double quote to verify it's correctly escaped

src/lexer/Lexer.ts
```diff
@@ -927,6 +943,45 @@ export class Lexer {
...
+        let nextCharNeedsEscaped = false;
+
+        //finite loop to prevent infinite loop if something went wrong
+        for (let i = this.current; i < this.source.length; i++) {
+
+            //if we reached the end of the regex, consume any flags
+            if (this.check('/') && !nextCharNeedsEscaped) {
+                this.advance();
+                //consume all flag-like chars (let the parser validate the actual values)
+                while (/[a-z]/i.exec(this.peek())) {
```
philippe-elsass-deltatre, line 963: Wouldn't a single peek return all the characters following `/` as one literal?

src/lexer/Lexer.spec.ts
TwitchBronBron: OH! I knew I was forgetting something. I just pushed a commit  that verifies the lexer handles this fine. However, the parser does _not_ properly escape the quotemark. Nice catch! I'll work on that now.

TwitchBronBron: Fixed.

chrisdp:
Question. Should we support running the different functions inline?

https://developer.roku.com/en-ca/docs/references/brightscript/interfaces/ifregex.md#

So for example:
```
result = /[a-z]+/i.match(myString)

```

This could always be something we maybe support later but it just popped into my head.

src/parser/tests/expression/RegexLiteralExpression.spec.ts
```diff
@@ -0,0 +1,67 @@
...
+                sub main()
+                    print CreateObject("roRegex", "hello", "")
+                end sub
+            `);
+        });
+
+        it('handles weird escapes', () => {
+            testTranspile(`
+                sub main()
+                    print /\\r\\n\\//
```
chrisdp, line 45: Are we able to do this without the escaping as if it was in a string already?

So something like `/\r\n\/`? Not sure if that's quite right. The "\\//" part is messing me up a bit.

TwitchBronBron:
> Question. Should we support running the different functions inline?
> 
> https://developer.roku.com/en-ca/docs/references/brightscript/interfaces/ifregex.md#
> 
> So for example:
> 
> ```
> result = /[a-z]+/i.match(myString)
> ```
> 
> This could always be something we maybe support later but it just popped into my head.

This SHOULD work already since it's an expression. If you can do `CreateObject("roRegex", "whatever").match(my string)` then this new regex literal will also support that.

src/parser/tests/expression/RegexLiteralExpression.spec.ts
TwitchBronBron: The escaping is happening in typescript land. It IS a little harder to read in the test this way, I agree, but I'm not sure how else to get around that.

chrisdp: Ohhhhh right... the test is just a string. I'm dum haha.

chrisdp (approved):

PR #456, opened 2021-10-04
Upgrade typescript and eslint

TwitchBronBron:
Upgrade to typescript@4.4.3, eslint@7.32.0, @typescript-eslint/eslint-plugin@4.32.0 and a few lint issues introduced by those upgrades.


.eslintrc.js
```diff
@@ -70,6 +80,7 @@ module.exports = {
         '@typescript-eslint/require-array-sort-compare': 'off',
         '@typescript-eslint/restrict-plus-operands': 'off',
         '@typescript-eslint/restrict-template-expressions': 'off',
+        '@typescript-eslint/sort-type-union-intersection-members': 'off',
```
TwitchBronBron, line 83: Disabled this as most of the time it's enforcement simply for enforcement's sake. Don't see much real-world value in this.

.eslintrc.js
```diff
@@ -40,6 +41,15 @@ module.exports = {
         '@typescript-eslint/no-invalid-this': 'off',
         '@typescript-eslint/no-magic-numbers': 'off',
         '@typescript-eslint/no-parameter-properties': 'off',
+        //had to add this rule to prevent eslint from crashing
+        '@typescript-eslint/no-restricted-imports': ['off', {}],
+        //mitigating this sometimes results in undesirably verbose code. Should investigate enabling again in the future.
+        '@typescript-eslint/no-unsafe-argument': 'off',
```
TwitchBronBron, line 47: In theory, I like preventing `any` from being in the code. But there are a lot of changes required to enable this rule, and overall, most of the changes were a little too verbose for certain situations that don't really need it. So disabling the rule for now, but we can revisit in the future if if we want to tighten up the `any` rules.

chrisdp: Make sense to open a GH issue with a good first issue label or help wanted label?

src/lexer/Lexer.ts
```diff
@@ -83,7 +83,7 @@ export class Lexer {
      * @param options options used to customize the scan process
      * @returns an object containing an array of `errors` and an array of `tokens` to be passed to a parser.
      */
-    public scan(toScan: string, options?: ScanOptions): Lexer {
+    public scan(toScan: string, options?: ScanOptions): this {
```
chrisdp, line 86: Oh, interesting. I didn't know this was a thing.

chrisdp (approved):

src/lexer/Lexer.ts
TwitchBronBron: I didn't either, until the linter complained about it. I like that because it helps show which functions can be chained.

.eslintrc.js
TwitchBronBron: Sure! #457

PR #458, opened 2021-10-04
Fix bugs when parsing regex

### LINKED ISSUE #455: Regex literal syntax bug
There's a critical bug in the regular expression literal feature introduced by #94. The lexer parses all text starting with slash and ending with slash (and optionally including the flags). This will fail for the following code:
```
marginLeft = m.top.width/2 - m.top.loadingLabel.boundingRect().width/2
```

TwitchBronBron:
Resolves #455

### DIFF
--- a/src/lexer/Lexer.ts
+++ b/src/lexer/Lexer.ts
@@ -5,6 +5,7 @@ import { isAlpha, isDecimalDigit, isAlphaNumeric, isHexDigit } from './Character
 import type { Range, Diagnostic } from 'vscode-languageserver';
 import { DiagnosticMessages } from '../DiagnosticMessages';
 import util from '../util';
+import { PreceedingRegexTypes } from '.';
 
 export class Lexer {
     /**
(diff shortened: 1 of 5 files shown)


src/lexer/TokenKind.ts
```diff
@@ -598,3 +598,47 @@ export const DeclarableTypes = [
     TokenKind.Void,
     TokenKind.Function
 ];
+
+/**
+ * The tokens that might preceed a regex literal
+ */
+export const PreceedingRegexTypes = new Set([
```
TwitchBronBron, line 605: This is the full list of tokens that can come before a regex literal. If it's not in this list, then a regex literal will not be tokenized even if we encounter a forward slash. I find it safer to create an allowlist rather than a denylist, as it's easier to test, and we won't accidentally break functionality when we forgot to put something in the denylist, whereas only new adopters of regexp literals would encounter issues when something is missing from the allowlist.

src/lexer/Lexer.ts
```diff
@@ -953,29 +966,36 @@ export class Lexer {
...
+        //regexps can only occur when preceeded by exactly one of these tokens:
+        const previousKind = this.getPreviousNonWhitespaceToken()?.kind;
 
-            //if we reached the end of the regex, consume any flags
-            if (this.check('/') && !nextCharNeedsEscaped) {
-                this.advance();
-                //consume all flag-like chars (let the parser validate the actual values)
-                while (/[a-z]/i.exec(this.peek())) {
+        //preceeded by an allowed token, or if there are no previous tokens (i.e. this is the first token in the file).
+        if (PreceedingRegexTypes.has(previousKind) || !previousKind) {
```
TwitchBronBron, line 973: This is the core of the change. Check for only an approved token preceeding the opening slash.

markwpearce:
I believe the RegexLiteral check is being done in Brightscript (.BRS) mode? If it is, I don't think it should be

TwitchBronBron:
Oops! You're right. Good catch.

TwitchBronBron:
@markwpearce I addressed your concern in 57cca5d010bb680b08b74d4fe3fabd9be97be12c

markwpearce (approved):

PR #478, opened 2022-01-07
AST editing

TwitchBronBron:
Adds ability to edit AST during `beforeFileTranspile`, and then the changes are reverted after the `afterFileTranspile` event.
 - Adds new `AstEditor` class that applies edits and tracks how to undo them.
 - Adds `editor` property to the `beforeFileTranspile` and `afterFileTranspile` lifecycle events.
 - Once the `afterFileTranspile` event has fired for all plugins, the `editor.undoAll()` method is called which will revert all of the AST changes made during the transpile flow.
 - Update docs with example
 - any files that were edited using `editor` will have the `needsTranspiled` property set to `true` automatically.

### DIFF
--- a/docs/plugins.md
+++ b/docs/plugins.md
@@ -228,32 +228,44 @@ export default function () {
 };
 ```
 
-### Example AST modifier plugin
+## Modifying code
+Sometimes plugins will want to modify code before the project is transpiled. While you can technically edit the AST directly at any point in the file's lifecycle, this is not recommended as those changes will remain changed as long as that file exists in memory and could cause issues with file validation if the plugin is used in a language-server context (i.e. inside vscode).
 
-AST modification can be done after parsing (`afterFileParsed`), but it is recommended to modify the AST only before transpilation (`beforeFileTranspile`), otherwise it could cause problems if the plugin is used in a language-server context.
+Instead, we provide an instace of an `AstEditor` class in the `beforeFileTranspile` event that allows you to modify AST before the file is transpiled, and then those modifications are undone `afterFileTranspile`.
+
+For example, consider the following brightscript code:
+```brightscript
+sub main()
+    print "hello <FIRST_NAME>"
+end sub
+```
+
+Here's the plugin:
 
 ```typescript
-// removePrint.ts
-import { CompilerPlugin, Program, TranspileObj } from 'brighterscript';
-import { EmptyStatement } from 'brighterscript/dist/parser';
-import { isBrsFile, createStatementEditor, editStatements } from 'brighterscript/dist/parser/ASTUtils';
+import { CompilerPlugin, BeforeFileTranspileEvent, isBrsFile, WalkMode, createVisitor, TokenKind } from './';
 
 // plugin factory
 export default function () {
     return {
         name: 'removePrint',
         // transform AST before transpilation
-        beforeFileTranspile: (entry: TranspileObj) => {
-            if (isBrsFile(entry.file)) {
-                // visit functions bodies and replace `PrintStatement` nodes with `EmptyStatement`
-                entry.file.parser.functionExpressions.forEach((fun) => {
-                    const visitor = createStatementEditor({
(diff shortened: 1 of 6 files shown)


docs/plugins.md
```diff
@@ -228,32 +228,44 @@ export default function () {
...
-                    const visitor = createStatementEditor({
-                        PrintStatement: (statement) => new EmptyStatement()
-                    });
-                    editStatements(fun.body, visitor);
+        beforeFileTranspile: (event: BeforeFileTranspileEvent) => {
+            if (isBrsFile(event.file)) {
+                event.file.ast.walk(createVisitor({
+                    LiteralExpression: (literal) => {
+                        //replace every occurance of <FIRST_NAME> in strings with "world"
+                        if (literal.token.kind === TokenKind.StringLiteral && literal.token.text.includes('<FIRST_NAME>')) {
```
philippe-elsass-deltatre, line 258: Isn't the string tokenized as `Hello <FIRST_NAME>`?

src/Program.spec.ts
```diff
@@ -1640,6 +1641,45 @@ describe('Program', () => {
...
+            let literalExpression: LiteralExpression;
+            //replace all strings with "goodbye world"
+            program.plugins.add({
+                name: 'TestPlugin',
+                beforeFileTranspile: (event) => {
+                    if (isBrsFile(event.file)) {
+                        event.file.ast.walk(createVisitor({
+                            LiteralExpression: (literal) => {
+                                literalExpression = literal;
+                                event.astEditor.setProperty(literal.token, 'text', '"goodbye world"');
```
philippe-elsass-deltatre, line 1660: What about just `editor`?

src/interfaces.ts
```diff
@@ -204,8 +205,8 @@ export interface CompilerPlugin {
     beforeFileParse?: (source: SourceObj) => void;
     afterFileParse?: (file: BscFile) => void;
     afterFileValidate?: (file: BscFile) => void;
-    beforeFileTranspile?: (entry: TranspileObj) => void;
-    afterFileTranspile?: (entry: TranspileObj) => void;
+    beforeFileTranspile?: PluginHandler<BeforeFileTranspileEvent>;
```
philippe-elsass-deltatre, line 208: Breaking public API

docs/plugins.md
philippe-elsass-deltatre: Sorry I didn't notice you're testing "includes"

src/interfaces.ts
TwitchBronBron: Unless I'm missing something, it should still be the same public API. 

Here's the definition of `PluginHandler`. 
```
export type PluginHandler<T> = (event: T) => void;
```

Also, the `BeforeTranspileEvent` interface includes both properties from `TranspileObj`, so that also is still API compatible since _we_ produce the event for the subscribers (i.e. users don't need to create the events). 

```typescript
export interface BeforeFileTranspileEvent {
    file: BscFile;
    outputPath: string;
    astEditor: AstEditor;
}

export interface TranspileObj {
    file: BscFile;
    outputPath: string;
}
```

philippe-elsass-deltatre: Ah ok

src/astUtils/AstEditor.ts
```diff
@@ -0,0 +1,107 @@
...
+    public removeFromArray<T extends any[]>(array: T, index: number) {
+        const change = new RemoveFromArrayChange(array, index);
+        this.changes.push(change);
+        change.apply();
+    }
+
+    /**
+     * Unto all changes.
+     */
+    public undoAll() {
```
elsassph, line 41: `undoAll` shouldn't be in the API exposed to plugins

TwitchBronBron: Fair point. I updated the interface for the transpile events: 

https://github.com/rokucommunity/brighterscript/blob/c617cc98302c968c32ac6369c67c28a882d8547e/src/interfaces.ts#L252

PR #485, opened 2022-01-15
Feat/enum completion

georgejecook:
Adds completions for enums and enum statements. Works as follows:

 - if current token is an enum member, then only members for that enum are returned
 - includes global scope enum types in results, without affecting completions
 - includes namespace enum types; non exclusively (i.e. will include the classes and functions as well)


src/Scope.ts
```diff
@@ -370,6 +393,8 @@ export class Scope {
                         ns.classStatements[statement.name.text.toLowerCase()] = statement;
                     } else if (isFunctionStatement(statement) && statement.name) {
                         ns.functionStatements[statement.name.text.toLowerCase()] = statement;
+                    } else if (isFunctionStatement(statement) && statement.name) {
```
TwitchBronBron, line 396: right now this is a duplicate of the two lines above. Is this supposed to be enum-related?

```ts
 } else if (isEnumStatement(statement) && statement.name) {
    ns.enumStatements[statement.name.toLowerCase()] = statement;
}
```

TwitchBronBron (changes requested):
I pushed some commits to clean up formatting and a few other small issues. I have one question, but other than that, looks great so far!

src/Scope.ts
```diff
@@ -359,7 +381,8 @@ export class Scope {
                             namespaces: new Map<string, NamespaceContainer>(),
                             classStatements: {},
                             functionStatements: {},
-                            statements: []
+                            statements: [],
+                            enumStatements: {}
```
TwitchBronBron, line 385: Let's use a `Map<string, EnumStatement>` here like we do for namespaces. (I intend on converting the class and function statements to using maps in the future)

src/Scope.ts
```diff
@@ -1010,6 +1035,7 @@ interface NamespaceContainer {
     statements: Statement[];
     classStatements: Record<string, ClassStatement>;
     functionStatements: Record<string, FunctionStatement>;
+    enumStatements: Record<string, EnumStatement>;
```
TwitchBronBron, line 1038: Let's use a Map<string, EnumStatement> here like we do for namespaces. (I intend on converting the class and function statements to using maps in the future

src/Scope.ts
georgejecook: oops - must have merged bombed this.

src/Scope.ts
georgejecook: done

georgejecook:
somethings gone a bit weird here, @TwitchBronBron  - I accidentally did the validation work onthis branch, so I dropped that commit and force pushed.
I then noticed all my works was missing, so I merged it back in.
Now I find that the references for the primary dotted gets is missing.
not sure what's going on.. 

hayyylp :)

TwitchBronBron:
I reverted that code because it wasn't helpful and didn't solve what we needed. Stick with the walker for now...I'll come up with a more efficient approach later.

georgejecook:
@TwitchBronBron updated to use the walker; but it no longer runs. I think it might be a circular reference, issue; but this ts error is beyond me.

```

> brighterscript@0.42.0 test
> nyc mocha


TypeError: Class extends value undefined is not a constructor or null
    at Object.<anonymous> (/home/george/hope/open-source/vsc/brighterscript/src/XmlScope.ts:2:601)
    at Module._compile (node:internal/modules/cjs/loader:1092:14)
    at Module.replacementCompile (/home/george/hope/open-source/vsc/brighterscript/node_modules/append-transform/index.js:60:13)
    at Module.m._compile (/home/george/hope/open-source/vsc/brighterscript/node_modules/ts-node/src/index.ts:839:23)
    at Module._extensions..js (node:internal/modules/cjs/loader:1121:10)
    at require.extensions.<computed> (/home/george/hope/open-source/vsc/brighterscript/node_modules/ts-node/src/index.ts:842:12)
    at Object.<anonymous> (/home/george/hope/open-source/vsc/brighterscript/node_modules/append-transform/index.js:64:4)
    at Module.load (node:internal/modules/cjs/loader:972:32)
    at Function.Module._load (node:internal/modules/cjs/loader:813:14)
```

TwitchBronBron:
Check all your newly added import statements in .ts files and make sure none of them are coming from ".." or "../../".  (Without a file name). Those regularly cause circular ref issues. Always import from an exact files.

TwitchBronBron (approved):

nievesmiguel:
feat/enum-completion

PR #486, opened 2022-01-16
Feat/enum validation

georgejecook:


georgejecook:
something's gone a bit odd here, @TwitchBronBron  - I had acceidentally done the validation work on this branch, so I dropped that commit; but then I found loads of my work was missing, so I merged it back in.
Now I find that primaryDottedGets is missing from references; not sure why.

PR #487, opened 2022-01-17
Parser references expressions

TwitchBronBron:
Adds support for tracking distinct expressions during parse. This will eliminate the need for AST walking for many validators. 

- [ ] track them during initial parse
- [ ] re-calculate them when `invalidateReferences()` is called (partially done)

### DIFF
--- a/src/parser/Parser.ts
+++ b/src/parser/Parser.ts
@@ -90,7 +90,7 @@ import {
 } from './Expression';
 import type { Diagnostic, Range } from 'vscode-languageserver';
 import { Logger } from '../Logger';
-import { isAnnotationExpression, isCallExpression, isCallfuncExpression, isClassMethodStatement, isCommentStatement, isDottedGetExpression, isIfStatement, isIndexedGetExpression, isVariableExpression } from '../astUtils/reflection';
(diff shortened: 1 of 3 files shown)


src/parser/Parser.ts
```diff
@@ -2868,6 +2959,18 @@ export class References {
...
+     * A collection of full expressions. This excludes intermediary expressions.
+     *
+     * Example 1:
+     * `a.b.c` is composed of `a` (variableExpression)  `.b` (DottedGetExpression) `.c` (DottedGetExpression)
+     * This will only contain the final `.c` DottedGetExpression because `.b` and `a` can both be derived by walking back from the `.c` DottedGetExpression.
+     *
+     * Example 2:
+     * `name.space.doSomething(a.b.c)` will result in 2 entries in this list. the `CallExpression` for `doSomething`, and the `.c` DottedGetExpression.
+     */
+    public expressions = new Set<Expression>();
```
TwitchBronBron, line 2972: This is the main point of the change. It contains a list of all full expressions found in the parsed file. Can be very useful for enum validations, other property validations, etc.

philippe-elsass-deltatre: Why a set and not an array?

elsassph: Why not an array?

elsassph (approved):

src/parser/Parser.ts
TwitchBronBron: I'm using the `Set` to prevent duplicates, and also I remove elements occasionally.

PR #492, opened 2022-01-25
Remove nested index files

TwitchBronBron:
Having export [barrels](https://github.com/basarat/typescript-book/blob/master/docs/tips/barrel.md) regularly causes debugging issues within this project, because ts-node doesn't handle circular dependencies well, and sometimes neither does typescript. 

 - removes all barrels except for the primary `src/index.ts` file which is used by external consumers. 
 - change all internal imports from barrels into explicit file imports


src/index.ts
```diff
@@ -8,8 +8,18 @@ export { Watcher } from './Watcher';
 export * from './interfaces';
 export * from './LanguageServer';
 export * from './XmlScope';
-export * from './lexer';
```
TwitchBronBron, line 11: All the imports from these removed lines were expanded out to their individual file exports. This ensures that our "public" api hasn't changed. One caveat, if plugin authors doing things like `import { createVisitor } from 'brighterscript/astUtils' then this is technically a breaking change. However, since we were already exporting those directly from `src/index.ts`, I'd argue they were not using our library properly. 

Thoughts?

chrisdp (approved):

livecano (approved):

PR #495, opened 2022-01-26
Adds support for function docs in hover

TwitchBronBron:
This shows the function documentation comments when hovering over a function. 

Here's an example:
image

All comments will be parsed as markdown.

### DIFF
--- a/src/LanguageServer.ts
+++ b/src/LanguageServer.ts
@@ -893,16 +893,6 @@ export class LanguageServer {
 
         //return the first non-falsey hover. TODO is there a way to handle multiple hover results?
         let hover = hovers.filter((x) => !!x)[0];
-
-        //TODO improve this to support more than just .brs files
-        if (hover?.contents) {
-            //create fenced code block to get colorization
-            hover.contents = {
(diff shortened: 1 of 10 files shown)


src/LanguageServer.ts
```diff
@@ -893,16 +893,6 @@ export class LanguageServer {
 
         //return the first non-falsey hover. TODO is there a way to handle multiple hover results?
         let hover = hovers.filter((x) => !!x)[0];
-
-        //TODO improve this to support more than just .brs files
```
TwitchBronBron, line 897: This was forcing all hovers to be brightscript...but we need to support any documentation information provided, so force the underlying files to provide the full markdown text instead of casting it all here.

livecano:
@TwitchBronBron, does the documentation has to be in a specific format? do you have a code snippet with a sample?

TwitchBronBron:
@livecano as of right now, it just reads every comment statement between functions. For example:

```vb
'I'm shown for one()
sub one()
end sub
' I'm shown for two()
' I'm shown also for two()

'I'm shown for two() as well
sub two()
end sub

livecano (commented):
Looks good!

PR #499, opened 2022-02-02
Array types

markwpearce:
Adds fully parsed array type syntax

Allows for valid Brighterscript like:

```
function getSum(numbers as float[])
  sum = 0.0
  for each num in numbers
    sum += num
  end for
  return sum
end function

function getTwoPis() as float[]
   return [3.14, 3.14]
end function

sub main()
  twoPi = getTwoPis()
  myNumbers = [1.2, 4.5, twoPi[0] ]
  print getSum(myNumbers)
end sub

```

### DIFF
--- a/src/Scope.ts
+++ b/src/Scope.ts
@@ -7,7 +7,7 @@ import type { CallableContainer, BsDiagnostic, FileReference, BscFile, CallableC
 import type { FileLink, Program } from './Program';
 import { BsClassValidator } from './validators/ClassValidator';
 import type { NamespaceStatement, Statement, FunctionStatement, ClassStatement } from './parser/Statement';
-import type { NewExpression } from './parser/Expression';
+import type { FunctionExpression, NewExpression } from './parser/Expression';
 import { ParseMode } from './parser/Parser';
 import { standardizePath as s, util } from './util';
 import type { MinMax } from './util';
@@ -23,6 +23,7 @@ import type { CustomType } from './types/CustomType';
 import { UninitializedType } from './types/UninitializedType';
 import { ObjectType } from './types/ObjectType';
 import { getTypeFromContext } from './types/BscType';
+import type { BscType, TypeContext } from './types/BscType';
 import { DynamicType } from './types/DynamicType';
 import type { Token } from './lexer/Token';
 
@@ -147,9 +148,18 @@ export class Scope {
         });
     }
 
-    public getAncestorTypeList(className: string, namespaceName?: string): CustomType[] {
+    public getAncestorTypeListByContext(thisType: BscType, context?: TypeContext): CustomType[] {
+        const funcExpr = context?.file?.getFunctionExpressionAtPosition(context?.position);
+        if (isCustomType(thisType)) {
+            return this.getAncestorTypeList(thisType.name, funcExpr);
+        }
+        return [];
+    }
+
+    public getAncestorTypeList(className: string, functionExpression?: FunctionExpression): CustomType[] {
+        const lowerNamespaceName = functionExpression.namespaceName?.getName().toLowerCase();
         const ancestors: CustomType[] = [];
-        let currentClass = this.getClassFileLink(className, namespaceName)?.item;
+        let currentClass = this.getClassFileLink(className, lowerNamespaceName)?.item;
         if (currentClass) {
             ancestors.push(currentClass?.getCustomType());
         }
@@ -775,19 +785,13 @@ export class Scope {
                             break;
                         }
                         let argType = arg.type ?? new UninitializedType();
-                        let assignable = false;
                         const paramType = getTypeFromContext(param.type, paramTypeContext);
                         if (!paramType) {
                             // other error - can not determine what type this parameter should be
                             continue;
                         }
                         argType = getTypeFromContext(argType, argTypeContext);
-                        if (isCustomType(argType)) {
-                            const lowerNamespaceName = expCall.functionExpression.namespaceName?.getName().toLowerCase();
-                            assignable = argType.isAssignableTo(paramType, argTypeContext, this.getAncestorTypeList(argType.name, lowerNamespaceName));
-                        } else {
-                            assignable = argType?.isAssignableTo(paramType, argTypeContext);
-                        }
+                        let assignable = argType?.isAssignableTo(paramType, argTypeContext);
                         if (!assignable) {
                             // TODO TYPES: perhaps this should be a strict mode setting?
                             assignable = argType?.isConvertibleTo(paramType, argTypeContext);
--- a/src/Scope.ts
+++ b/src/Scope.ts
@@ -341,6 +341,15 @@ export class Scope {
     }
 
     /**
+     * Get the global callable with the specified name.
+     * If there are overridden callables with the same name, the closest callable to this scope is returned
+     * @param name
+     */
+    public getGlobalCallableByName(name: string) {
+        return globalCallableMap.get(name.toLowerCase());
+    }
+
+    /**
      * Iterate over Brs files not shadowed by typedefs
      */
     public enumerateBrsFiles(callback: (file: BrsFile) => void) {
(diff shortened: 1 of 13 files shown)


src/Scope.ts
```diff
@@ -340,6 +340,15 @@ export class Scope {
         }
     }
 
+    /**
+     * Get the global callable with the specified name.
+     * If there are overridden callables with the same name, the closest callable to this scope is returned
+     * @param name
+     */
+    public getGlobalCallableByName(name: string) {
```
markwpearce, line 348: Added helper function here to fix circular reference in BrsFile.spec.ts

src/files/BrsFile.spec.ts
```diff
@@ -2837,4 +2882,225 @@ describe('BrsFile', () => {
             testPluginTranspile();
         });
     });
+
+    describe('getSymbolTypeFromToken', () => {
```
markwpearce, line 2886: Added a bunch more tests for this common type checking function

src/files/BrsFile.ts
```diff
@@ -901,11 +901,15 @@ export class BrsFile {
                 // the next symbol to check will be the return value of this function
                 symbolType = getTypeFromContext(symbolType.returnType, typeContext);
                 if (tokenFoundCount < tokenChain.length) {
-                    // We're still
+                    // We still have more tokens, but remember the last known reference
                     symbolTypeBeforeReference = symbolType;
                 }
             }
 
+            if (isArrayType(symbolType) && tokenUsage === TokenUsage.ArrayReference) {
```
markwpearce, line 909: If this token was using an array reference (e.g `token[i]`)  then the actual type it represents is the type of the elements in the array

src/parser/Parser.ts
```diff
@@ -2446,6 +2450,18 @@ export class Parser {
             // just get whatever's next
             typeToken = this.advance();
         }
+
+        if (typeToken && this.options.mode === ParseMode.BrighterScript) {
```
markwpearce, line 2454: identify types like `float[]` or `customType[]`

src/types/ArrayType.ts
```diff
@@ -33,19 +38,23 @@ export class ArrayType implements BscType {
...
-    public isConvertibleTo(targetType: BscType) {
-        return this.isAssignableTo(targetType);
+    public isConvertibleTo(targetType: BscType, context?: TypeContext) {
+        return this.isAssignableTo(targetType, context);
     }
 
-    public toString() {
-        return `Array<${this.innerTypes.map((x) => x.toString()).join(' | ')}>`;
+    public toString(context?: TypeContext) {
+        return `Array<${this.innerTypes.map((x) => x.toString(context)).join(' | ')}>`;
```
markwpearce, line 46: I think this should be changed to `\`${this.defaultType.toString(context)}[]\``  ... thoughts?

src/files/BrsFile.spec.ts
```diff
@@ -2837,4 +2882,225 @@ describe('BrsFile', () => {
...
+                end sub
+
+                class MyKlass
+                    name as string
+                    age as integer
+                end class
+            `);
+            const funcExpr = file.parser.references.functionExpressions[0];
+            const lookups = [
+                { line: 3, col: 35, name: 'MyKlass.name', type: new StringType() },
```
TwitchBronBron, line 2970: Since brighterscript is case-insensitive (and we're building our classes case insensitive), we should make sure these types of lookups work case-insensitive too. Can you add some tests to verify? Perhaps something like:
```
klass = new MyKlass()
print KLASS.NAME
print klass.AGE
```

src/files/BrsFile.ts
```diff
@@ -933,9 +937,10 @@ export class BrsFile {
                 tokenText.push(token.text);
                 break;
             }
-            if (tokenText.length > 2) {
-                tokenText.shift(); // only care about last two symbols
-            }
+
+        }
+        if (tokenText.length > 2) {
+            tokenText = tokenText.slice(-2); // only care about last two symbols
```
TwitchBronBron, line 943: Could you add a more clear comment as to why we're removing characters here?

src/parser/Parser.spec.ts
```diff
@@ -1349,7 +1365,51 @@ describe('parser', () => {
                 `, ParseMode.BrighterScript);
                 expectZeroDiagnostics(parser.diagnostics);
                 const currentSymbolTable = parser.references.functionExpressions[0].symbolTable;
-                expect(currentSymbolTable.getSymbolType('datum').toString()).to.eq('dynamic');
+                expect(isDynamicType(currentSymbolTable.getSymbolType('datum'))).to.be.true;
+            });
+
+            it('determines the type of the variable in a for each if the target is an array literal', () => {
+                const parser = parse(`
+                sub doLoop()
```
TwitchBronBron, line 1373: indent one more tab for this text block

src/parser/Parser.spec.ts
```diff
@@ -1349,7 +1365,51 @@ describe('parser', () => {
...
+                end sub
+                `, ParseMode.BrighterScript);
+                expectZeroDiagnostics(parser.diagnostics);
+                const currentSymbolTable = parser.references.functionExpressions[0].symbolTable;
+                expect(isIntegerType(currentSymbolTable.getSymbolType('datum'))).to.be.true;
+            });
+
+            it('determines the type of the variable in a for each if the target is an array', () => {
+                const parser = parse(`
+                sub doLoop(someData as integer[])
```
TwitchBronBron, line 1387: indent one more tab for this text block

src/parser/Parser.spec.ts
```diff
@@ -1349,7 +1365,51 @@ describe('parser', () => {
...
+                end sub
+                `, ParseMode.BrighterScript);
+                expectZeroDiagnostics(parser.diagnostics);
+                const currentSymbolTable = parser.references.functionExpressions[0].symbolTable;
+                expect(isIntegerType(currentSymbolTable.getSymbolType('datum'))).to.be.true;
+            });
+
+            it('determines the type of the variable in a for each if the target is an array of some custom type', () => {
+                const parser = parse(`
+                sub doLoop(someData as MyKlass[])
```
TwitchBronBron, line 1400: indent one more tab for this text block

src/parser/Parser.ts
```diff
@@ -2446,6 +2450,18 @@ export class Parser {
...
+            // Check if it is an array - that is, if it has `[]` after the type
+            // eg. `string[]` or `SomeKlass[]`
+            if (this.check(TokenKind.LeftSquareBracket)) {
+                this.advance();
+                if (this.check(TokenKind.RightSquareBracket)) {
+                    const rightBracket = this.advance();
+                    typeToken = createToken(TokenKind.Identifier, typeToken.text + '[]', util.getRange(typeToken, rightBracket));
```
TwitchBronBron, line 2461: I think this would work much better long-term if we added a `TypeExpression` instead, which would support all sorts of this type of pattern. 

We don't need to do that in this PR though, we can tweak it in the future.

src/types/ArrayType.ts
TwitchBronBron: I had originally imagined this to support arrays with multiple types. For example, something like:
`Array<string|integer>`. I'm fine with your suggestion because I don't know that we're ready for my design quite yet. But would `this.defaultType` already support that concept?

src/types/FunctionType.ts
```diff
@@ -38,20 +38,20 @@ export class FunctionType implements BscType {
         return this;
     }
 
-    public isAssignableTo(targetType: BscType) {
+    public isAssignableTo(targetType: BscType, context?: TypeContext) {
```
TwitchBronBron, line 41: Will our new `TypeExpression` concept negate the need for this new `TypeContext` param?

TwitchBronBron (approved):
I had a few small comments, but nothing critical. This is the v1 branch, so I'm happy to iterate on these designs. So let me know if you'd like to merge this now and work on `TypeExpression` in a separate PR, or do it all together.

src/types/FunctionType.ts
markwpearce: No ... TypeContext is used to find out "What context is this type referenced from?" .. for example, a file included in `ComponentA` might have a class defined a certain way, but that same file included in a different component might have a different definition for the same class name, just based on other files included in that scope.

PR #502, opened 2022-02-02
Fire plugin events for getTranspiledFile. Add new event properties

TwitchBronBron:
- The "preview transpiled code" feature in vscode utilizes the `getTranspiledFileContents` function. However, plugin-related changes were not showing. This resolves that issue.
 - adds `code`, `map`, `typedef` properties to the `afterFileTranspile` event to allow plugins to manually tweak the output code (at their own risk). 
 - mostly unrelated: return a reference to the plugin in the `add`, `addFirst`, and `remove` methods of the `PluginInterface` class.

### DIFF
--- a/src/Program.ts
+++ b/src/Program.ts
@@ -26,7 +26,7 @@ import { ParseMode } from './parser/Parser';
 import { TokenKind } from './lexer/TokenKind';
 import { BscPlugin } from './bscPlugin/BscPlugin';
 import { AstEditor } from './astUtils/AstEditor';
-import { SourceMapGenerator } from 'source-map';
+import type { SourceMapGenerator } from 'source-map';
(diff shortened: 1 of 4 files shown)


src/Program.ts
```diff
@@ -1194,12 +1195,59 @@ export class Program {
...
+            this.getFile(pathAbsolute)
+        );
+    }
+
+
+    /**
+     * Internal function used to transpile files.
+     * This does not write anything to the file system
+     */
+    private _getTranspiledFileContents(file: BscFile, outputPath?: string): FileTranspileResult {
```
TwitchBronBron, line 1208: Perhaps I should come up with a better name for this? I don't like the underscore. I don't want to use the word "internal" either though.

chrisdp (approved):

PR #504, opened 2022-02-03
Move file validation into  program.validate()

TwitchBronBron:
moves the file validation logic out of `addReplaceFile` and into `program.validate()`. This accomplishes several things:
 - aligns with the flow in v1 which makes more logical sense
 - improve performance slightly when typing in an editor because file validations get deferred until the validate phase
 - prevent file validation events on files shadowed by typedef files, which was a bug introduced in #494

This is technically a breaking change to the event flow. 

Event flow prior to this change:
 - `afterFileParse`
 - `beforeFileValidate`
 - `onFileValidate`
 - `afterFileValidate`
 - `beforeProgramValidate`

Event flow after this change:
- `afterFileParse`
- `beforeProgramValidate`
- `beforeFileValidate`
- `onFileValidate`
- `afterFileValidate`

### DIFF
--- a/src/Program.spec.ts
+++ b/src/Program.spec.ts
@@ -2406,6 +2406,7 @@ describe('Program', () => {
             };
             program.plugins.add(plugin);
             program.addOrReplaceFile('source/main.brs', '');
+            program.validate();
             expect(plugin.beforeFileValidate.callCount).to.equal(1);
             expect(plugin.onFileValidate.callCount).to.equal(1);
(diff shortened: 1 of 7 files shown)


TwitchBronBron:
@elsassph @georgejecook @markwpearce @triwav @chrisdp do any of you have concerns with this change in the event flow for v0?

georgejecook:
oh gosh.. let me check.

PR #507, opened 2022-02-08
Adds TypeExpressions

markwpearce:
Since Types are getting more complicated now, `TypeToken`s in `Expressions` and `Statements` have been refactored to being `TypeExpressions`

Also adds support for multidimensional arrays:

```
sub doLoop(data as float[][])
  for each row in data
    for each item in row
      print item ' Brighterscript knows item is a float
    end for
  end for
end sub
```

### DIFF
--- a/src/parser/Expression.ts
+++ b/src/parser/Expression.ts
@@ -17,6 +17,7 @@ import type { BscType, SymbolContainer } from '../types/BscType';
 import { SymbolTable } from '../SymbolTable';
 import { FunctionType } from '../types/FunctionType';
 import { ObjectType } from '../types/ObjectType';
+import { ArrayType } from '../types/ArrayType';
 
 export type ExpressionVisitor = (expression: Expression, parent: Expression) => void;
 
@@ -131,7 +132,7 @@ export class FunctionExpression extends Expression implements TypedefProvider {
         readonly leftParen: Token,
         readonly rightParen: Token,
         readonly asToken?: Token,
-        readonly returnTypeToken?: Token,
+        readonly returnType?: TypeExpression,
         /**
          * If this function is enclosed within another function, this will reference that parent function
          */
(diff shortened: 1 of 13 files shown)


src/util.ts
```diff
@@ -982,12 +981,7 @@ export class Util {
             case TokenKind.Void:
                 return new VoidType();
             case TokenKind.Identifier:
-                let tokenText = token.text.replace(/\s/g, '').toLowerCase();
```
markwpearce, line 985: Array stuff is handled in `TypeExpression` now

src/parser/Parser.ts
```diff
@@ -1199,6 +1197,13 @@ export class Parser {
             });
             throw this.lastDiagnosticAsError();
         }
+        let itemType = new DynamicType();
```
markwpearce, line 1200: Moved the typing of the for-each item to BEFORE the parsing of the inside, because there could be an inside loop that was iterating over some property of the outside item

src/parser/Expression.ts
```diff
@@ -131,7 +132,7 @@ export class FunctionExpression extends Expression implements TypedefProvider {
         readonly leftParen: Token,
         readonly rightParen: Token,
         readonly asToken?: Token,
-        readonly returnTypeToken?: Token,
+        readonly returnTypeExpr?: TypeExpression,
```
TwitchBronBron, line 135: Don't abbreviate this. I think we could probably just call it `returnType`

src/parser/Expression.ts
```diff
@@ -280,7 +281,7 @@ export class FunctionParameterExpression extends Expression {
         public equalsToken?: Token,
         public defaultValue?: Expression,
         public asToken?: Token,
-        public typeToken?: Token,
+        public typeExpr?: TypeExpression,
```
TwitchBronBron, line 284: Thinking we can call this `type`, remove the `type: BscType` parameter, and instead leverage the `getType()` metohd on `TypeExpression`?

src/parser/Expression.ts
```diff
@@ -1478,6 +1479,74 @@ export class RegexLiteralExpression extends Expression {
     }
 }
 
+export class TypeExpression extends Expression {
+    public constructor(
+        public typeToken: Token,
+        public bracketPairs: Array<[Token, Token]>,
```
TwitchBronBron, line 1485: We should have a dedicated `ArrayTypeExpression` class for this instead.

TwitchBronBron (changes requested):
I'm loving the progress you've been making on this stuff! I had a few comments. 

Let's not abbreviate `expression` to `expr` in any public or private property. I think most of them could lose the `expression` suffix entirely, since we don't use it other places in the AST. But if we must keep it, don't abbreviate.

src/parser/Expression.ts
markwpearce: It's a little harder with this one, because the type of the parameter is derived either from the explicit type given (e.g. `as float`) OR from the initial value, and the initial value could be a functionCall, etc...

markwpearce:
there's a little bit of cleanup needed in the `Interface` PR after this is merged. I'd like to unify (by creating a TypeScript interface) getting a "type" from a statement or expression.

Issue #517, opened 2022-02-14
`as Function` shows error with parameter function that includes parameters


TwitchBronBron:
This produces an error, but should not. 

```typescript
it.only('allows any functional param to match `as Function` ', () => {
    program.setFile('source/main.brs', `
        sub callFuncWithParam(func as Function)
            print func
        end sub

        sub printValue(value)
            print value
        end sub

        sub main()
            callFuncWithParam(printValue)
        end sub
    `);

    program.validate();
    expectZeroDiagnostics(program);
});
```

![image](https://user-images.githubusercontent.com/2544493/153939958-64afbf28-e1b9-43e3-a034-63813f241bbd.png)

PR #521, opened 2022-02-18
Adds UniversalFunctionType for functions as params.

### LINKED ISSUE #516: `as function` doesn't support parameters
This produces an error. I believe that `as Function` should include all function permutations (any params, any return type, void return type). 

```typescript
it('allows `Function` to have zero or more parameters', () => {
    program.setFile('source/main.brs', `
        sub callFuncWithParam(func as Function, value)
            func(value)
        end sub

        sub printValue(value)
            print value
        end sub

        sub main()
            callFuncWithParam(printValue, "hello world")
        end sub
    `);

    program.validate();
});
```

image

markwpearce:
Fixes #516 and #517

### DIFF
--- a/src/files/BrsFile.ts
+++ b/src/files/BrsFile.ts
@@ -22,7 +22,7 @@ import { BrsTranspileState } from '../parser/BrsTranspileState';
 import { Preprocessor } from '../preprocessor/Preprocessor';
 import { LogLevel } from '../Logger';
 import { serializeError } from 'serialize-error';
-import { isClassMethodStatement, isClassStatement, isCommentStatement, isDottedGetExpression, isFunctionStatement, isFunctionType, isLibraryStatement, isNamespaceStatement, isStringType, isVariableExpression, isXmlFile, isImportStatement, isClassFieldStatement, isEnumStatement, isArrayType, isCustomType, isDynamicType, isObjectType, isPrimitiveType, isRegexLiteralExpression, isUniversalFunctionType } from '../astUtils/reflection';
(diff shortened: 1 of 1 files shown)


markwpearce:
- [ ] Add test for using return type from a function provided as a parameter
- [ ] Add test for using return type from a function provided as a parameter in an assignement

markwpearce:
Another approach is to just have one FunctionType but have a `isGeneric` flag on the constructor. Pros of this approach is that really, the type defined `as Function` is the same an actual function declaration.

TwitchBronBron:
I'm struggling with the term `generic`, because we will eventually support [generics](https://www.typescriptlang.org/docs/handbook/2/generics.html). What about something like `GeneralFunction`, `NonSpecificFunction`, `AllPurposeFunction`, `CatchallFunction`, `TopLevelFunction`. Any of those?

I'm leaning towards keeping your current design (separate dedicated type). It helps keep the current FunctionType logic a little cleaner. But I could be convinced either way. Thoughts?

markwpearce:
I hear what you're saying about `GenericFunction` ... How about `UniversalFunction` 

I think the best way would be to have `UniversalFunctionType` extend `FunctionType`. `ReturnType` would always return Dynamic, and the `addParam` method would be noops.

Then the `isFunctionType` reflection function look for either `FunctionType` or `UniversalFunctionType`.

This way we don't need special case handling everywhere in the code.

TwitchBronBron (approved):

PR #526, opened 2022-02-23
Static AST ranges

TwitchBronBron:
The `range` property on brightscript AST nodes represents the locations of the original tokens in the file. However, many of our AST nodes have dynamic getters for range that might also incorporate plugin-injected locations, which would produce incorrect source maps. 

This PR results in _most_ `range` properties being computed in the constructor, eliminating the range getter. `NamespaceStatement`, `ClassMethodStatement`, `FunctionStatement`, `FunctionExpression`, and `AnnotationExpression` still have dynamic range getters. This is because those statements are constructed before they have been completed, which are attached later. To handle their range calculations, we call the `get range` getter after they are constructed, which finalizes the range property.


src/parser/Parser.ts
```diff
@@ -3543,3 +3559,10 @@ export function getBscTypeFromExpression(expression: Expression, functionExpress
     //fallback to dynamic
     return new DynamicType();
 }
+
+/**
+ * A function that does nothing
+ */
+function noop(...args) {
```
chrisdp, line 3566: What is this for? lol

TwitchBronBron: It's used in relation to this part of my description: `To handle their range calculations, we call the get range getter after they are constructed, which finalizes the range property.` 

Here's one of the places it's used: https://github.com/rokucommunity/brighterscript/pull/526/files#diff-0d27a1b5098cc3c9f464ebe55376e540e5ffd0779dd682227eb284abf03e6478R680

Which results in things like this getting cached at parse-time: https://github.com/rokucommunity/brighterscript/pull/526/files#diff-8f54ec4bf9e1be0f718a7b8712e8ba4f9df4f443432f19e34e1b60aed9551874R159

chrisdp: hummmm, could we not do this by having a function on the expression it self? like `expression.casheRange()` or something? It's kind of a magic thing thats happening when done in this was currently.

TwitchBronBron: Yeah, fair enough. I'll change it.

TwitchBronBron: Changed. Please review again.

chrisdp (approved):

PR #527, opened 2022-02-24
Fix memory leak

TwitchBronBron:
Fixes a memory leak when replacing files. 

**Before**
ram-bad
**After**
ram-better


elsassph:
Plenty of typos in the comments but cool!

TwitchBronBron:
@elsassph I fixed the typos, and found/fixed another memory leak!

TwitchBronBron:
As a result of this fix (and increasing my nodejs ram limit to 8gig to allow the benchmark to finish), we should see significant performance boosts?

image

triwav (approved):

PR #532, opened 2022-03-05
Fix logger.time() not returning accurate timings

luis-soares-sky:
While diagnosing why the compiler is taking so long in our project, we noticed that the reported time is not correct.

image-before

After debugging a bit, I found that `stopwatch.stop()` was called right after `action(...)`, which doesn't work for async methods (which "Parsing files"/`loadAllFilesAST` is).

Moving the stop inside the `done` callback seems to do the trick.

image-after


TwitchBronBron:
Oh, good catch!

TwitchBronBron:
@luis-soares-sky this was included in release v0.45.4

PR #544, opened 2022-03-22
Better handling of `object` types

TwitchBronBron:
In BrightScript, the `object` type can represent every type except `uninitialized`. This PR aligns the type system with that assumption.


markwpearce:
Wow. I was unaware of how that worked. :(

I suppose this means we should implement `AssocArray` as a type in Brighterscript... In fact, probably all of the types listed here: https://developer.roku.com/en-ca/docs/references/scenegraph/xml-elements/interface.md

markwpearce (approved):

PR #546, opened 2022-03-23
Optional chaining

TwitchBronBron:
Adds syntax support for optional chaining. Since the brightscript runtime supports this natively in Roku OS 11, there's no need for any special transpile logic at this time. However, there are significant dangers of crashing in the native implementation, so I think we will still add a transpile step in the future to make this feature more useful.

This PR is just to support the syntax from a parser level.

### DIFF
--- a/src/files/tests/optionalChaning.spec.ts
+++ b/src/files/tests/optionalChaning.spec.ts
@@ -43,4 +43,39 @@ describe('optional chaining', () => {
             end sub
         `);
     });
+
+    it(`transpiles '?.[`, () => {
+        testTranspile(`
+            sub main()
+                print m?["value"]
+            end sub
+        `);
+    });
+
+    it(`transpiles '?(`, () => {
+        testTranspile(`
+            sub main()
+                print m.notExists?()
+            end sub
+        `);
+    });
+
+    it('transpiles various use cases', () => {
+        testTranspile(`
+            print arr?.["0"]
+            print arr?.value
+            print assocArray?.[0]
+            print assocArray?.getName()?.first?.second
+            print createObject("roByteArray")?.value
+            print createObject("roByteArray")?["0"]
+            print createObject("roList")?.value
+            print createObject("roList")?["0"]
+            print createObject("roXmlList")?["0"]
+            print createObject("roDateTime")?.value
+            print createObject("roDateTime")?.GetTimeZoneOffset
+            print createObject("roSGNode", "Node")?[0]
+            print pi?.first?.second
+            print success?.first?.second
+        `);
+    });
 });
--- a/src/files/tests/optionalChaning.spec.ts
+++ b/src/files/tests/optionalChaning.spec.ts
@@ -52,6 +52,25 @@ describe('optional chaining', () => {
         `);
     });
 
+    it(`transpiles '?@`, () => {
+        testTranspile(`
+            sub main()
+                print xmlThing?@someAttr
+            end sub
+        `);
+    });
+
+    it(`transpiles '?(`, () => {
+        testTranspile(`
+            sub main()
+                localFunc = sub()
+                end sub
+                print localFunc?()
+                print m.someFunc?()
+            end sub
+        `);
+    });
+
     it('transpiles various use cases', () => {
         testTranspile(`
             print arr?.["0"]
@@ -68,6 +87,8 @@ describe('optional chaining', () => {
             print createObject("roSGNode", "Node")?[0]
             print pi?.first?.second
             print success?.first?.second
+            print a.b.xmlThing?@someAttr
+            print a.b.localFunc?()
         `);
     });
 });
--- a/src/files/tests/optionalChaning.spec.ts
+++ b/src/files/tests/optionalChaning.spec.ts
@@ -52,14 +52,6 @@ describe('optional chaining', () => {
         `);
     });
 
-    it(`transpiles '?(`, () => {
-        testTranspile(`
-            sub main()
-                print m.notExists?()
-            end sub
-        `);
-    });
-
     it('transpiles various use cases', () => {
         testTranspile(`
             print arr?.["0"]
--- a/src/files/tests/optionalChaning.spec.ts
+++ b/src/files/tests/optionalChaning.spec.ts
@@ -1,0 +1,46 @@
+import * as sinonImport from 'sinon';
+import * as fsExtra from 'fs-extra';
+import { Program } from '../../Program';
+import { standardizePath as s } from '../../util';
+import { getTestTranspile } from '../../testHelpers.spec';
+
+let sinon = sinonImport.createSandbox();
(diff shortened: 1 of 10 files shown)


src/files/tests/optionalChaning.spec.ts
```diff
@@ -0,0 +1,81 @@
...
+        testTranspile(`
+            sub main()
+                print m.notExists?()
+            end sub
+        `);
+    });
+
+    it('transpiles various use cases', () => {
+        testTranspile(`
+            print arr?.["0"]
```
philippe-elsass-deltatre, line 65: Does both `arr?.[` and `arr?[` work?!

src/parser/Parser.ts
```diff
@@ -1449,12 +1449,12 @@ export class Parser {
...
-        this.warnIfNotBrighterScriptMode('ternary operator');
+    private ternaryOrIndexedGet(test?: Expression): TernaryExpression | IndexedGetExpression {
         if (!test) {
             test = this.expression();
         }
         const questionMarkToken = this.advance();
+        const start = this.current;
 
         //consume newlines or comments
         while (this.checkAny(TokenKind.Newline, TokenKind.Comment)) {
```
philippe-elsass-deltatre, line 1460: Should that immediately bust out the possibility of an indexed get?

src/parser/Parser.ts
```diff
@@ -2372,29 +2382,26 @@ export class Parser {
                 expr = this.finishCall(this.previous(), expr);
                 //store this call expression in references
                 referenceCallExpression = expr;
-            } else if (this.match(TokenKind.LeftSquareBracket)) {
+                //Indexed get. Also supports worthless leading dot. example: prop.["someKey"]
```
philippe-elsass-deltatre, line 2385: "worthless" -> "optional"

src/parser/Parser.ts
```diff
@@ -2830,6 +2856,22 @@ export class Parser {
...
+    /**
+     * Get the token that is {diff} indexes away from {this.current}
+     * @param diff the number of steps away from current index to fetch
+     * @param tokenKinds the desired token must match one of these
+     * @example
+     * getToken(-1); //returns the previous token.
+     * getToken(0);  //returns current token.
+     * getToken(1);  //returns next token
+     */
+    private getToken(diff: number, ...tokenKinds: TokenKind[]): Token {
```
philippe-elsass-deltatre, line 2868: Maybe should be more explicit like `getPreviousTokenMatchingAny`

src/util.ts
```diff
@@ -891,6 +891,43 @@ export class Util {
         };
     }
 
+    /**
+     * Given a list of ranges, create a range that starts with the first non-null lefthand range, and ends with the first non-null
+     * righthand range. Returns undefined if none of the items have a range.
+     */
+    public createBoundingRange(...locatables: Array<{ range?: Range }>) {
```
philippe-elsass-deltatre, line 898: Very welcome and should be applied to a bunch of existing statements

src/files/tests/optionalChaning.spec.ts
TwitchBronBron: Yep! We tested it at runtime, and they both work. This might be inspiration for a new `consistent-optional-chaining-operator` bslint rule.

src/parser/Parser.ts
TwitchBronBron: Yeah that's a better way to word it. :D

src/parser/Parser.ts
TwitchBronBron: Good point. I renamed it `getMatchingTokenAtOffset`

src/util.ts
TwitchBronBron: Agreed. This has been in the `v1` branch for a while now, and most of the AST nodes were refactored to use this function there. I should probably do something similar in master at some point since that is a non-breaking change.

src/parser/Parser.ts
TwitchBronBron: Yes, we can assume it won't be an indexedGet if we see any token other than a left square brace. However, this function parses ternary by default and then scraps ternary in favor of optional chained indexedGet. So, there's really nothing to bust out of here.

src/files/tests/optionalChaning.spec.ts
```diff
@@ -0,0 +1,73 @@
...
+    it(`transpiles '?.[`, () => {
+        testTranspile(`
+            sub main()
+                print m?["value"]
+            end sub
+        `);
+    });
+
+    it('transpiles various use cases', () => {
+        testTranspile(`
```
chrisdp, line 56: missing the `?@` uscases.

src/parser/Parser.spec.ts
```diff
@@ -139,6 +139,88 @@ describe('parser', () => {
         });
     });
 
+    describe('optional chaining operator', () => {
```
chrisdp, line 142: Should we have something in there that validates it doesn't match something like:

```
?["test"]
```
It should be some sort of print. Not sure what test area that would go in.

src/parser/tests/expression/TernaryExpression.spec.ts
```diff
@@ -274,30 +274,30 @@ describe('ternary expressions', () => {
...
-            testTranspile(
-                `a = user = invalid ? 1 : "logged in"`,
-                `a = bslib_ternary(user = invalid, 1, "logged in")`
-            );
-
-            testTranspile(
-                `a = user = invalid ? 1.2 : "logged in"`,
-                `a = bslib_ternary(user = invalid, 1.2, "logged in")`
-            );
+            // testTranspile(
```
arturocuya, line 277: Could be a good idea to mention why these are left here. Either they are not going to be used and were forgotten, or will be included someday in the test, in which case a TODO comment is appreciated :+1:

src/files/tests/optionalChaning.spec.ts
TwitchBronBron: Good catch. the `?@` wasn't transpiling properly. :D

src/parser/tests/expression/TernaryExpression.spec.ts
TwitchBronBron: Oops! Yeah, those should not have been commented out. Re-enabled, and tests still pass!

src/parser/Parser.spec.ts
TwitchBronBron: Oops. Good catch. Yeah, I need to tweak the lexer to handle this situation. I need to look to the left of the question mark, and if it's preceeded by only whitespace and then a newline or colons, then the `?` should be its own token, separate from the trailing `[` or `(`

TwitchBronBron: I fixed it in d317171a9c18c9f35bbda5592f56c30e3ff8a399

chrisdp (approved):

PR #554, opened 2022-04-02
Create Object Validation, targeted to V1 Branch 

markwpearce:
See: https://github.com/rokucommunity/brighterscript/pull/435

Since the typing system in v1 is so different than v0, I think we should target v1 for validation and typing of built in types.

This PR is a rebase of `create-object-validation` onto the `v1` branch. All the future tasks from the above PR apply here too.


TwitchBronBron:
I know the typing is different between v0 and v1, but there are still several significant benefits we can add to v0 related to these changes. I've been thinking about adding a platform.d.bs file into brighterscript v0 which would add many interfaces and types generated based on these docs. Until v1 is more stabilized, I want to make sure to try and keep developing stuff against v0 whenever possible. Anyway, these are just some things I've been bouncing around in my head.

markwpearce:
@TwitchBronBron Cool. I'll close this, and try to get some of the fixes needed for the v0 version of this branch.

PR #561, opened 2022-04-07
Added additional logic to check for mismatch in CreateObject argument count

markwpearce:
I know I was supposed to add this directly to the `create-object-validation` branch, but I thought this was enough of a change that you may want to review it seperately.

The scrape task wouldn't run with whatever current versions of `marked` and other packages I installed via `npm ci`, 

So I fixed the scripts so they would run again.

I also noticed a few docs had bad Brightscript in the CreateObject() example. I added a bit of code to the constructor checker to still gracefully find parameters when the docs were "c" style:

https://developer.roku.com/en-ca/docs/references/brightscript/components/roregion.md:
```
CreateObject("roRegion", Object bitmap, Integer x, Integer y,Integer width, Integer height)
```

or just missing the "as":

https://developer.roku.com/en-ca/docs/references/brightscript/components/roarray.md
```
CreateObject("roArray", size As Integer, resizeAs Boolean)
```

All this was because I wanted to check to make sure there were 6 arguments in a `CreateObject("roRegion", ...)` call, but `roRegion` wasn't being scraped properly... haha


TwitchBronBron (approved):
Looks great. Thanks for these changes!

Issue #259, opened 2020-12-19
use `new` syntax for `CreateObject` calls


TwitchBronBron:
Support using the `new` keyword as shorthand for `CreateObject` things. For example, this:
```BrightScript
newXfer = CreateObject("roUrlTransfer")
```
would become this:
```BrightScript
newXfer = new RoUrlTransfer()
```

This would obviously prevent you from creating classes with the same name as all the `CreateObject` things, but I don't think that would be a big deal since most of them are prefixed with `"ro"` anyway. 

This would also need to support passing in constructor arguments, which would become the arguments to `CreateObject`. 

This would require building a complete list of valid strings for `CreateObject`, and then we could also validate the constructor arguments.

PR #569, opened 2022-04-20
Adds 'new Component()' syntax as shorthand for CreateObject()

markwpearce:
Implements Issue #259 

Uses `new` syntax as syntactic sugar to replace the `CreateObject()` call

eg.:

```
poster = new roSGNode("Poster")
regex = new roRegex("[a-z]+", "i")
bitmap = new roBitmap("somePath")
```

transpiles to:

```
poster = CreateObject("roSGNode", "Poster")
regex = CreateObject("roRegex", "[a-z]+", "i")
bitmap = CreateObject("roBitmap", "somePath")
```

All current `createObject` validation (wrong number of arguments, deprecation, etc), is applied to these calls as well.

image

### DIFF
--- a/src/parser/Expression.ts
+++ b/src/parser/Expression.ts
@@ -939,6 +939,56 @@ export class NewExpression extends Expression {
     }
 }
 
+/**
+ * This expression changes a call expression into a createObject call,
+ * eg `new roSGNode("Poster")` transpiles to `CreateObject("roSGNode", "Poster")`
+ */
+export class NewCreateObjectExpression extends Expression {
+    constructor(
+        readonly newKeyword: Token,
+        readonly call: CallExpression
+    ) {
+        super();
+        this.range = util.createRangeFromPositions(this.newKeyword.range.start, this.call.range.end);
+    }
+
+    /**
+     * The name of the component to initialize (with optional namespace prefixed)
+     */
+    public get componentName() {
+        //the parser guarantees the callee of a new statement's call object will be
+        //a NamespacedVariableNameExpression
+        return (this.call.callee as NamespacedVariableNameExpression);
+    }
+    public readonly range: Range;
+
+    public transpile(state: BrsTranspileState) {
+        let result = [];
+        result.push(
+            'CreateObject',
+            state.transpileToken(this.call.openingParen),
+            state.sourceNode(this.call.callee, ['"', this.componentName.getName(ParseMode.BrightScript), '"'])
+        );
+        //transpile args
+        for (let i = 0; i < this.call.args.length; i++) {
+            //add comma between args
+            result.push(', ');
+            let arg = this.call.args[i];
+            result.push(...arg.transpile(state));
+        }
+        result.push(
+            state.transpileToken(this.call.closingParen)
+        );
+        return result;
+    }
+
+    walk(visitor: WalkVisitor, options: WalkOptions) {
+        if (options.walkMode & InternalWalkMode.walkExpressions) {
+            walk(this, 'call', visitor, options);
+        }
+    }
+}
+
 export class CallfuncExpression extends Expression {
     constructor(
         readonly callee: Expression,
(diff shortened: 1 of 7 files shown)


georgejecook:
Looks useful.

Question: Would this prevent us using the new keyword from doing something like the following in the future?
```
myTextInput = new MyCustomTextInput(argThatGetsPassedToConstructor1, argThatGetsPassedToConstructor2)
```

which transpiles to something more or less like

```
myTextinput = createSGNode("MyCustomTextInput")
myTexTInput.callfunc("_initialize", argThatGetsPassedToConstructor1, argThatGetsPassedToConstructor2)
```

or 

```
myPoster = new Poster()
```

markwpearce:
> Question: Would this prevent us using the new keyword from doing something like the following in the future?

The issue here is overloading `new` ... does it mean:

1. Instantiate an instance of a class?
2. Use `CreateObject`? eg. `CreateObject(<component name>)`
3. Use `CreateObject` but create a new SG Node - eg. `CreateObject("roSGNode", <component name>)`

Since in Brighterscript, the first option already exists, I elected here to only add the second.

It doesn't preclude the use of `new` for the third option in the future, but it does muddy exactly what is happening.

src/parser/Expression.ts
```diff
@@ -939,6 +939,56 @@ export class NewExpression extends Expression {
     }
 }
 
+/**
+ * This expression changes a call expression into a createObject call,
+ * eg `new roSGNode("Poster")` transpiles to `CreateObject("roSGNode", "Poster")`
+ */
+export class NewCreateObjectExpression extends Expression {
```
TwitchBronBron, line 946: Let's not create a separate statement for this. At the end of the day, it's still just a `NewExpression` semantically.

src/parser/Expression.ts
```diff
@@ -939,6 +939,56 @@ export class NewExpression extends Expression {
...
+     * The name of the component to initialize (with optional namespace prefixed)
+     */
+    public get componentName() {
+        //the parser guarantees the callee of a new statement's call object will be
+        //a NamespacedVariableNameExpression
+        return (this.call.callee as NamespacedVariableNameExpression);
+    }
+    public readonly range: Range;
+
+    public transpile(state: BrsTranspileState) {
```
TwitchBronBron, line 965: This logic should probably live in BrsFilePreTranspileProcessor plugin function. Either that, or we have an if statement inside NewExpression that checks whether this is a ro[Whatever] or a class.

TwitchBronBron (changes requested):

src/parser/Expression.ts
markwpearce: I don't understand how that works or why it should be in BrsFilePreTranspileProcessor.... Can you elaborate?

src/parser/Expression.ts
markwpearce: Hmm... What about using NewExpression, but having a public flag on that class that says if this `isCreateObjectCall` or something?
We still want to be able to tell external to the new expression if it's for `createObject` for validation, etc.

TwitchBronBron: In the future, the entire list of those `ro*` components will be supplied from type definitions, some of which might be user-defined (to support if Roku has super-secret components we're not allowed to know about), so we won't know at parse time what the item should resolve to. Plus, if we eventually support something like `new Poster()`, you definitely won't know at parse time because it could be a user-defined component. 

With that in mind, how would you be able to create the flag? What if the component we're calling `new` on hasn't been parsed yet?

src/parser/Expression.ts
TwitchBronBron: I've been slowly moving lots of functionality into the BscPlugin for brighterscript instead of hardcoding them in the AST itself. However, thinking more about this, I don't think this specific implementation should be in the plugin. So, ignore the `BrsFilePreTranspileProcessor` comment. This should be moved into the `NewExpression` transpile function.

PR #573, opened 2022-04-23
Config for force transpile brs file 

justinMBullard:
Configuration setting added to BsConfig to allow for BRS file to be transpiled like BS file.

### DIFF
--- a/src/BsConfig.ts
+++ b/src/BsConfig.ts
@@ -152,4 +152,10 @@ export interface BsConfig {
      * @default true
      */
     sourceMap?: boolean;
+
+    /**
+     * Force BRS file to be compiled using BrigtherScript
+     * @default false
+     */
+    forceBrighterScript?: boolean;
 }
--- a/src/BsConfig.ts
+++ b/src/BsConfig.ts
@@ -154,8 +154,8 @@ export interface BsConfig {
     sourceMap?: boolean;
 
     /**
-     * Force BRS file to be compiled using BrigtherScript
+     * Force BRS file to be transpiled
      * @default false
      */
-    forceBrighterScript?: boolean;
+     transpileBrs?: boolean;
 }
--- a/src/files/BrsFile.ts
+++ b/src/files/BrsFile.ts
@@ -48,7 +48,7 @@ export class BrsFile {
         this.extension = util.getExtension(this.pathAbsolute);
 
         //all BrighterScript files need to be transpiled
-        if (this.extension?.endsWith('.bs')) {
+        if (this.extension?.endsWith('.bs') || program?.options?.forceBrighterScript) {
             this.needsTranspiled = true;
             this.parseMode = ParseMode.BrighterScript;
         }
--- a/src/files/BrsFile.ts
+++ b/src/files/BrsFile.ts
@@ -48,7 +48,7 @@ export class BrsFile {
         this.extension = util.getExtension(this.pathAbsolute);
 
         //all BrighterScript files need to be transpiled
-        if (this.extension?.endsWith('.bs') || program?.options?.forceBrighterScript) {
+        if (this.extension?.endsWith('.bs') || program?.options?.transpileBrs) {
             this.needsTranspiled = true;
             this.parseMode = ParseMode.BrighterScript;
         }
--- a/src/util.ts
+++ b/src/util.ts
@@ -323,7 +323,7 @@ export class Util {
         config.showDiagnosticsInConsole = config.showDiagnosticsInConsole === false ? false : true;
         config.sourceRoot = config.sourceRoot ? standardizePath(config.sourceRoot) : undefined;
         config.cwd = config.cwd ?? process.cwd();
-        config.forceBrighterScript = config.forceBrighterScript === true ? true : false;
+        config.transpileBrs = config.transpileBrs === true ? true : false;
         config.emitDefinitions = config.emitDefinitions === true ? true : false;
         if (typeof config.logLevel === 'string') {
             config.logLevel = LogLevel[(config.logLevel as string).toLowerCase()];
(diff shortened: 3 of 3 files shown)


src/files/BrsFile.ts
```diff
@@ -48,7 +48,7 @@ export class BrsFile {
         this.extension = util.getExtension(this.pathAbsolute);
 
         //all BrighterScript files need to be transpiled
-        if (this.extension?.endsWith('.bs')) {
+        if (this.extension?.endsWith('.bs') || program?.options?.transpileBrs) {
             this.needsTranspiled = true;
             this.parseMode = ParseMode.BrighterScript;
```
elsassph, line 53: Is it intended to force `.brs` files to be parsed with `.bs` features? (e.g. classes and all)

justinMBullard: @elsassph yes it is. I am trying to support legacy code while enabling adoption of Brighterscript features. Switching all the .brs files to .bs would add a level of undesirable complexity.

elsassph: I think it needs to be mentioned in the help/doc. And maybe a few things need to be double-checked as to whether this is sufficient to fully enable brighterscript features in brs files.

justinMBullard: @elsassph should I add this to https://github.com/rokucommunity/brighterscript#bsconfigjson-options?
I have been using this to enable "import" and it also works with the ternary operator.

elsassph: I guess if @TwitchBronBron is ok with this compiler change he'll adjust the doc accordingly.

Oh you should add this option to the CLI as well BTW.

TwitchBronBron:
Should we have separate flags for "run the brs code through the transpiler" and "allow brighterscript features within .brs files"? 

The `transpileBrs` option sounds like the first item, and doesn't make it immediately evident that you can now use brighterscript features  within .brs files. 

This also doesn't handle whether xml files should be transpiled or not. 

Should we have _two_ flags instead, that represent that following actions?
 - run brs and xml files through the transpiler (not sure why you'd need to do that though on its own)?
 - allow BrighterScript code inside .brs files

TwitchBronBron:
Currently, brightscript files that have zero AST changes will be copied as-is, and won't run through the transpiler. So in my head, that's a separate thing from "allow brighterscript features in brightscript files". However, that's just a product of the current design. Also, plugins that modify the AST would also need to transpile their files even if not using brighterscript syntax enhancements. Perhaps you're right, there is no real difference from an end-user perspective?

I've been thinking maybe we should name this so that it's more clear that you're allowing brighterscript features to be written within the .brs file extension. Maybe something like
- `allowBsInBrs`
- `allowBrighterScriptInBrightScript`
- `treatBrsAsBs`

elsassph:
I kind of like `transpileBrs`, but it should be clear what it means.

justinMBullard:
@TwitchBronBron @elsassph 
How about allowBrighterScriptInBrs or transpileBrsWithBrighterScript?

justinMBullard:
@TwitchBronBron  @elsassph Any decision on the name of the setting?

TwitchBronBron:
I'm leaning towards `allowBrighterScriptInBrightScript`. @elsassph @chrisdp any objections? I'd like to get this finalized early this week. It's waited long enough (I'm sorry @justinMBullard for the delay).

justinMBullard:
> I'm leaning towards `allowBrighterScriptInBrightScript`. @elsassph @chrisdp any objections? I'd like to get this finalized early this week. It's waited long enough (I'm sorry @justinMBullard for the delay).

@TwitchBronBron Should I make name change?

TwitchBronBron:
Yeah, go ahead and make the change. Sorry for being so slow on this. Let's go with `allowBrighterScriptInBrightScript`.

justinMBullard:
> Yeah, go ahead and make the change. Sorry for being so slow on this. Let's go with `allowBrighterScriptInBrightScript`.
@justinMBullard  I did it and also Rename transpile-br to allow-brighterscript-in-brightscript.

benchmarks/targets/transpile-brs.js
```diff
@@ -20,7 +20,7 @@ module.exports = async (suite, name, brighterscript, projectPath, options) => {
     }
 
     if (files.length === 0) {
-        console.log('[transpile-brs] No brs|bs|d.bs files found in program');
+        console.log('[allow-brighterscript-in-brightscript] No brs|bs|d.bs files found in program');
```
TwitchBronBron, line 23: I don't think this one needs changed. The name of this benchmark is "transpile-brs", and is unrelated to this new feature.

src/BsConfig.ts
```diff
@@ -152,4 +152,10 @@ export interface BsConfig {
      * @default true
      */
     sourceMap?: boolean;
+
+    /**
+     * Force BRS file to be transpiled
```
TwitchBronBron, line 157:
```suggestion
     * Allow brighterscript features (classes, interfaces, etc...) to be included in BrightScript (`.brs`) files, and force those files to be transpiled. 
```

src/cli.ts
```diff
@@ -10,6 +10,7 @@ let options = yargs
     .help('help', 'View help information about this tool.')
     .option('create-package', { type: 'boolean', defaultDescription: 'true', description: 'Creates a zip package. This setting is ignored when deploy is enabled.' })
     .option('source-map', { type: 'boolean', defaultDescription: 'false', description: 'Enables generating sourcemap files, which allow debugging tools to show the original source code while running the emitted files.' })
+    .option('allow-brighterscript-in-brightscript', { type: 'boolean', defaultDescription: 'false', description: 'Enables BrighterScript code om BrightScript files.' })
```
TwitchBronBron, line 13:
```suggestion
    .option('allow-brighterscript-in-brightscript', { alias: 'allow-brighter-script-in-bright-script', type: 'boolean', defaultDescription: 'false', description: 'Allow brighterscript features (classes, interfaces, etc...) to be included in BrightScript (`.brs`) files, and force those files to be transpiled..' })
```

I agree with this name, but I think the cli will provide it as `allowBrighterscriptInBrightscript` instead of `allowBrighterScriptInBrightScript`, so the cli argument won't actually get applied. To work around that, we can use an alias which will sync both options.

TwitchBronBron (changes requested):
Looks great. Just a few small suggestions.

TwitchBronBron (approved):

PR #575, opened 2022-04-27
Component Statement

TwitchBronBron:
Adds an in-code way to define SceneGraph components without needing to mess with xml files. 

TODO: add lengthier description as this feature progresses.

Is there a better name? I'm currently calling it `in-code component`. NodeClass is _alright_ but I really like using the word "component" in the name because that's what they transpile to.


chrisdp:
@TwitchBronBron we landed on Component Statement. Feel free to update the description.

PR #579, opened 2022-05-01
Adds enum types

markwpearce:
Adds `EnumType`s, and allows enums to be used as parameter types and return values.

Due to the weirdness of enums, I elected to actually define members as a different type than the Enum itself, but we can assign a member to a value of the EnumType.

When enums are used as parameter types, arguments are correctly validated.

### DIFF
--- a/src/Scope.ts
+++ b/src/Scope.ts
@@ -3,7 +3,7 @@ import { CompletionItemKind, Location } from 'vscode-languageserver';
 import chalk from 'chalk';
 import type { DiagnosticInfo } from './DiagnosticMessages';
 import { DiagnosticMessages } from './DiagnosticMessages';
-import type { CallableContainer, BsDiagnostic, FileReference, BscFile, CallableContainerMap, FileLink, FunctionCall, InheritableStatement, InheritableType } from './interfaces';
+import type { CallableContainer, BsDiagnostic, FileReference, BscFile, CallableContainerMap, FileLink, FunctionCall, InheritableStatement, InheritableType, NamedTypeStatement } from './interfaces';
 import type { Program } from './Program';
 import { BsClassValidator } from './validators/ClassValidator';
 import type { NamespaceStatement, Statement, FunctionStatement, ClassStatement, InterfaceStatement, EnumStatement } from './parser/Statement';
@@ -85,8 +85,8 @@ export class Scope {
      * @param name - The name, including the namespace of the interface if possible
      * @param containingNamespace - The namespace used to resolve relative names. (i.e. the namespace around the current statement trying to find the interface or class)
      */
-    public getNamedTypeStatement(name: string, containingNamespace?: string): InheritableStatement {
-        return this.getFileLink(name, containingNamespace)?.item;
+    public getNamedTypeStatement(name: string, containingNamespace?: string): NamedTypeStatement {
+        return this.getNamedTypeFileLink(name, containingNamespace)?.item;
     }
 
     /**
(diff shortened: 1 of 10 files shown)


src/Scope.ts
```diff
@@ -204,7 +232,7 @@ export class Scope {
     * @param namespaceName - the current namespace name
     */
     public hasInterface(ifaceName: string, namespaceName?: string): boolean {
-        return !!this.hasInterface(ifaceName, namespaceName);
+        return !!this.getInterface(ifaceName, namespaceName);
```
markwpearce, line 235: This was a infinite loop mistake that slipped by!

src/Scope.ts
```diff
@@ -762,6 +790,13 @@ export class Scope {
             const ifaceStmt = pair[1]?.item;
             ifaceStmt?.buildSymbolTable(this.getParentInterface(ifaceStmt));
         }
+
+        //also link enums
+        const enumMap = this.getEnumMap();
```
markwpearce, line 795: I don't think it strictly necessary to build the symbol tables here instead of at parse time, but i don't think it hurts

src/SymbolTable.ts
```diff
@@ -118,6 +118,22 @@ export class SymbolTable {
                     const existingType = getTypeFromContext(symbol.type, context);
                     // if the type is not known yet, imply that the first assignment is the type
                     inferredType = inferredType ?? existingType;
+
+                    // TODO TYPES ... maybe this should look as 'convertibility' to infer a type
```
markwpearce, line 122: I know eventually there will be union types, but ....

src/Scope.ts
```diff
@@ -137,11 +137,39 @@ export class Scope {
...
+     * @param containingNamespace - The namespace used to resolve relative enum names. (i.e. the namespace around the current statement trying to find a enum)
+     */
+    public getEnumFileLink(enumName: string, containingNamespace?: string): FileLink<EnumStatement> {
+        const lowerName = enumName?.toLowerCase();
+        const enumMap = this.getEnumMap();
+
+        let enumeration = enumMap.get(
+            util.getFullyQualifiedClassName(lowerName, containingNamespace?.toLowerCase())
+        );
+        //if we couldn't find the iface by its full namespaced name, look for a global class with that name
```
TwitchBronBron, line 151:
```suggestion
        //if we couldn't find the enum by its full namespaced name, look for a global enum with that name
```

src/Scope.ts
TwitchBronBron: Yikes! nice catch.

src/Scope.ts
TwitchBronBron: I like it being here because that ensures we support future enum value types that come from other files.

src/files/BrsFile.ts
```diff
@@ -948,7 +948,7 @@ export class BrsFile {
             }
 
             if (symbolType?.memberTable) {
-                if (isCustomType(symbolType) || isInterfaceType(symbolType)) {
+                if (isCustomType(symbolType) || isInterfaceType(symbolType) || isEnumType(symbolType)) {
                     // we're currently looking at a customType or interface, that has it's own symbol table
```
TwitchBronBron, line 952:
```suggestion
                    // we're currently looking at a type that has its own symbol table
```

src/types/EnumType.ts
```diff
@@ -0,0 +1,66 @@
...
+
+    public equals(targetType: BscType): boolean {
+        return isEnumType(targetType) && targetType?.name.toLowerCase() === this.name.toLowerCase();
+    }
+}
+
+
+export class EnumMemberType implements BscType {
+    constructor(
+        public enumName: string, public memberName: string
```
TwitchBronBron, line 37:
```suggestion
        public enumName: string,
        public memberName: string
```

TwitchBronBron (changes requested):
Looks great! Just a few small suggestions.

TwitchBronBron (approved):

PR #582, opened 2022-05-06
Rename `ClassFieldStatement` and `ClassMethodStatement`

TwitchBronBron:
#575 will need to reuse the `ClassFieldStatement` and `ClassMethodStatement`. This PR renames those to `FieldStatement` and `MethodStatement` respectively so they apply more generically across the project.

Backwards compatibility is retained by some javascript trickery, but these items will be removed in v1.


markwpearce (approved):
Looks good.

Issue #512, opened 2022-02-10
Parse brs comment documentation


TwitchBronBron:
We should parse leading comments before all statements. This will allow us to gain more information about the statements, helping both brightscript and brighterscript. As a first step:

 - when parsing any statement, collect all sequential comments prior to the function and include them in a new `documentation` property on the `Statement`. Only one statement should "own" comments. Example:
    ```brightscript
    'Collect me
    ' 
    ' me too
    ' me three
    sub main()
    end sub
    ```
 - Many projects use the brightscript `'` token but then also add `/**` jsdoc blocks to help integrate with existing jsdoc parsers. If `/**` is detected, scan through every comment and remove leading `*` and `*/` tokens. Only do this if EVERY comment starts with one of those. That way, our parser is more flexible and supports existing practices.  Example: 
    ```brightscript
    ' /**
    '  * Some jsdoc style comment
    '  */
    function DoSomething()
    end function
    ```
- we should parse and collect `tags` which can be leveraged by our code, as well as any plugins that might find use for them. Tags are on their own line, and start with the `@` symbol. 
    ```brightscript
    ' Function that returns a name
    ' @param {string} firstName
    ' @return {string}
    function returnName(firstName as string)
        return firstName
    end function
    ```
 - needs to support the syntax of BrightScriptDoc explained [here](https://developer.roku.com/docs/developer-program/dev-tools/brightscript-doc.md). I envision most of these just get included as "tags" related to the previous bullet point, but just keep them in mind. 

I would imagine the `documentation` object on `Statement` would look something like this in typescript:
```typescript
class Statement {
    //...existing stuff
    public documentation: DocumentationStatement;
}
class DocumentationStatement{
    public comment: Token; //contains the leading comment
    public tags: TagExpression[];
    public range: Range;
}
class DocumentationTagExpression {
    public tokens: {
        at: Token; //the "@" symbol
        name: Identifier; //the word "param" in "@param", or "return" in "@return"
        leadingTypeCurlyBrace: Token;
        type: Identifier;
        trailingTypeCurlyBrace: Token;
        comment: Token; //any remaining unparsed text
    }
    public range: Range;
}
```

This ASTExplorer from typescript is very helpful to use for reference. 
![image](https://user-images.githubusercontent.com/2544493/153440449-547e0e09-ac56-44e9-a069-2b0f045fb250.png)

markwpearce:
Nice!

[...]

PR #585, opened 2022-05-12
Doc Scraper Fixes

markwpearce:
A few fixes to the doc scraper to grab a few more missing types

FWIW, here are a few fixes to the doc scrapper... There were a few places where the property names and such had extra slashes in them, and a few places where the types were not found or the like.

I originally wanted to look at generating Typedefs, by using `InterfaceStatement`, etc... but typedefs **should** have the comments in them, especially for `@param`, etc. and the Statements don't have those included.... so that work needs to wait until https://github.com/rokucommunity/brighterscript/issues/512 is addressed.


src/roku-types/data.json
```diff
@@ -10285,23 +10350,29 @@
...
...td></tr><tr><td>\"vp9\"</td><td>3.0, 3.1, 4.0, 4.1, 5.0, 5.1</td></tr></tbody></table></td><td>Optional</td></tr><tr><td>ChCnt</td><td>integer</td><td>Specifies the required number of audio channels</td><td>Optional</td></tr><tr><td>PassThru</td><td>integer</td><td>Specifies whether the bitstream needs to be decoded on the device. By default, this reports the codec properties of both the Roku device and the attached HDMI device.<br><br>You can filter the values returned by including one of the following options:<ul><li><strong>0</strong> - Reports only codecs that are supported on the Roku device itself.</li><li><strong>1</strong> - Reports only codecs of the attached HDMI device (for example, an audio receiver).</li></ul></td><td>Optional</td></tr><tr><td>Atmos</td><td>Integer</td><td>Specifies whether atmos is required. Include 'atmos=1' if it is.</td><td>Optional</td></tr><tr><td>SampleRate</td><td>integer</td><td>Specifies the sample rate</td><td>Optional</td></tr><tr><td>BitRate</td><td>integer</td><td>Specifies the bit rate in Kbit/sec</td><td>Optional</td></tr><tr><td>Container</td><td>string</td><td>Specifies the container format</td><td>Optional</td></tr></tbody></table>",
```
TwitchBronBron, line 10353: Should we be preventing adding these as raw html? (all the table html here)

src/roku-types/data.json
```diff
@@ -9595,7 +9654,7 @@
                             "description": "The [roScreen](https://developer.roku.com/docs/references/brightscript/components/roscreen.md or [roImageCanvas](https://developer.roku.com/docs/references/brightscript/components/roimagecanvas.md instance to be associated with this caption renderer.",
                             "isRequired": true,
                             "name": "screen",
-                            "type": "roScreen or roImageCanvas"
+                            "type": "Object"
```
TwitchBronBron, line 9657: "Object" is technically correct, but it would be good to eventually support multiple types for when we get union types in brighterscript. Would it be better to go ahead and store an array of types for this "type" key instead?

TwitchBronBron (commented):
Two quick questions, but if you don't think they're a big deal, go ahead and merge.

src/roku-types/data.json
markwpearce: Added `node-html-markdown`

markwpearce:
@TwitchBronBron -- I think this deals with all the issues that you raised previously

TwitchBronBron (changes requested):
Looks great! Looks like your branch is out of date with master. Once you update with master, I'm ready to merge this.

TwitchBronBron (approved):

PR #587, opened 2022-05-14
add diagnostic messages for overriden component functions

arturocuya:


src/Scope.ts
```diff
@@ -933,6 +933,63 @@ export class Scope {
...
+                                    shadowedCallable.callable.isSub ? 'sub' : 'function'
+                                ),
+                                range: container.callable.nameRange,
+                                file: container.callable.file
+                            });
+                        }
+
+                        // Raise a diagnostic error if the own callable return type is different
+                        // than the shadowed callable.
+                        if (container.callable.type.returnType.toTypeString() !== shadowedCallable.callable.type.returnType.toTypeString()) {
```
markwpearce, line 952: Should it be allowed to have a "more specific type" for the return type? Eg., use `isAssignableTo` ?

for example, shadowedCallable returns `dynamic`, but container returns `integer` - this is not technically an error, because container's return type (`integer`) can be assigned to a `dynamic`.

FWIW, Typescript uses assignability for evaluating if return types are valid on overridden functions

src/Scope.ts
```diff
@@ -933,6 +933,63 @@ export class Scope {
...
+                                    container.callable.params.length
+                                ),
+                                range: container.callable.nameRange,
+                                file: container.callable.file
+                            });
+                        }
+
+                        // Raise a diagnostic error if the own callable parameter types
+                        // are different than the shadowed callable.
+                        for (const [i, param] of container.callable.params.entries()) {
```
markwpearce, line 979: Likewise, if a function override uses a more specific type for a param, (eg. a type that can be assigned to the shadowedCallable's param type), it should not cause a diagnostic.

markwpearce (commented):
A few comments about type assignability

TwitchBronBron:
@arturocuya are you able to finish up this pr, or should I close it?

arturocuya:
@TwitchBronBron closing...

PR #600, opened 2022-05-25
Add object and key to visitor callbacks.

### LINKED ISSUE #477: Walk functionality should return the key path from the parent to the child
All of the walk callbacks should provide a way to get from the parent to that child. Like a string key path. For instance, `'Block.statements[0].name'`

TwitchBronBron:
- Provides the parent object and the key used to fetch the item from the parent object in AST walk visitor callbacks. This makes AST editing a little easier when it's desired to actually remove the element from the parent AST. 
 - update all AST `walk` methods to support adding/removing items during the walk cycle without messing up the iteration.

Also 
Resolves #477

### DIFF
--- a/src/astUtils/visitors.ts
+++ b/src/astUtils/visitors.ts
@@ -11,7 +11,7 @@ import type { AstEditor } from './AstEditor';
  */
 export function walkStatements(
     statement: Statement,
-    visitor: (statement: Statement, parent?: Statement) => Statement | void,
+    visitor: (statement: Statement, parent?: Statement, owningObject?: any, key?: any) => Statement | void,
     cancel?: CancellationToken
 ): void {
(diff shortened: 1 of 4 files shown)


src/astUtils/visitors.ts
```diff
@@ -11,7 +11,7 @@ import type { AstEditor } from './AstEditor';
  */
 export function walkStatements(
     statement: Statement,
-    visitor: (statement: Statement, parent?: Statement) => Statement | void,
+    visitor: (statement: Statement, parent?: Statement, owningObject?: any, key?: any) => Statement | void,
```
TwitchBronBron, line 14: @chrisdp what do you think of the name `owningObject`? Do you hate it? love it? have an alternative?

chrisdp (approved):

PR #601, opened 2022-05-26
Allow Enum and Interface return types

georgejecook:
Fixes bug preventing enums and interfaces from being used as function return types.


TwitchBronBron (approved):
Looks great! Thanks for the test too. :)

PR #603, opened 2022-05-26
fixes enums and interfaces resulting in diagnostics error when used as field types

georgejecook:


TwitchBronBron:
Nothing to fix here. :)

PR #606, opened 2022-05-29
Add hover to namespace functions

arturocuya:
Fix problem where namespace functions don't get a hover.

```
'  v IDK what's supposed to happen when I hover here, in the ns name. Right now we get no hover
mySpace.myFunc()
'        ^ hover here works now, in the func name
```


TwitchBronBron (approved):
Looks great! Thanks for this.

PR #613, opened 2022-06-02
Load projects based on bsconfig.json presence

TwitchBronBron:
Enhances the language server to scan the project for all bsconfig.json files. If any are found, each is treated like its own bright(er)script project. If no bsconfig file is found, then the root of the workspace is treated as a bright(er)script project instead. 

This will cause problems for any projects that have a root bsconfig _and_ a nested bsconfig. The mitigation for that issue is to rename the outer (or inner) bsconfig.json to something else, like `bsconfig.root.json` for example.

**Notable changes:**
- renamed the word `workspace` to `project` in the language server in many places. Now the core item is a `project`, and 1 workspace can have many projects
- treat the root workspace as project _only_ if there are no bsconfig.json files nested in the workspace folder
- don't crash language server when bsconfig.json has syntax errors
- better handling when adding/removing bsconfig.json files (i.e. load/unload projects accordingly)
- eliminates the need for vscode code-workspaces in most cases.

**TODO:**

- [x] unit tests for removing and adding bsconfig.json files dynamically to a workspace
- [x] unit test for honoring vscode `files.exclude` setting


src/DiagnosticCollection.spec.ts
```diff
@@ -9,7 +9,7 @@ import { expect } from 'chai';
 describe('DiagnosticCollection', () => {
     let collection: DiagnosticCollection;
     let diagnostics: BsDiagnostic[];
-    let workspaces: Workspace[];
+    let workspaces: Project[];
```
iObject, line 12: Should you update workspaces to projects?

TwitchBronBron: Oops. Yep, good catch.

src/LanguageServer.spec.ts
```diff
@@ -129,19 +137,19 @@ describe('LanguageServer', () => {
...
     }
 
-    describe('createStandaloneFileWorkspace', () => {
+    describe('createStandaloneFileProject', () => {
         it('never returns undefined', async () => {
             let filePath = `${rootDir}/.tmp/main.brs`;
             writeToFs(filePath, `sub main(): return: end sub`);
-            let firstWorkspace = await server['createStandaloneFileWorkspace'](filePath);
-            let secondWorkspace = await server['createStandaloneFileWorkspace'](filePath);
+            let firstWorkspace = await server['createStandaloneFileProject'](filePath);
```
iObject, line 144: Should you update workspace styled variables to project?  Such as now `firstProject`, etc?

TwitchBronBron: Resolved

iObject (approved):

livecano (approved):
Looks good to me!

PR #617, opened 2022-06-08
Function generics

TwitchBronBron:
Basline work for generics in function declarations. 

This is a work-in-progress. I wanted to open the PR so I don't forget to finish this work eventually...


TwitchBronBron:
This is very out of date, and is targeting the wrong v1 version. Abandoning this, we will hopefully implement this feature some time in the future when v1 lands.

PR #618, opened 2022-06-08
Brs ast token props

TwitchBronBron:
Toying with the idea to move all AST tokens into a `.tokens` property. This makes token access more consistent. But with the eventual move to a red-green tree (see [this article](https://rome.tools/blog/2021/09/21/rome-will-be-rewritten-in-rust#whats-the-deal-with-abstract-syntax-trees) for more information), perhaps this isn't necessary? I opened this so I won't forget about the work, we can evaluate whether we want to keep it or not at a later date.


TwitchBronBron:
This has been implemented in a slightly different way in #1025 as part of the v1 release (#1053)

PR #621, opened 2022-06-09
Better enum/class/namespace semantic tokens

TwitchBronBron:
Adds much better syntax highlighting (via semantic tokens) for enums, classes, and namespaces.
**Before:**
image

**After:**
image


livecano (approved):
Looks great!

PR #639, opened 2022-07-14
Red/Green Tree - Parser rewrite (Draft)

TwitchBronBron:
Rewrites the parser to use the red-green tree pattern. This produces a concrete syntax tree (CST) that can fully reproduce the source code. 

This is a work-in-progress, and is nowhere near complete.


TwitchBronBron:
We won't be doing this anytime soon, and probably won't implement it this away anyway.

PR #646, opened 2022-07-21
Language feature: named arguments in function calls

TwitchBronBron:
Initial proposal for named arguments in function calls.

### DIFF
--- a/docs/named-arguments.md
+++ b/docs/named-arguments.md
@@ -1,0 +1,160 @@
+# Named arguments
+Named arguments allow you to match an argument to its position in the function parameter list by name.
+
+**NOTE:** Named arguments are not currently supported for function calls within complex expressions such as ternary, null coalescence, etc. This may change in the future as the compiler improves.
+
+## Simple usage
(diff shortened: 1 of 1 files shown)


livecano (commented):
Is it possible to omit a name if maybe prefix with an underscore?

neilsb:
Sorry for the delay in looking at this - was away on vacation.

This all looks great and seems comprehensive.  Personally not keen "Named arguments after ordered parameters", as for clarity I would argue they should either all be named, or none (all be left in the default order) - however may well not be everyone's thinking. :)

TwitchBronBron:
> Sorry for the delay in looking at this - was away on vacation.
> 
> This all looks great and seems comprehensive. Personally not keen "Named arguments after ordered parameters", as for clarity I would argue they should either all be named, or none (all be left in the default order) - however may well not be everyone's thinking. :)

I think most of that could be enforced by a linter rule..I modeled a lot of this design off of the C# docs. 

https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/named-and-optional-arguments

PR #656, opened 2022-08-02
Scope validation performance boost

TwitchBronBron:
Several fixes:
- leverage child-parent linking introduced in #650 to speed up scope validation
- optimize scope validation by eliminating duplicate file checks when possible

While this hasn't restored the original performance lost in 0.53.0, it's a step in the right direction.
image

Breaking changes:
- removed `NamespaceStatement.symbolTable` in favor of `NamespaceStatement.getSymbolTable()` because the symbol table for namespaces actually reside within the `NamespaceStatement.body` property. Since the symbolTable logic was only recently added, it's unlikely that many projects are referencing it. 
- removed the `namespaceName` from several AstNodes, in favor of external consumers needing to call `.findAncestor(isNamespaceStatement)` instead. Note, this relies heavily on the `AstNode.parent` property which is set at the start of `onFileValidate` for each file. This is breaking because previously, the .namespaceName property was available immediately in `afterFileParse`. However, that logic was unsustainable for plugins that manipulate AST, and it's unlikely that many plugins are leveraging that property, so it is worth the breaking change. 

  Here are the nodes with the removed property:
  -  `VariableExpression`
  - `DottedGetExpression`
  - `CallExpression`
  - `FunctionStatement`


TwitchBronBron:
Some of the fixes have further improved performance. Still not as good as 0.52.3, but better than when this PR was originally submitted.
image

src/bscPlugin/validation/BrsFileValidator.ts
```diff
@@ -58,21 +60,99 @@ export class BrsFileValidator {
...
+            }
+        });
+
         this.event.file.ast.walk((node, parent) => {
             // link every child with its parent
             this.setParent(node, parent);
-
-            //do some file-based validations
-            if (isEnumStatement(node)) {
-                this.validateEnumDeclaration(node);
```
TwitchBronBron, line 74: This was moved into the EnumStatement callback above.

src/bscPlugin/validation/ScopeValidator.ts
```diff
@@ -27,181 +27,133 @@ const platformComponentNames = new Set(Object.values(components).map(x => x.name
...
 
-    private events: OnScopeValidateEvent[] = [];
+    /**
+     * The event currently being processed. This will change multiple times throughout the lifetime of this validator
+     */
+    private event: OnScopeValidateEvent;
 
     public processEvent(event: OnScopeValidateEvent) {
-        this.events.push(event);
-        event.scope.linkSymbolTable();
```
TwitchBronBron, line 34: Symbol table linking/unlinking happens at the program event level, we don't need to do it here anymore.

src/bscPlugin/validation/ScopeValidator.ts
```diff
@@ -27,181 +27,133 @@ const platformComponentNames = new Set(Object.values(components).map(x => x.name
...
-    private onceCache = new Cache<string, boolean>();
 
-    private addDiagnostic(event: OnScopeValidateEvent, diagnostic: BsDiagnostic) {
-        event.scope.addDiagnostics([diagnostic]);
-    }
+    private expressionsByFile = new Cache<BrsFile, Readonly<ExpressionInfo>[]>();
+    private iterateFileExpressions(file: BrsFile) {
+        const { scope } = this.event;
+        //build an expression collection ONCE per file
+        const expressionInfos = this.expressionsByFile.getOrAdd(file, () => {
```
TwitchBronBron, line 60: This is the core of the performance improvement. The expression info is computed _once_ for each file (previously this would happen once per scope per file).

PR #659, opened 2022-08-04
Fix typescript error for ast parent setting

TwitchBronBron:
Fixes ts error introduced in #650 introduced that only shows up in the typedef files. 

Seems to provide a _slight_ performance boost as well (probably due to the `if (childNode?.parent)` check which prevents duplicate reparenting. 
image

### DIFF
--- a/src/parser/Expression.ts
+++ b/src/parser/Expression.ts
@@ -348,18 +348,6 @@ export class NamespacedVariableNameExpression extends Expression {
     }
     range: Range;
 
-    // @ts-expect-error override the property
-    public get parent() {
-        return this._parent;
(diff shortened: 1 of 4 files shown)


src/parser/Expression.ts
```diff
@@ -348,18 +348,6 @@ export class NamespacedVariableNameExpression extends Expression {
     }
     range: Range;
 
-    // @ts-expect-error override the property
```
TwitchBronBron, line 351: This comment isn't persisted into the d.ts file, so it causes errors in consumers.

chrisdp (approved):

PR #664, opened 2022-08-15
Dedupe code completions in components

### LINKED ISSUE #662: Callfunc operator completions shows duplicate entries
Brighterscript shows some "slightly helpful" completions after the callfunc operator. However, those values aren't deduped. We should only show distinct method names.

image

This can be fixed by tweaking the code [here](https://github.com/rokucommunity/brighterscript/blob/7ef24cf89a11ab3a659e106d27874040206d0e49/src/bscPlugin/completions/CompletionsProcessor.ts#L22) to build the callfunc completions into its own array, deduping them, and _then_ adding them to the final `event.completions` array.

Add another unit test like [this one](https://github.com/rokucommunity/brighterscript/blob/7ef24cf89a11ab3a659e106d27874040206d0e49/src/Program.spec.ts#L1236) that has multiple functions with the same name across multiple components, and make sure only a single completion for each name is returned.

iObject:
Fixes #662

### DIFF
--- a/src/Program.spec.ts
+++ b/src/Program.spec.ts
@@ -1276,6 +1276,40 @@ describe('Program', () => {
         ).to.eql(['sayHello', 'sayHello2', 'sayHello3', 'sayHello4']);
     });
 
+    it('gets completions for callfunc invocation with multiple nodes and validates single code completion results', () => {
+        program.setFile('source/main.bs', `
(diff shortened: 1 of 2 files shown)


TwitchBronBron (approved):
Looks great, thanks!

Issue #665, opened 2022-08-17
distribute stand alone lsp


ZeeD:
Maybe I misunderstood, but I see that, while this project offer an implementation for lsp, this works only in vscode.
I would prefer to continue to use my preferred IDE.
My understanding is that other projects offers a stand alone executable that communicate back and forth with the client via stdin / stdout.
Would it be possible to adapt the lsp implementation to run outside vscode?

ZeeD:
I was "poking" at the project, and - inspired by https://github.com/rokucommunity/vscode-brightscript-language/blob/159fd098918d62a25591b4e55ea228c5ff40f0ab/src/LanguageServerRunner.ts#L15-L17 I tough I could have a dummy wrapper of https://github.com/rokucommunity/brighterscript/blob/1bfe31dae33f56518cf64e389847de2df8f4c56f/src/LanguageServer.ts , like

```javascript
#!/usr/bin/env node

const { LanguageServer } = require('brighterscript');
const server = new LanguageServer();
server.run();
```

and run it with `--stdio` flag, but it seems that, while the server starts, and wait for input, if I try to "attach" a client the server terminate unexpectedly.

is this the "right" direction? it's possible to use the connection with `--stdio` or the `LanguageServer` have some assumptions (it seems to me that the vscode plugin relies on `--node-ipc` to communicate)?

TwitchBronBron:
I haven't played with integrating the language server into any other IDEs, but the whole point of bundling the language server into brighterscript was for this exact purpose (to allow the brighterscript project to be THE BeightScript/BrighterScript language server for all other IDEs). 

I've tried very hard to keep the language server built according to the spec and not depend on vscode. But obviously I've never tried it in another IDE to see if it actually works.  

What IDE are you trying to integrate with?

ZeeD:
my preferred IDE is eclipse, that support LSP, when I use [Wild Web Developer](https://marketplace.eclipse.org/content/wild-web-developer-html-css-javascript-typescript-nodejs-angular-json-yaml-kubernetes-xml), and I also tried kate, that has an [LSP Client Plugin](https://docs.kde.org/trunk5/en/kate/kate/kate-application-plugin-lspclient.html)

is there a "message" that I can send to the language server to see if it works as expected? Or at least a way to have some "logs" to do some troubleshouting myself?

TwitchBronBron:
[...]

PR #667, opened 2022-08-17
Fix crashes for clients that don't support "workspace/configuration" requests

TwitchBronBron:
Prevent several crashes related to the language server incorrectly assuming that all clients support "workspace/configuration".

### DIFF
--- a/src/LanguageServer.spec.ts
+++ b/src/LanguageServer.spec.ts
@@ -1003,6 +1003,24 @@ describe('LanguageServer', () => {
         });
     });
 
+    describe('getConfigFilePath', () => {
+        it('honors the hasConfigurationCapability setting', async () => {
+            server.run();
+            sinon.stub(server['connection'].workspace, 'getConfiguration').returns(
+                Promise.reject(
+                    new Error('Client does not support "workspace/configuration"')
+                )
+            );
+            server['hasConfigurationCapability'] = false;
+            fsExtra.outputFileSync(`${workspacePath}/bsconfig.json`, '{}');
+            expect(
+                await server['getConfigFilePath'](workspacePath)
+            ).to.eql(
+                s`${workspacePath}/bsconfig.json`
+            );
+        });
+    });
+
     describe('getWorkspaceExcludeGlobs', () => {
         it('honors the hasConfigurationCapability setting', async () => {
             server.run();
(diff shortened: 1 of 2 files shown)


src/LanguageServer.spec.ts
```diff
@@ -388,29 +389,6 @@ describe('LanguageServer', () => {
...
-
-            //no child bsconfig.json files, use the workspacePath
-            expect(
-                server.projects.map(x => x.projectPath)
-            ).to.eql([
-                workspacePath
-            ]);
-        });
-
-        it('ignores bsconfig.json files from vscode ignored paths', async () => {
```
TwitchBronBron, line 402: Deleted this totally duplicate unit test

ZeeD:
I know this is a bit unrelated to the PR, but do you think you can add another binary that expose the language server as a standalone application?
Or do you prefer to keep it outside the project?

TwitchBronBron:
> I know this is a bit unrelated to the PR, but do you think you can add another binary that expose the language server as a standalone application? Or do you prefer to keep it outside the project?

Let's discuss this back in #665

chrisdp (approved):

src/LanguageServer.ts
```diff
@@ -230,14 +230,18 @@ export class LanguageServer {
...
-        let config = await this.connection.workspace.getConfiguration({
-            scopeUri: workspaceFolder,
-            section: 'files'
-        }) as {
-            exclude: Record<string, boolean>;
+    private async getWorkspaceExcludeGlobs(workspaceFolder: string): Promise<string[]> {
+        let config = {
+            exclude: {} as Record<string, boolean>
         };
+        //if supported, ask vscode for the `files.exclude` configuration
```
KamasamaK, line 237: Shouldn't this comment use the generic "client" instead of specifically mentioning "vscode"?

TwitchBronBron: The `files` and `files.exclude` are specific to VSCode. I have no idea what other editors to do handle including/excluding files from the workspace, but they're _probably_ doing it differently than this. So, the comment still makes sense, and we can enhance the comment in the future as we add support for other editors.

KamasamaK: That's true that the specific configuration options are not part of the standard. Although there are other clients such as Theia that model their settings after VS Code.

PR #668, opened 2022-08-17
add "--lsp" flag to bsc to start an LSP server

ZeeD:
I'm not sure this is the best approach. Please suggest a better integration - my real doubt is that almost no flow is shared between the LSP and the °normal° bsc functionality

### DIFF
--- a/src/BsConfig.ts
+++ b/src/BsConfig.ts
@@ -158,10 +158,4 @@ export interface BsConfig {
      * @default false
      */
     allowBrighterScriptInBrightScript?: boolean;
-
-    /**
-     * Start a standalone LSP server
-     * @default false
-     */
-    lsp?: boolean;
 }
--- a/src/BsConfig.ts
+++ b/src/BsConfig.ts
@@ -158,4 +158,10 @@ export interface BsConfig {
      * @default false
      */
     allowBrighterScriptInBrightScript?: boolean;
+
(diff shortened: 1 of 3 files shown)


TwitchBronBron:
Yeah, this is a bit more involved than I had intended. The primary reason for wanting to use something like `--lsp` was so we could keep the same entry point instead of exposing a whole separate binary name (like `npx bsc-lsp`). You can run programs without even installing them by using their full name. Like this: `npx brighterscript --lsp` it will download the package right then, and then spin it up and use it. 

In the CLI, we have full control over how we run the code, so I was envisioning something more like this:

```typescript
//cli.ts
let options = yargs
    // ...
   .option('lsp', { type: 'boolean', defaultDescription: 'false', description: 'Run brighterscript as a language server.' })
   .argv;

if (options.lsp === true) {
     const server = new LanguageServer();
     server.run();
} else{
    let builder = new ProgramBuilder();
    // ...


```

ZeeD:
I changed the PR following your suggestions

TwitchBronBron:
Looks great! Did you have a chance to test this against your Eclipse implementation to verify that it works?

ZeeD:
yes, it works as expected!

TwitchBronBron:
This change has been included in release v0.56.0

PR #675, opened 2022-08-22
Add npm caching to speed up workflow times

TwitchBronBron:
Hopefully we'll get a significant time reduction for our CI jobs by using the built-in npm caching for @setup-node


TwitchBronBron:
Honestly, this didn't improve performance that much. Maybe 1-3 seconds per build. Not worth the potential sync issues with npm caching.

PR #682, opened 2022-08-31
Add class method binding docs

TwitchBronBron:
Add section to class docs explaining how function instance binding works.


ZeeD (approved):
LGTM

PR #685, opened 2022-09-06
fix(parser): consider namespace function transpiled names

arturocuya:


TwitchBronBron (approved):
Looks great! Thanks for this

PR #687, opened 2022-09-12
Merge master into v1

TwitchBronBron:
Merge master into v1.


src/SymbolTable.ts
```diff
@@ -9,7 +9,8 @@ import { DynamicType } from './types/DynamicType';
  */
 export class SymbolTable {
     constructor(
-        parent?: SymbolTable | undefined
+        parent?: SymbolTable | undefined,
+        public identifier: string = ''
```
TwitchBronBron, line 13: Let's call this `name` instead, so it's less likely to be confused with the `Identifier` token types. Also, the type is implied by the default value, so no need for the explicit type.
```suggestion
        public name = ''
```

src/SymbolTable.ts
```diff
@@ -51,7 +52,9 @@ export class SymbolTable {
     private parentStack: SymbolTable[] = [];
 
     public pushParent(parent?: SymbolTable) {
-        this.parentStack.unshift(parent);
+        if (parent) {
+            this.parentStack.unshift(parent);
+        }
```
TwitchBronBron, line 57: Remove this. Although pushing undefined parents is not good, the calling code might plan on calling pop later, so we need to keep the undefined in the list so we can pop it.

src/bscPlugin/validation/ScopeValidator.ts
```diff
@@ -351,4 +360,108 @@ export class ScopeValidator {
...
+            diagnostics.push(...this.diagnosticDetectInvalidFunctionCalls(event.scope, file, callableContainerMap));
+        });
+        event.scope.addDiagnostics(diagnostics);
+    }
+
+
+    /**
+    * Find functions with either the wrong type of parameters, or the wrong number of parameters
+    */
+    private diagnosticDetectInvalidFunctionCalls(scope: Scope, file: BscFile, callableContainersByLowerName: CallableContainerMap) {
```
TwitchBronBron, line 379: Rename this function. The word `diagnostic` is a holdover from the scope file. Maybe `detectInvalidFunctionCallsForFile`?

src/files/BrsFile.ts
```diff
@@ -183,13 +183,15 @@ export class BrsFile {
...
+    public getClosestExpression(position: Position | Range) {
         const handle = new CancellationTokenSource();
         let containingNode: Expression | Statement;
         this.ast.walk((node) => {
             const latestContainer = containingNode;
             //bsc walks depth-first
-            if (util.rangeContains(node.range, position)) {
+            if (isPosition(position) && util.rangeContains(node.range, position)) {
+                containingNode = node;
+            } else if (isRange(position) && util.rangeContainsRange(node.range, position)) {
```
TwitchBronBron: This seems a bit wasteful to run the `isPosition` and `isRange` call for every ast walk iteration. Consider caching the values at the top of the function?

```typescript
const paramIsPosition = isPosition(position);
const paramIsRange = isRange(position);
```

src/files/BrsFile.ts
```diff
@@ -863,13 +865,12 @@ export class BrsFile {
...
     }
 
-
     private findNamespaceFromTokenChain(originalTokenChain: TokenChainMember[], scope: Scope): NamespacedTokenChain {
         let namespaceTokens: Token[] = [];
         let startsWithNamespace = '';
         let namespaceContainer: NamespaceContainer;
         let tokenChain = [...originalTokenChain];
-        while (tokenChain[0] && tokenChain[0].usage === TokenUsage.Direct) {
+        while (tokenChain[0] && (tokenChain[0].usage === TokenUsage.Direct)) {
```
TwitchBronBron, line 871: Are the added parens necessary?

src/parser/Statement.ts
```diff
@@ -1141,7 +1141,7 @@ export class NamespaceStatement extends Statement implements TypedefProvider {
     ) {
         super();
         this.name = this.nameExpression.getName(ParseMode.BrighterScript);
-        this.symbolTable = new SymbolTable(parentSymbolTable);
+        this.symbolTable = new SymbolTable(parentSymbolTable, `Namespace ${keyword.text}`);
```
TwitchBronBron, line 1144: `keyword.text` would contain "namespace". Did you mean `nameExpression.getName()`?

src/parser/Statement.ts
```diff
@@ -1548,6 +1559,8 @@ export class ClassStatement extends Statement implements TypedefProvider, Member
                 this.memberMap[statement?.name?.text.toLowerCase()] = statement;
             }
         }
+        this.symbolTable.identifier = `Class (symbols) ${name?.text}`;
+        this.memberTable.identifier = `Class (members) ${name?.text}`;
```
TwitchBronBron, line 1563: Put the `(symbols)` and `(members)` at the end. 
```suggestion
        this.symbolTable.identifier = `Class ${name?.text} (symbols) `;
        this.memberTable.identifier = `Class ${name?.text} (members)`;
```

src/files/BrsFile.ts
markwpearce: just removed the whole thing here... not needed

src/files/BrsFile.ts
markwpearce: fixed

src/SymbolTable.ts
markwpearce: changed

PR #696, opened 2022-09-23
Finds and includes more deeply embedded expressions

markwpearce:
Does not give up so easily when an expression can't be parsed - will continue looking at further expressions.

Examples:

- `DottedGetExpressions` ending in a `.` -> first part is parsed as an expression and added to the AST
- `CallExpressions` -> if there is no closing `)`, or if an argument is unable to be parsed, the call expression is still added to the AST
- `ArrayLiteralExpression`  if there is no closing `]`, or if an element is unable to be parsed, the ArrayLiteralExpression is still added to the AST
- etc. (see tests!)

This has the nice side effect of provide WAY more diagnostics - so diagnostics like "Unmatched Bracket" actually show up, instead of "Unexpected newline", as well as being able to provide diagnostics on multiple problem arguments in the same function call.

### DIFF
--- a/src/files/BrsFile.spec.ts
+++ b/src/files/BrsFile.spec.ts
@@ -1321,6 +1321,13 @@ describe('BrsFile', () => {
                     DoC(
                 end function
             `);
+            expectDiagnostics(file.parser.diagnostics, [
+                DiagnosticMessages.expectedRightParenAfterFunctionCallArguments(),
+                DiagnosticMessages.expectedNewlineOrColon(),
+                DiagnosticMessages.unexpectedToken('end function'),
+                DiagnosticMessages.expectedRightParenAfterFunctionCallArguments(),
+                DiagnosticMessages.expectedNewlineOrColon()
+            ]);
             expect(file.functionCalls.length).to.equal(2);
 
             expect(file.functionCalls[0].range).to.eql(Range.create(2, 20, 2, 27));
--- a/src/files/BrsFile.spec.ts
+++ b/src/files/BrsFile.spec.ts
@@ -1311,6 +1311,25 @@ describe('BrsFile', () => {
             expect(file.functionCalls[1].nameRange).to.eql(Range.create(5, 20, 5, 23));
         });
 
+        it('finds function calls that are unfinished', () => {
+            let file = new BrsFile('absolute_path/file.brs', 'relative_path/file.brs', program);
+            file.parse(`
+                function DoA()
+                    DoB("a"
+                end function
+                function DoB(a as string)
+                    DoC(
+                end function
+            `);
+            expect(file.functionCalls.length).to.equal(2);
+
+            expect(file.functionCalls[0].range).to.eql(Range.create(2, 20, 2, 27));
+            expect(file.functionCalls[0].nameRange).to.eql(Range.create(2, 20, 2, 23));
+
+            expect(file.functionCalls[1].range).to.eql(Range.create(5, 20, 5, 24));
+            expect(file.functionCalls[1].nameRange).to.eql(Range.create(5, 20, 5, 23));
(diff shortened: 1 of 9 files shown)


src/files/BrsFile.ts
```diff
@@ -692,7 +692,7 @@ export class BrsFile {
                     }
                 }
                 let functionCall: FunctionCall = {
-                    range: util.createRangeFromPositions(expression.range.start, expression.closingParen.range.end),
+                    range: expression.range, //util.(expression.range.start, expression.closingParen.range.end),
```
TwitchBronBron:
```suggestion
                    range: expression.range,
```

src/parser/Expression.ts
```diff
@@ -429,8 +431,8 @@ export class IndexedGetExpression extends Expression {
             ...this.obj.transpile(state),
             this.questionDotToken ? state.transpileToken(this.questionDotToken) : '',
             state.transpileToken(this.openingSquare),
-            ...this.index.transpile(state),
-            state.transpileToken(this.closingSquare)
+            ...(this.index ? this.index.transpile(state) : []),
```
TwitchBronBron, line 434: This can be slightly simplified:
```suggestion
            ...(this.index?.transpile(state) ?? []),
```

src/files/BrsFile.spec.ts
```diff
@@ -1311,6 +1311,25 @@ describe('BrsFile', () => {
...
+            let file = new BrsFile('absolute_path/file.brs', 'relative_path/file.brs', program);
+            file.parse(`
+                function DoA()
+                    DoB("a"
+                end function
+                function DoB(a as string)
+                    DoC(
+                end function
+            `);
+            expect(file.functionCalls.length).to.equal(2);
```
TwitchBronBron, line 1324: Can you add a `expectDiagnostics()` call here with all the diagnostics you think this should produce?

src/parser/tests/expression/AssociativeArrayLiterals.spec.ts
```diff
@@ -205,6 +207,68 @@ describe('parser associative array literals', () => {
         ]);
     });
 
+    describe('unfinished', () => {
+        it('will still be parsed', () => {
+            // No closing brace:  _ = {name: "john", age: 42, address: data.address
+            let { statements, diagnostics } = Parser.parse([
```
TwitchBronBron, line 213: These verbose tests are just a holdover from the `brs` project. You can convert this test into a standard template string instead of building the whole array of tokens.

src/parser/Parser.ts
```diff
@@ -2911,6 +2936,19 @@ export class Parser {
         return this.tokens[this.current - 1];
     }
 
+    /**
+     * There are sometimes we get catch an error that is a diagnostic
+     * If that's the case, we want to continue parsing
+     * Otherwise, re-throw the error
```
TwitchBronBron, line 2942:
```suggestion
     * Sometimes we catch an error that is a diagnostic.
     * If that's the case, we want to continue parsing.
     * Otherwise, re-throw the error
```

TwitchBronBron (changes requested):
Great PR! I appreciate that it's hyper-focused, and this will be a fantastic improvement to the parser. Just a few suggestions, and then this is good to go.

src/parser/tests/expression/AssociativeArrayLiterals.spec.ts
markwpearce: fixed

src/files/BrsFile.spec.ts
markwpearce: done

TwitchBronBron (approved):

PR #697, opened 2022-09-27
Syntax and transpile support for continue statement

TwitchBronBron:
Roku OS 11.5 introduced the `continue` statement. This PR adds the necessary parsing and transpiling to support using that feature as defined by Roku. 

This does not address backwards-compatibile transpiling (i.e. transpile to labels and `goto` expressions). That effort is still ongoing in #489 and will be addressed at a future time.

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,6 +63,9 @@ BrighterScript adds several new features to the BrightScript language such as Na
 ## Who uses Brighterscript?
 <br/>
 <p align="center">
+    <a href="https://www.fubo.tv/">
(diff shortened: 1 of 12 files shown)


arturocuya:
Does the new `Continue` statement support plugin annotations?

TwitchBronBron:
@arturocuya do you mean like this? If so, then no. I believe annotations are currently only applicable to the higher-level things like namespaces and functions. 
```bs
while true 
    @YourCustomAnnotation
    continue while
end while
```

TwitchBronBron:
Oh, well if annotations work with `throw`, then they should also work for `continue`, as both are "statements".

PR #698, opened 2022-09-27
Fix if statement block var bug

TwitchBronBron:
Prevent variable shadowing in blocks.


TwitchBronBron:
I ran this against one of our larger projects, and it has all sorts of errors, even though the bsc unit tests are all passing. I need to investigate this a bit more.

xgouchet (approved):

TwitchBronBron:
The issues I was encountering in the other project were not caused by this PR, so this is good to merge.

PR #707, opened 2022-10-02
fixes signature help resolution for callexpressions

georgejecook:
This pr addresses the following issues:

 - Cannot bring up signature help for nested call expressions
 - Cannot bring up signature help for multi-lined functions
 - Cannot bring up signature help for function calls with parenthesis in the arguments.

This implementation uses the new AST searchign, and parent offerings; plus some of the work done to imrpove AST fail-over when a parsing failure occurs (e.g. for presently-edited-lines, which may have invalid parenthesis/bracket pairing).

I took the approach of consolidating code that was in Program and in BrsFile, into SignatureHlepInfo, which is conceptually part of the bscPlugin.

I also added the CallExpressionInfo class, to allow us to process and store info about a call expressions' state and type, which is useful for signature resolution, and, in my next few contributions, improvements to completions, and go to definition, too.

Testing is exhaustive, having covered the types:
 - call expression
 - call func expression
 - namespaced call expression
 - class method call expression
 - class method on my own class (i.e. my m)
 - constructor call expression
 - mutliline
 - commas
 - white spaces
 - undefined/error scenarios.

However, due to having written this on program originally, then extracting to the utility classes, the tests are more _integration_ tests in their nature.

I have also shared video of how the feature works in the IDE, which clearly demonstrates the great improvements herein.


georgejecook:
Added @triwav because I know he worked on some of this before :)

georgejecook:
note - 2 tests are failing due to sourcemap errors. reached out to @TwitchBronBron about that.

```
     Error: You must provide the URL of lib/mappings.wasm by calling SourceMapConsumer.initialize({ 'lib/mappings.wasm': ... }) before using SourceMapConsumer
      at readWasm (/home/georgejecook/hope/open-source/vsc/brighterscript/node_modules/source-map/lib/read-wasm.js:8:13)
      at wasm (/home/georgejecook/hope/open-source/vsc/brighterscript/node_modules/source-map/lib/wasm.js:25:16)
      at /home/georgejecook/hope/open-source/vsc/brighterscript/node_modules/source-map/lib/source-map-consumer.js:264:14
      at processTicksAndRejections (node:internal/process/task_queues:95:5)
      at Context.<anonymous> (/home/georgejecook/hope/open-source/vsc/brighterscript/src/files/BrsFile.spec.ts:2507:28)

```

triwav (approved):
Only gave it a quick review but looks good from what I could see. Thanks @georgejecook

src/bscPlugin/SignatureHelpUtil.ts
```diff
@@ -0,0 +1,155 @@
...
+import { ParseMode } from '../parser/Parser';
+import type { ClassStatement } from '../parser/Statement';
+import type { SignatureInfoObj } from '../Program';
+import type { XmlScope } from '../XmlScope';
+import type { CallExpressionInfo } from './CallExpressionInfo';
+import { CallExpressionType } from './CallExpressionInfo';
+import type { FileLink } from '../interfaces';
+import { util } from '../util';
+
+export class SignatureHelpUtil {
```
markwpearce, line 14: Should this really be a class? It looks like three basically independent functions....

src/bscPlugin/CallExpressionInfo.ts
```diff
@@ -0,0 +1,161 @@
...
+    call = 'call',
+    callFunc = 'callFunc',
+    constructorCall = 'constructorCall',
+    myClassCall = 'myClassCall',
+    otherClassCall = 'otherClassCall',
+    unknown = 'unknown'
+}
+
+//Util Class, for extracting info about an expression, which can be used in the IDE
+export class CallExpressionInfo {
```
markwpearce, line 22: Everything else in the `bscPlugin` folder is event driven --- should signature help be event driven too?

markwpearce (changes requested):

src/parser/Expression.ts
```diff
@@ -377,6 +377,21 @@ export class DottedGetExpression extends Expression {
             walk(this, 'obj', visitor, options);
         }
     }
+
+    getNameParts() {
```
TwitchBronBron, line 381: This function incorrectly assumes that all `DottedGetExpression` instances represent only dotted gets the whole way through. That's not how `DottedGetExpression` works. 

There's also already a util function that handles this logic: https://github.com/rokucommunity/brighterscript/blob/bc7367a570be3183db7c5889f526460e41de1a03/src/util.ts#L1377-L1381

The following expression contains `DottedGetExpression` instances mixed with `CallExpression`, `IndexedGetExpression`, and others. 
```
alpha.beta.charlie()["delta"]@.echo().foxtrot.golf
```

Here's a unit test proving this function explodes with the above expression.

```typescript
it('test getNameParts', () => {
    const file = program.setFile<BrsFile>('source/main.bs', `
        sub main()
            alpha = {}
            value = alpha.beta.charlie()["delta"]@.echo().foxtrot.golf
        end sub
    `);

    expect(
        file.ast.findChild<DottedGetExpression>(isDottedGetExpression).getNameParts()
    ).to.eql([]);
});
```

image

TwitchBronBron (changes requested):
Overall, this looks pretty good! I went ahead and refactored the unit tests to use proper indentation and eliminate the `beforeEach` usage in favor of dedicated code block for each test. 

Just one primary thing to address and one question and then I think this should be good to go.

src/bscPlugin/SignatureHelpUtil.ts
TwitchBronBron: I actually prefer the util class as implemented here. Our primary `Util.ts` file is also a class, so this is following that same pattern. Using a class helps improves the test experience since standalone functions are harder to mock in unit tests.

src/parser/Expression.ts
georgejecook: addressed. thanks

src/Program.spec.ts
```diff
@@ -2127,452 +2125,670 @@ describe('Program', () => {
...
+                    end namespace
+                `);
+                program.validate();
+                expectZeroDiagnostics(program);
+                assertSignatureHelp(2, 57, 'sub person.greeter.sayHello(name as string, age as integer)', 1);
+            });
         });
 
-        it('gets signature help for partially typed line', () => {
+        it.skip('gets signature help for partially typed line', () => {
```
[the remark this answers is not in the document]
georgejecook: fixed. Note - this implementation does not support param index > 1 (i.e. more then 0 commas), if there is no matching outer parentheses.

I think 
- the rest of the value here far outweighs this limitation
- for most folks, with auto paren insertion, won't be an issue
- this should be fixed in the lexer/somewhere else imo, to better fail, rather than put more of code in here.

georgejecook:
> Overall, this looks pretty good! I went ahead and refactored the unit tests to use proper indentation and eliminate the `beforeEach` usage in favor of dedicated code block for each test.

That was jolly decent of you - I really appreciate your help there. T h a n k s

TwitchBronBron (approved):

PR #708, opened 2022-10-02
Allow nested namespaces

### LINKED ISSUE #624: can we have nested namespaces please?
will be especially handy for enums..
so..

```

namespace nba.content
  enum LeagueIds
    nba = "00"
    summerLeague = "15"
  end enum

  enum GameUpdaterMode
    date = "date"
    game = "game"
  end enum
end namespace

namespace nba.content.tasks
  enum GamesUpdaterTaskCommand
    updateGamesOnDay = "updateGamesOnDay"
    updateGamesWithId = "updateGamesWithId"
  end enum
end namespace
```

would be nicer as

```

namespace nba.content
  enum LeagueIds
    nba = "00"
    summerLeague = "15"
  end enum

  enum GameUpdaterMode
    date = "date"
    game = "game"
  end enum
  
  namespace tasks
    enum GamesUpdaterTaskCommand
      updateGamesOnDay = "updateGamesOnDay"
      updateGamesWithId = "updateGamesWithId"
    end enum
  end namespace
end namespace
```

elliot-nelson:
### SUMMARY

Allow nested namespaces (fixes #624).

### DETAILS

_Disclaimer -  this is my first go-round in the codebase, so if I've everything I've done here is totally wrong, don't hesitate to say so! Any feedback appreciated._ 

Here's the problems I attempted to solve:

1. We need to unblock `namespace` statements, but, they aren't _generically_ allowed -- they can't be inside a function or a class, for example. To accomplish this I invented a new concept called a `ParserScope`, kept on a stack, that tries to express the stack of namespaces, classes, and functions below us. As long as the stack consists only of namespaces, it's OK to declare a namespace.

2. Once you have declared a namespace, you need a way to fully qualify the name to include your parent namespaces. The way I've chosen to do this is determine the set of parent name expressions _at parse time_ -- we have the stack, so we can grab the name expressions of all containing namespaces and hold onto them in the namespace statement, consuming them at the appropriate time (when we know whether it's brightscript or brighterscript mode).

3. Last, we need a diagnostic adjustment. I chose to leave the root-level diagnostic alone (there's no callers anymore, but there might be one in the future), and invent a new one for the "root- or namespace-level" requirement.

### POSSIBLE ALTERNATIVES

I am most hesitant about the code for problem (2). I actually _didn't need to do this_ to get it working; I could create the NamespaceStatement as normal, and have the NamespaceStatement recursively walk up the `findAncestor` tree at `getName()` time to construct the qualified name.

What I didn't like about that approach is that running the compiler on my sample program produced the right output, but the unit tests gave me a lot of grief -- I kept running the parser, examining the namespace statement references, and all those statements were divorced of any tree, so the names were wrong (unqualified), even though the code generated the right output. (Maybe this is expected and I'm testing the wrong thing here; if so, perhaps we could swap implementation of (2) to walk ancestors instead, since I see that pattern in a couple other places in `Statements.ts`).

### TESTING
 - Ran locally on a sample program with a few namespaces and methods, produced the expected output.
 - Added one new test, adjusted one expected test.
 - Adjusted one _unexpected_ test -- it seems like there was a syntax error in a test regarding completions, which I'm not sure was intentional or not, but had to remove to get the test passing.


elliot-nelson:
Update... as I think about this more, if the output of Parser is an AST then my approach is definitely wrong because you want to be able to clip and paste nodes around in the tree and have their output be correct.

So I think I only need advice on the unit test -- maybe the Parser test just confirms that Level3 is named "Level3", and there's a spot for a higher level unit test that confirms that program produces output function "Level1_Level2_Level3_main".

elsassph:
Hey @elliot-nelson, thanks for your contribution!

Yes from an AST point of view I'm not sure this is right. Maybe you could follow how classes build their qualified name (`getName`) by walking the AST using the `findAncestor` method.

TwitchBronBron:
I plan on fully reviewing this PR tomorrow when I'm back to work, but I agree with @elsassph . These need to follow the same pattern as classes (using .findAncestor).  This is because we need to support plugins that manipulate the AST, and so doing this in the parser is far too early in the process and would break AST modifications by plugins. 

The .parent property is set during onProgramValidate, so any unit tests will need to call program.validate() in before building/verifying the fully qualified name.

src/parser/Statement.ts
```diff
@@ -1113,7 +1115,13 @@ export class NamespaceStatement extends Statement implements TypedefProvider {
...
 
     public getName(parseMode: ParseMode) {
-        return this.nameExpression.getName(parseMode);
+        const fullyQualifiedNames = [
+            ...this.parentNameExpressions,
+            this.nameExpression
+        ].map(expression => expression.getName(parseMode));
+        const sep = parseMode === ParseMode.BrighterScript ? '.' : '_';
+
+        return fullyQualifiedNames.join(sep);
```
TwitchBronBron, line 1124: Refactor this to leverage `.findAncestor()`.  Keep in mind, `.parent` is not set until the `program.validate()` cycle, but that's acceptable since many other BrighterScript AST features also rely on `.parent` as well.

Something like this:

```typescript
const parentNamespaces: NamespaceStatement[] = [this];
let ancestor = this;
//build a list of namespace ancestors
while (ancestor) {
    ancestor = this.findAncestor<NamespaceStatement>(isNamespaceStatement);
    if(ancestor){
        parentNamespaces.push(ancestor);
    }
}
//somehow build the fully qualified name now
```

src/parser/Parser.ts
```diff
@@ -1313,17 +1328,24 @@ export class Parser {
...
+        const parentNamespaces: NamespaceStatement[] = [];
+
+        for (const scope of this.parserScopeStack) {
+            if (scope instanceof NamespaceScope) {
+                parentNamespaces.push(scope.namespaceStatement);
+            } else {
+                this.diagnostics.push({
+                    ...DiagnosticMessages.keywordMustBeDeclaredAtNamespaceLevel('namespace'),
+                    range: keyword.range
+                });
```
TwitchBronBron, line 1340: Move this validation check into the `BrsFileValidator` class, in the `NamespaceStatement visitor. (here-ish https://github.com/rokucommunity/brighterscript/blob/e8ae1d95e25298d2d629d9a73249968ab43b76fe/src/bscPlugin/validation/BrsFileValidator.ts#L103-L109)

The validation should walk up the .parent chain and add the diagnostic if a namespace has any parents that aren't also namespaces (or Body also I think).

elliot-nelson:
Thanks all for the feedback, based on comments I have:

 - Cut the `ParserScope` stuff
 - Updated `NamespaceStatement` to walk its ancestors to get the full dotted namespace name
 - The test for Parser just confirms you can create nested namespaces
 - The test for Statement now confirms that a NamespaceStatement retrieves its full dotted name, if in a program
 - The test for BrsFileValidator now confirms that a namespace can't be declared in the wrong place
 - Updated namespace docs to add the new allowed syntax

elliot-nelson:
Thanks!

 - Updated BrsFileValidator tests to use diagnostic expectations.

TwitchBronBron (approved):
Looks great! Thanks for this.

PR #709, opened 2022-10-03
Expose an 'isThrowStatement' utility function for plugins

xgouchet:
This PR adds a method `isThrowStatement`, similar to other methods in `reflection.ts` to let plugins easily know if a givent statement is a `throw`.


TwitchBronBron (approved):
Thanks, this looks good. Just need you to resolve the conflicts and then I can merge it.

PR #714, opened 2022-10-07
Copy information on files field from roku-deploy and restructure bsconfig section of README

josephjunker:
As mentioned in a conversation in the RokuCommunity slack, the authoritative documentation on the `files` field lives in the `roku-deploy` documentation. This PR copies this documentation into the README here.

The new documentation is much longer than would fit in a bulleted list, so I restructured the surrounding document to use a series of paragraphs rather than a list. As part of this restructuring I tweaked a bit of formatting and added clarification to points that confused me.

I considered leaving the bulleted list intact and pulling the `files` field's documentation into its own markdown file in the `docs` directory. Please let me know if you would prefer that I take that approach instead. Even if the reformatting I did is kept intact I think extracting an additional markdown file might still be reasonable.

### DIFF
--- a/README.md
+++ b/README.md
@@ -221,17 +221,19 @@ Type: `string`
 
 Relative or absolute path to another `bsconfig.json` file that this `bsconfig.json` file should import and then override. Prefix with a question mark (?) to prevent throwing an exception when the file does not exist. Defaults to `undefined`.
 
+Note: child config properties completely replace parent config properties. For example: if a parent and child bsconfigs both specify an array for `plugins`, or `files`, or `diagnosticFilters`, etc., then the parent's setting will be completely ignored and the child's setting will be used instead.
+
 #### `cwd`
 
 Type: `string`
 
-If present, overrides the current working directory when invoking `bsc`. Defaults to `undefined`.
+If present, overrides the current working directory when invoking `bsc`. Defaults to `process.cwd()`.
 
 #### `rootDir`
 
 Type: `string`
 
-The root directory of your roku project. Defaults to the directory which contains the `bsconfig.json` file.
+The root directory of your roku project. Defaults to `process.cwd()`.
 
 #### `stagingDir`
 
@@ -430,7 +432,7 @@ If `true`, the server will keep running and will watch and recompile on every fi
 
 Type: `boolean`
 
-If `true`, after a successful build, the project will be deployed to the Roku specified in host. Defaults to `false`. If this field is set to `true`, then the `host`, `username`, and `password` fields must be set as well.
+If `true`, after a successful build, the project will be deployed to the Roku specified in host. Defaults to `false`. If this field is set to `true`, then the `host` and `password` fields must be set as well.
 
 #### `host`
 
@@ -473,6 +475,8 @@ A list of filters used to hide diagnostics.
 
 Defaults to `undefined`.
 
+If a child bsconfig extends from a parent bsconfig, and both bsconfigs specify `diagnosticFilters`, the parent bsconfig's `diagnosticFilters` field will be completely overwritten.
+
 #### `diagnosticLevel`
 
 Type: `"hint" | "info" | "warn" | "error"`
@@ -504,6 +508,8 @@ If a child bsconfig extends from a parent bsconfig, and both bsconfigs specify a
 Type: `Array<string>`
 
 List of node scripts or npm modules to load during the startup sequence. Useful for running things like `ts-node/require`. Defaults to `undefined`.
+
+If a child bsconfig extends from a parent bsconfig, and both bsconfigs specify a `require` field, the child's `require` field will completely overwrite the parent's `require` field.
 
 #### `allowBrighterScriptInBrightScript`
 
--- a/README.md
+++ b/README.md
@@ -202,41 +202,262 @@ There are situations where you want to store some compiler settings in a config 
     });
     ```
 
-### bsconfig.json options
+### `bsconfig.json` options
 
 These are the options available in the `bsconfig.json` file.
 
- - **project**: `string` - A path to a project file. This is really only passed in from the command line or the NodeJS API, and should not be present in `bsconfig.json` files. Prefix with a question mark (?) to prevent throwing an exception when the file does not exist.
+#### `project`
 
- - **extends**: `string` - Relative or absolute path to another `bsconfig.json` file that this `bsconfig.json` file should import and then override. Prefix with a question mark (?) to prevent throwing an exception when the file does not exist.
+Type: `string`
 
- - **cwd**: `string` - Override the current working directory
+A path to a project file. This is really only passed in from the command line or the NodeJS API, and should not be present in `bsconfig.json` files. Prefix with a question mark (?) to prevent throwing an exception when the file does not exist.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -202,41 +202,262 @@ There are situations where you want to store some compiler settings in a config
...
+Relative or absolute path to another `bsconfig.json` file that this `bsconfig.json` file should import and then override. Prefix with a question mark (?) to prevent throwing an exception when the file does not exist. Defaults to `undefined`.
 
- - **files**: ` (string | string[] | { src: string | string[]; dest?: string })[]` - The list file globs used to find all files for the project. If using the {src;dest;} format, you can specify a different destination directory for the matched files in src.
+#### `cwd`
 
- - **outFile**: `string` -  The path (including filename) where the output file should be placed (defaults to `"./out/[WORKSPACE_FOLDER_NAME].zip"`).
+Type: `string`
 
- - **createPackage**: `boolean` - Creates a zip package. Defaults to true. This setting is ignored when `deploy` is enabled.
+If present, overrides the current working directory when invoking `bsc`. Defaults to `undefined`.
```
josephjunker, line 225: I was unclear on what this does based on the original documentation:
```
**cwd**: `string` - Override the current working directory
```
Please let me know if I've misphrased this.

README.md
```diff
@@ -202,41 +202,262 @@ There are situations where you want to store some compiler settings in a config
...
 
- - **watch**: `boolean` -  If true, the server will keep running and will watch and recompile on every file change.
+#### `rootDir`
 
- - **deploy**: `boolean` -  If true, after a successful build, the project will be deployed to the Roku specified in host.
+Type: `string`
 
- - **host**: `string` -  The host of the Roku that this project will deploy to.
+The root directory of your roku project. Defaults to the directory which contains the `bsconfig.json` file.
```
josephjunker, line 231: I believe the default of `rootDir` is where `bsconfig.json` lives, not the cwd when `bsc` is invoked.

README.md
```diff
@@ -202,41 +202,262 @@ There are situations where you want to store some compiler settings in a config
...
+
+Type: `boolean`
+
+If `true`, the server will keep running and will watch and recompile on every file change. Defaults to `false`.
+
+#### `deploy`
+
+Type: `boolean`
+
+If `true`, after a successful build, the project will be deployed to the Roku specified in host. Defaults to `false`. If this field is set to `true`, then the `host`, `username`, and `password` fields must be set as well.
```
josephjunker, line 430: I haven't tested whether `username` is required for `deploy` yet, so this is an assumption on my part.

README.md
```diff
@@ -246,18 +467,46 @@ These are the options available in the `bsconfig.json` file.
...
         }]
         ```
- - **diagnosticLevel**: `'hint' | 'info' | 'warn' | 'error'` - Specify what diagnostic levels are printed to the console. This has no effect on what diagnostics are reported in the LanguageServer. Defaults to 'warn'
 
- - **autoImportComponentScript**: `bool` - BrighterScript only: will automatically import a script at transpile-time for a component with the same name if it exists.
+Defaults to `undefined`.
+
+#### `diagnosticLevel`
+
+Type: `"hint" | "info" | "warn" | "error"`
```
josephjunker, line 475: Single quotes changed to double quotes because most of this section is using JSON

README.md
```diff
@@ -246,18 +467,46 @@ These are the options available in the `bsconfig.json` file.
...
+
+Specify what diagnostic levels are printed to the console. This has no effect on what diagnostics are reported in the LanguageServer. Defaults to `"warn"`.
+
+#### `autoImportComponentScript`
+
+Type: `bool`
+
+BrighterScript only: will automatically import a script at transpile-time for a component with the same name if it exists. Defaults to `false`.
```
josephjunker, line 483: I hope I am correct about the default value here.

README.md
```diff
@@ -246,18 +467,46 @@ These are the options available in the `bsconfig.json` file.
...
+
+Override the root directory path where debugger should locate the source files. The location will be embedded in the source map to help debuggers locate the original source files. This only applies to files found within `rootDir`. This is useful when you want to preprocess files before passing them to BrighterScript, and want a debugger to open the original files. This option also affects the `SOURCE_FILE_PATH` and `SOURCE_LOCATION` source literals.
+
+#### `plugins`
+
+Type: `Array<string>`
+
+List of node scripts or npm modules to load as plugins to the BrighterScript compiler. Defaults to `undefined`.
+
+If a child bsconfig extends from a parent bsconfig, and both bsconfigs specify a `plugins` field, the child's `plugins` field will completely overwrite the parent's `plugins` field.
```
josephjunker, line 497: I added this line because this confused me at first. I tested this but it wouldn't be a bad idea to confirm that I'm not missing something.

README.md
TwitchBronBron: It defaults to `process.cwd()`, not undefined. Otherwise your writeup looks correct.

README.md
TwitchBronBron: username is not required. Roku devices do not allow developers to _change_ the username, so it's always `rokudev`. Since that's the case, it's not required because our tools can default to `"rokudev"`.

README.md
TwitchBronBron: Sounds good. :)

README.md
TwitchBronBron: Yep, you are correct, it defaults to false.

README.md
TwitchBronBron: This is technically correct, but it's not limited to just `plugins`. We're doing a "replace every property in parent with properties found in child". so it would also apply to `files`, `diagnosticFilters`, etc... Might be worth explaining it a bit more generically, or making this a specific example. Perhaps something like this:

>Note: child config properties completely replace parent config properties. For example: ifa child bsconfig extends from a parent bsconfig, and both bsconfigs specify a `plugins` field, the child's `plugins` field will completely overwrite the parent's `plugins` field.

README.md
TwitchBronBron: So, it _should_ work this way, but currently it defaults to the `cwd`. I just tested and verified that I'm correct. I'd say leave it the way it's currently documented for now, and we can adjust the docs once it gets fixed in bsc.

TwitchBronBron (changes requested):
Just a few small changes and then this is good to go!

README.md
josephjunker: Thanks, updated in the new version.

README.md
josephjunker: Thanks, updated in the new version.

README.md
josephjunker: Thanks, updated in the new version.

README.md
josephjunker: I wanted to provide this information in general while also making it easy for people who are only looking at the documentation for the specific field they're stuck on, so in the new version I added the general explanation to the `extends` field and specifically called it out in the `files`, `plugins`, `diagnosticFilters`, and `require` fields. Its a lot of duplication but I'm hoping it will be more helpful to people who are quickly skimming.

TwitchBronBron (approved):

PR #725, opened 2022-10-21
Add 'severityOverride' option

### LINKED ISSUE #724: Allow overriding errors severity
Bsc, bslint and other plugins can emit diagnostics, sometimes with arbitrary severity (maybe too high or too low) and often without ways to configure them.

Some projects may be transitioning to bsc/bslint and suffer from too strict errors, while some established projects may want stricter errors.

Instead of having to chase each individual diagnostic producer to get them to offer configuration level a project needs, `bsconfig.json` could include a `severityOverride` option where one could adjust the severity to match their needs, e.g.

```
{
  "severityOverride": {
    "BSLINT1001": "warn", // we have too many "false positives" due to sloppy imports
    1004: "error" // maybe we want to fail the build on a hint
  }
}
```

elsassph:
`bsconfig.json` now accepts a `severityOverride` map allowing to adjust a diagnostic severity.

Fixes #724 - see issue for details


elliot-nelson:
Really interesting. Like tsc, bsc doesn't have an "eslint-esque" rules engine where there's a very clear way to turn on and off errors and give them different severity levels.

Instead, TS "rules" are governed by human-readable options that are kind of disjointed from the error message you get... i.e., when your build pops out `TS7006: foobar has implicit type 'any'`, that doesn't inform you that you could disable that by setting `compilerOptions.noImplicitAny: false`.

So, I like your approach, although I see two issues:

 - If you change _errors_ to _non-errors_, might you accidentally cause your build to succeed (exit code 0) even though your formatted output might be incorrect / non-runnable? Or are there still errors that are so bad the whole compiler just chokes up. (Maybe there's a "fatal" level that never reaches the diagnostic stage.)
 - The name `severityOverrides` seems very specific and almost a dirty hack, whereas I think this use case probably has a number of legit purposes. Would a structure more like `compilerOptions: { diagnostics: { 1004: "error" } }` seem less dirty?

elsassph:
Yes the whole point of this setting is to be a workaround - it should feel nasty!

Unlike eslint, plugins contributing to diagnostics can be heterogeneous so yes, this property is meant to be seen as a "workaround", which is useful when transitioning a project to bsc/bslint (as our colleagues are doing) where a lot of errors can be non-fatal in the current state of the codebase but need time to clean up.

src/BsConfig.ts
```diff
@@ -106,6 +106,11 @@ export interface BsConfig {
      */
     ignoreErrorCodes?: (number | string)[];
 
+    /**
+     * A map of error codes with their severity level override (error|warn|info)
+     */
+    severityOverride?: Record<number | string, 'error' | 'warn' | 'info' | 'hint'>;
```
TwitchBronBron, line 112: Let's call this `diagnosticSeverityOverrides` instead? Keeps it more in line with `diagnosticFilters`

PR #744, opened 2022-11-25
Let users use or discard explicit types

xgouchet:
Using the `as type` syntax for parameters and return type on function seems to have a performance impact on Roku channels. 

image (2)

This PR adds a `useExplicitType` configuration option to let the Brighterscript transpiler omit those to generate code that will be more efficient.


xgouchet:
> Thanks for this! Can you rename the property to `removeParameterTypes` instead?

Yes, and I had to invert the default value so that it didn't change the legacy behavior

TwitchBronBron (approved):

PR #757, opened 2022-12-06
Ensure enums and interfaces persist in typedefs

### LINKED ISSUE #756: please keep parameter types and return value in definitions when they are interfaces or enums
When `bsc` emit the definitions it seems that both `enum`s and `interface`s are missing.

starting from this `example.bs`

```brighterscript
interface Foo
    field as string
end interface

enum Bar
    value
end enum

function baz(parameter as Foo) as Bar
    return Bar.value
end function
```

the `example.d.bs` file generated is

```brighterscript
interface Foo
    field as string
end interface

enum Bar
    value
end enum
function baz(parameter as object) as object
end function
```

now `baz` has lost the type definition and return value information. 

I'm using
```bash
$ npx bsc --version
0.60.6
```

TwitchBronBron:
Typedefs should include enum and interface information in function parameter and function return types.

Fixes #756

### DIFF
--- a/src/files/BrsFile.spec.ts
+++ b/src/files/BrsFile.spec.ts
@@ -16,7 +16,7 @@ import { DiagnosticMessages } from '../DiagnosticMessages';
 import type { StandardizedFileEntry } from 'roku-deploy';
 import util, { standardizePath as s } from '../util';
 import PluginInterface from '../PluginInterface';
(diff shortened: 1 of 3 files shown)


TwitchBronBron:
This has been fixed in v0.61.1

PR #792, opened 2023-04-14
Adds ReferenceType

markwpearce:
Adds `ReferenceType`

A `ReferenceType` is a `BscType` that is given a `SymbolTableProvider`, so that it can do lookups in a symbol table.
This is useful for when a type is unknown at parse time - eg. when a function is defined in a different file (or even just further down a file)

An additional change here is to make sure SymbolTables are linked for the appropriate scope when doing a hover.

### DIFF
--- a/src/bscPlugin/hover/HoverProcessor.ts
+++ b/src/bscPlugin/hover/HoverProcessor.ts
@@ -42,87 +42,94 @@ export class HoverProcessor {
 
     private getBrsFileHover(file: BrsFile): Hover {
         const scope = this.event.scopes[0];
-        const fence = (code: string) => util.mdFence(code, 'brightscript');
-        //get the token at the position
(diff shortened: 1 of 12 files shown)


src/bscPlugin/hover/HoverProcessor.ts
```diff
@@ -42,87 +42,94 @@ export class HoverProcessor {
...
-            TokenKind.EndFunction,
-            TokenKind.Sub,
-            TokenKind.EndSub
-        ];
-
-        //throw out invalid tokens and the wrong kind of tokens
-        if (!token || !hoverTokenTypes.includes(token.kind)) {
-            return null;
-        }
+        try {
```
markwpearce, line 45: Moved this to a `try {}` so I could unlink the scope in a `finally{}`

TwitchBronBron (approved):

PR #794, opened 2023-04-19
Added members to `BscType`, and adds ReferenceType (future lookups)

markwpearce:
1. Refactors BscType so each type can have members

Members of a type are possible symbols that can be referenced after the 'dot' ... eg, a class's fields and methods

2. Adds `ReferenceType` (and `PropertyReferenceType`)

`ReferenceType` is a Proxy for a type that will be established later. eg. a Class that is defined later in the code (or in a different file), or a Namespace that hasn't been walked in the AST yet. `PropertyReferenceType` is used when we need to access `returnType` on a BscType, but we're not sure if the type has been defined yet.

3.  Changes the `getType()` method of many Statements and Expressions so that the result is more accurate.

4. Eg. On `DottedGetExpression.getType()`, the expression is "walked back" and the code takes into account the type's `getMemberType` method to find the correct type of code like `(new NameSpaceA.Klass()).method().property`

5. Adds `NamespaceType` and `EnumType` so `getMemberType` can be called on them.


src/types/BscType.ts
```diff
@@ -1,6 +1,45 @@
...
+
+    popMemberProvider() {
+        this.symbolTable.popParentProvider();
+    }
+
+    addMember(name: string, range: Range, type: BscType, flags: SymbolTypeFlags) {
+        this.symbolTable.addSymbol(name, range, type, flags);
+    }
+
+    getMemberType(name: string) {
```
TwitchBronBron, line 28: Can we just call this `getMember()`? `addMember` adds a type, so `getMember` make sense for retrieving a type.

markwpearce: Maybe? It's not entirely analogous, because I want to remove the boilerplate `?.[0]?.type` code...

src/FunctionScope.ts
```diff
@@ -1,10 +1,13 @@
...
+import type { BrsFile } from './files/BrsFile';
 import type { LabelDeclaration, VariableDeclaration } from './interfaces';
 import type { FunctionExpression } from './parser/Expression';
 
 //TODO I think this class can be eliminated in favor of moving some of these onto the FunctionExpression AST node
 export class FunctionScope {
     constructor(
-        public func: FunctionExpression
+        public func: FunctionExpression,
+        public file: BrsFile
```
markwpearce, line 10: remove this change

src/Scope.spec.ts
```diff
@@ -1678,4 +1683,138 @@ describe('Scope', () => {
...
+                    fur
+                end enum
+            end namespace
+        `;
+
+        it('finds correct return type for class methods', () => {
+            const mainFile = program.setFile('source/main.bs', `
+                sub main()
+                    fooInstance = new Foo()
+                    myNum = foo.getNum()
```
markwpearce, line 1750: This should be fooInstance

src/types/BscType.ts
```diff
@@ -1,6 +1,45 @@
...
-    toString(): string;
-    toTypeString(): string;
+import type { SymbolTableProvider } from '../SymbolTable';
+import { SymbolTypeFlags } from '../SymbolTable';
+import { SymbolTable } from '../SymbolTable';
+import type { Range } from 'vscode-languageserver';
+
+export abstract class BscType {
+
+    protected symbolTable: SymbolTable;
```
markwpearce, line 8: change to `memberTable`

TwitchBronBron (approved):

PR #799, opened 2023-04-27
Adds .kind prop to AstNode.

TwitchBronBron:
- Adds a `.kind` to AstNode, and an `AstNodeKind` enum. 
- converts the reflection methods to check for `.kind` instead of the `thing?.constructor?.name === 'Whatever'` logic.

Raw benchmarks show this as a fairly significant boost:
image

What that means in practice is validation speeds improve by 10-12 percent. 

image

**Downsides:**
This is a breaking change, as all plugins that produce AST would need to be upgraded to the latest version of brighterscript in order to work with the version of brighterscript shipped with the vscode extension. Or, for example, latest version of brighterscript being used with a plugin that depends on an older version of brighterscript.


markwpearce:
Looks good. 
If breaking changes are needed, we shouldn't be afraid of them.

PR #802, opened 2023-05-01
Type chain refactor

markwpearce:
- Adds options to all `AstNode.getType()` calls
- Each AstNode is Responsible for adding to the the passed-in type chain
- Added Utility function to process a type chain

- [ ] Add typechain pushing to indexedGet

### DIFF
--- a/src/interfaces.ts
+++ b/src/interfaces.ts
@@ -14,6 +14,7 @@ import type { SourceMapGenerator, SourceNode } from 'source-map';
 import type { BscType } from './types/BscType';
 import type { AstEditor } from './astUtils/AstEditor';
 import type { Token } from './lexer/Token';
+import type { SymbolTypeFlags } from './SymbolTable';
 
 export interface BsDiagnostic extends Diagnostic {
     file: BscFile;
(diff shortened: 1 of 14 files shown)


src/types/BscType.ts
```diff
@@ -45,3 +45,12 @@ export abstract class BscType {
         throw new Error('Method not implemented.');
     }
 }
+
+
+export class TypeResolution {
```
markwpearce, line 50: Change name to TypeChainEntry

src/types/BscType.ts
```diff
@@ -45,3 +45,12 @@ export abstract class BscType {
...
+    constructor(public name: string, public type: BscType, public range: Range) {
+    }
+    get resolved() {
```
markwpearce, line 53: name change: isResolved

src/util.ts
```diff
@@ -1472,6 +1473,35 @@ export class Util {
             }]);
         }
     }
+
+    public processTypeChain(typeChain: TypeResolution[]): { cannotFindName: string; fullErrorName: string; fullChainName: string; range: Range } {
```
markwpearce, line 1477:
```
{
 cannotFindName => missingItemName;
fullErrorName: string; //remove 
add => lastResolvedTypeName;

 fullChainName: string;
 range: Range
} 
```

src/interfaces.ts
```diff
@@ -411,6 +411,22 @@ export interface FileLink<T> {
...
+
+export class TypeChainEntry {
+    constructor(public name: string, public type: BscType, public range: Range) {
+    }
+    get isResolved() {
+        return this.type?.isResolvable();
+    }
+}
+
+export interface TypeChainProcessResults {
```
TwitchBronBron, line 426:
```suggestion
export interface TypeChainProcessResult {
```
Results makes me think there are multiple "results", when really this represents a single "result" just with multiple properties.

PR #803, opened 2023-05-03
Adds Union types

markwpearce:
- Union Types can be made of any other other types
- Added a utility function that finds the most general type that will fit an array of types
- Refactored `BscType.getMemberType` to `getMemberTypes`... it would be nice to make that all use the helper function, but I had some circular reference issues

- [x] Refactor `isAssignableTo` to be `isTypeCompatible`

These next two will be in a different PR:

- [ ]  ~Add way to actually declare a Union Type in code:~

```
function intValue(num as string | integer) as integer
   if type(num) = "Integer"
     return num
   end if
   return Val(num, 10)
end function 
```

- [ ]  ~Make `UnionType.getMemberTypes` look at the intersection of member types of all the UnionTypes' types. And since this will require coding, I will probably also need to type it (with a keyboard!)~

### DIFF
--- a/src/types/UnionType.ts
+++ b/src/types/UnionType.ts
@@ -1,3 +1,4 @@
+import type { SymbolTypeFlags } from '../SymbolTable';
 import { isDynamicType, isUnionType } from '../astUtils/reflection';
 import { BscType } from './BscType';
 import { isInheritableType } from './InheritableType';
@@ -15,6 +16,10 @@ export class UnionType extends BscType {
     public addType(type: BscType) {
         this.types.push(type);
         this.memberTable.addSibling(type.memberTable);
+    }
+
+    getMemberTypes(name: string, flags: SymbolTypeFlags) {
+        return [];
     }
 
     isAssignableTo(targetType: BscType): boolean {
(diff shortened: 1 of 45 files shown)


src/types/helper.spec.ts
```diff
@@ -0,0 +1,51 @@
...
+        if (isUnionType(resultType)) {
+            expect(resultType.types.length).to.eq(3);
+            expect(resultType.types).to.include(IntegerType.instance);
+            expect(resultType.types).to.include(StringType.instance);
+            expect(resultType.types).to.include(FloatType.instance);
+
+        } else {
+            assert.fail('Should be UnionType');
+        }
+    });
```
markwpearce, line 50: Add some interface tests as well

src/types/UnionType.ts
```diff
@@ -1,27 +1,71 @@
...
     toString(): string {
-        throw new Error('Method not implemented.');
+        return joinTypesString(this.types);
     }
     toTypeString(): string {
-        throw new Error('Method not implemented.');
+        return 'dynamic';
+    }
+
+    canBeAssignedFrom(targetType: BscType) {
```
markwpearce, line 61: Add some docs and examples

markwpearce: `isAssignableFrom` and make it generic for all types

PR #806, opened 2023-05-09
UnionType Improvements

markwpearce:
Improvements/Refactors:
- `UnionType.getMemberTypes` - returns union of inner types with the given name
- `Parser.typeExpression()` - parsing logic changed to only allow VariableExpressions, DottedGetExpressions, or *new* BinaryExpressions with `or` operator. BinaryExpressions in a type results in a `UnionType`


src/parser/Expression.ts
```diff
@@ -850,6 +874,17 @@ export class AALiteralExpression extends Expression {
             walkArray(this.elements, visitor, options, this);
         }
     }
+
+    getType(options: GetTypeOptions): BscType {
```
markwpearce, line 878: I couldn't stop myself from adding this.

markwpearce: remove this for now --- should be a "AssocArray" type or something

PR #823, opened 2023-06-08
Fix injection of duplicate super calls into classes

josephjunker:
Hi, I've encountered the same issue described [in this slack thread](https://rokudevelopers.slack.com/archives/CKF1QQGTY/p1685740439381379), which is preventing my team from adopting Rooibos's code coverage feature. It was mentioned in the thread that the current behavior of only looking at the first statement of a constructor in order to find a `super` call may be a bug. Here is my attempt at a fix. I hope it's helpful, please let me know if there are any changes needed to it.

### DIFF
--- a/src/files/BrsFile.Class.spec.ts
+++ b/src/files/BrsFile.Class.spec.ts
@@ -298,6 +298,45 @@ describe('BrsFile BrighterScript classes', () => {
             `);
         });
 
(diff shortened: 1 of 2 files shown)


TwitchBronBron (approved):

TwitchBronBron:
Looks great! Thanks for this work.

PR #826, opened 2023-06-12
Fixes some issues related to Classes as Properties and Consts validation

markwpearce:
- Fixes when a Class is a Property of another class, and being unable to resolve it
- Fixes issue with Consts value looking at a `typetime` value of the RHS
- Does a better job of caching symbol lookups on memberTables


src/Program.ts
```diff
@@ -689,9 +689,7 @@ export class Program {
             this.logger.time(LogLevel.info, ['Validate all scopes'], () => {
                 for (let scopeName in this.scopes) {
                     let scope = this.scopes[scopeName];
-                    scope.linkSymbolTable();
```
TwitchBronBron, line 692: Why did you delete these?

PR #843, opened 2023-07-13
Make component ready on afterScopeCreate

iBicha:
I was facing a bug where components were not ready by scope creation.
Also another issue is the dependency graph is not set yet before registering the component.
This makes the scope ready when the event is fired


TwitchBronBron (approved):
Sweet! I can honestly say I hadn't encountered this before. Nice fix.

PR #844, opened 2023-07-14
Add beforeProgramDispose event

iBicha:
Adding this event to give plugins a chance for clean up, even in the event of failure (especially in the event of failure)

### DIFF
--- a/src/interfaces.ts
+++ b/src/interfaces.ts
@@ -200,7 +200,7 @@ export interface CompilerPlugin {
     afterProgramValidate?: (program: Program) => void;
     beforeProgramTranspile?: (program: Program, entries: TranspileObj[], editor: AstEditor) => void;
     afterProgramTranspile?: (program: Program, entries: TranspileObj[], editor: AstEditor) => void;
-    beforeProgramDispose?: (program: Program) => void;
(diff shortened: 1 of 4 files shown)


src/interfaces.ts
```diff
@@ -200,6 +200,7 @@ export interface CompilerPlugin {
     afterProgramValidate?: (program: Program) => void;
     beforeProgramTranspile?: (program: Program, entries: TranspileObj[], editor: AstEditor) => void;
     afterProgramTranspile?: (program: Program, entries: TranspileObj[], editor: AstEditor) => void;
+    beforeProgramDispose?: (program: Program) => void;
```
TwitchBronBron, line 203: Can you update this to follow the newer `PluginHandler` pattern? I know some of our other events have ordered params, but we've been moving all new events to this new pattern to make it more flexible.

```
beforeProgramDispose?: PluginHandler<BeforeProgramDisposeEvent>;
```

```
interface BeforeProgramDisposeEvent {
    program: Program;
}

TwitchBronBron (changes requested):
Thanks! Just a small tweak and then this is good. :)

src/interfaces.ts
iBicha: Good idea!

TwitchBronBron (approved):

PR #853, opened 2023-07-22
Functions as params

markwpearce:
- Fixes validations when a function is used as a parameter
- Validates non-functions passed as arguments
- Adds TypedFunctionType class for functions with known param types & return types


src/Program.ts
```diff
@@ -114,7 +114,7 @@ export class Program {
         this.globalScope.symbolTable.addSymbol('double', undefined, DoubleType.instance, SymbolTypeFlag.typetime);
         this.globalScope.symbolTable.addSymbol('dynamic', undefined, DynamicType.instance, SymbolTypeFlag.typetime);
         this.globalScope.symbolTable.addSymbol('float', undefined, FloatType.instance, SymbolTypeFlag.typetime);
-        this.globalScope.symbolTable.addSymbol('function', undefined, new FunctionType(DynamicType.instance), SymbolTypeFlag.typetime);
+        this.globalScope.symbolTable.addSymbol('function', undefined, new TypedFunctionType(DynamicType.instance), SymbolTypeFlag.typetime);
```
TwitchBronBron, line 117: Should this be `FunctionType` instead?

src/Scope.ts
```diff
@@ -973,7 +973,7 @@ export class Scope {
                 const lowerVarName = varName.toLowerCase();
 
                 //if the var is a function
-                if (isFunctionType(varDeclaration.getType())) {
+                if (isTypedFunctionType(varDeclaration.getType())) {
```
TwitchBronBron, line 976: Should this be `isCallableType`?

src/bscPlugin/hover/HoverProcessor.ts
```diff
@@ -158,7 +158,7 @@ export class HoverProcessor {
                 const fullName = processedTypeChain.fullNameOfItem || token.text;
                 const useCustomTypeHover = isInTypeExpression || expression?.findAncestor(isNewExpression);
                 let hoverContent = fence(`${fullName} as ${exprType.toString()}`);
-                if (isFunctionType(exprType)) {
+                if (isTypedFunctionType(exprType)) {
```
TwitchBronBron, line 161: Should this be `isCallableType`?

src/bscPlugin/validation/ScopeValidator.ts
```diff
@@ -384,7 +384,7 @@ export class ScopeValidator {
         //validate all function calls
         for (let expCall of file.functionCalls) {
             const funcType = expCall.expression?.callee?.getType({ flags: SymbolTypeFlag.runtime });
-            if (funcType?.isResolvable() && isFunctionType(funcType)) {
+            if (funcType?.isResolvable() && isTypedFunctionType(funcType)) {
```
TwitchBronBron, line 387: Should this be `isCallableType`?

src/util.ts
```diff
@@ -1058,7 +1058,7 @@ export class Util {
                     case 'float':
                         return new FloatType(token.text);
                     case 'function':
-                        return new FunctionType(new DynamicType(token.text));
+                        return new TypedFunctionType(new DynamicType(token.text));
```
TwitchBronBron, line 1061: Shouldn't this be `new FunctionType` ?

TwitchBronBron (changes requested):
I think there are a few places where you might be using `TypedFunctionType` when you needed to use `FunctionType` or `isCallableType`?

src/bscPlugin/hover/HoverProcessor.ts
markwpearce: this one should be TypedFunctionType, because only typed functions have docs and stuff

src/bscPlugin/validation/ScopeValidator.ts
markwpearce: It is correct, because we can only do function arg validation on typed functions

markwpearce:
Fixed the places that needed it

src/Program.ts
markwpearce: fixed

src/Scope.ts
markwpearce: changed

PR #856, opened 2023-07-26
Adds built in Interfaces to primitive types & Validates class method calls

markwpearce:
image

image

image


src/types/BuiltInInterfaceAdder.ts
```diff
@@ -0,0 +1,86 @@
...
+            // this could be because it's a class that has a parent
+            // the original ancestor should get the built ins
+            return;
+        }
+        const builtInComponent = components[componentName.toLowerCase()];
+        if (!builtInComponent) {
+            throw `Unknown Roku component '${componentName}'`;
+        }
+        if (!this.typedFunctionFactory) {
+            throw `Unable to build typed functions - no typed function factory`;
```
TwitchBronBron, line 35:
```suggestion
            throw new Error('Unable to build typed functions - no typed function factory');
```

src/types/BuiltInInterfaceAdder.ts
```diff
@@ -0,0 +1,86 @@
+import type { BRSInterfaceData } from '../roku-types';
+import { components, interfaces } from '../roku-types';
+import { Cache } from '../Cache';
+import type { TypedFunctionType } from './TypedFunctionType';
+import { SymbolTypeFlag } from '../SymbolTable';
+import type { BscType } from './BscType';
+import { isArrayType, isBooleanType, isCallableType, isClassType, isDoubleType, isEnumMemberType, isEnumType, isFloatType, isIntegerType, isInvalidType, isLongIntegerType, isStringType } from '../astUtils/reflection';
+
+
+// eslint-disable-next-line @typescript-eslint/no-extraneous-class
```
TwitchBronBron, line 10: Is this still necessary?

TwitchBronBron (changes requested):

src/types/BuiltInInterfaceAdder.ts
markwpearce: It's a pure static class... I think that's why.

PR #868, opened 2023-08-03
add noProject flag to bsc so BSConfig.json not expected

philanderson888:


src/cli.ts
```diff
@@ -30,6 +30,7 @@ let options = yargs
     .option('files', { type: 'array', description: 'The list of files (or globs) to include in your project. Be sure to wrap these in double quotes when using globs.' })
     .option('host', { type: 'string', description: 'The host used when deploying to a Roku.' })
     .option('ignore-error-codes', { type: 'array', description: 'A list of error codes that the compiler should NOT emit, even if encountered.' })
+    .option('no-project', { type: 'boolean', defaultDescription: 'false', description: 'when flag set and also project path is null, no bsconfig.json is expected' })
```
TwitchBronBron, line 33:
```suggestion
    .option('noproject', { type: 'boolean', defaultDescription: 'false', description: 'Disable loading all bsconfig.json files' })
```

src/cli.ts
```diff
@@ -62,6 +63,11 @@ async function main() {
     try {
         initProfiling();
 
+        console.log('main()');
+
+        console.log('options ...');
+        console.log(options);
+
```
TwitchBronBron, line 70:
```suggestion
```

src/BsConfig.ts
```diff
@@ -12,6 +12,13 @@ export interface BsConfig {
      */
     project?: string;
 
+    /**
+     * An optional parameter
+     * When value is set to 'true' and 'project' is null, no bsconfig.json is expected
+     */
+    noProject?: boolean;
```
TwitchBronBron, line 19:
```suggestion
    /**
     * Disable loading all bsconfig.json files
     */
    noProject?: boolean;
```

src/util.ts
```diff
@@ -297,12 +297,13 @@ export class Util {
     public normalizeAndResolveConfig(config: BsConfig) {
         let result = this.normalizeConfig({});
 
-        //if no options were provided, try to find a bsconfig.json file
-        if (!config || !config.project) {
-            result.project = this.getConfigFilePath(config?.cwd);
+        result.project = null;
```
TwitchBronBron, line 300: Perhaps you can just do this:

```typescript
if (config?.noproject) {
    return result;
}
```

src/ProgramBuilder.ts
```diff
@@ -101,6 +101,8 @@ export class ProgramBuilder {
             this.options = util.normalizeAndResolveConfig(options);
             if (this.options.project) {
                 this.logger.log(`Using config file: "${this.options.project}"`);
+            } else if (this.options.noProject) {
+                this.logger.log(`No bsconfig.json file expected because using flag --noProject`);
```
TwitchBronBron, line 105: Let's move the `noProject` check first, so it will override the `.project` option.
```suggestion
            this.options = util.normalizeAndResolveConfig(options);
            if (this.options.noProject) {
                this.logger.log(`bsconfig.json loading disabled because 'noProject' option is set`);
            } else if (this.options.project) {
                this.logger.log(`Using config file: "${this.options.project}"`);
            } else
```

src/cli.ts
TwitchBronBron, line 33: I really want to call this one `--noproject` for some reason, but `--no-project` correctly follows our other naming conventions. So...let's add an alias to support both. 

```suggestion
    .option('no-project', { type: 'boolean', defaultDescription: 'false', description: 'when flag set and also project path is null, no bsconfig.json is expected', alias: ['noproject', 'noProject'] })
```

TwitchBronBron (commented):
Hey, here are a few suggestions.

If `--noProject` is present at all, we should disable all project loading, even if the user has supplied `--project`. That should simplify the logic a bit, and makes logical sense to me

philanderson888:
Thank you @TwitchBronBron for that clarification - yes I was not sure which way to go, so went for the softer option!

Happy to go with the hard stop so if the flag is set, bsconfig.json loading is completely disabled.

Happy for you to review the code a second time

Thanks

All suggestions implemented, just a few changes to comments but other than that as-is as you specified.

Please check all code and let me know any faults or amendments, thanks 

:)

TwitchBronBron (approved):
Looks great. Thanks for this!

philanderson888:
Thanks @TwitchBronBron happy to help

Onwards and upwards to the next one!

PR #874, opened 2023-08-28
[WIP] Use Symbol Tables for Completions

markwpearce:
Changes all completion logic to use symbol tables.

Other changes: 
 - Hack to get a parser from an AST Body - this is used to get the documentations tokens. I don't like how this is done (AST Node should not know about the Parser! -> It would be really useful to have a way to map functions to comments!
 - Moved more completions tests to CompletionsProcessor.spec
 - Updates Symbols in SymbolTables to have a data field to store:
    - the node/expression that defines this symbol
    - a string description (from a comment if it is available)
    - completion priority (built in interfaces have lower priority than explicitly defined members)
 - HoverProcessor takes advantage of comments as well
 - fixes possibility that the same namespaceAggregate symbol table is add as sibling more than once 
 - Other small fixes to hovers
 - Adds more incomplete expressions to the AST, so they can be used for completions.


src/parser/Statement.ts
```diff
@@ -53,7 +54,8 @@ export class EmptyStatement extends Statement {
  */
 export class Body extends Statement implements TypedefProvider {
     constructor(
-        public statements: Statement[] = []
+        public statements: Statement[] = [],
+        public parser?: Parser
```
markwpearce, line 58: This is a hack part 1
Basically, I just want to be able to get tokens from an expression statement, so I can look for comments
Is there a better way to do this?

src/parser/Parser.ts
```diff
@@ -232,7 +232,7 @@ export class Parser {
         this.pendingAnnotations = [];
 
         this.ast = this.body();
-
+        this.ast.parser = this;
```
markwpearce, line 235: this is hack part 2

src/bscPlugin/completions/CompletionsProcessor.ts
```diff
@@ -194,341 +198,156 @@ export class CompletionsProcessor {
...
-                    results.set(cs.name.text, {
-                        label: cs.name.text,
-                        kind: CompletionItemKind.Class
-                    });
-                }
-            }
+        const root = symbol.data?.definingNode?.getRoot();
+        let parser: Parser;
+        if (isBody(root)) {
+            parser = root.parser;
```
markwpearce, line 292: This is a hack part 3
Basically, I just want to be able to get tokens from an expression statement, so I can look for comments
Is there a better way to do this?

src/Scope.ts
```diff
@@ -834,21 +834,25 @@ export class Scope {
                             : new NamespaceType(nameSoFar);
                         if (previousNSType) {
                             // adding as a member of existing NS
-                            previousNSType.addMember(nsNamePart, namespace.range, currentNSType, getTypeOptions.flags);
+                            previousNSType.addMember(nsNamePart, { definingNode: namespace }, currentNSType, getTypeOptions.flags);
```
TwitchBronBron, line 837: What was the need for the `definingNode` property? I'm not a huge fan of linking AST to types, would be nice to find a way to avoid this if possible.

TwitchBronBron (approved):
This is fantastic! Thank you.

PR #887, opened 2023-09-06
Task/support bs const in bsconfig

chrisdp:
Adds support for specifying bs_const values in the bsconfig.json. Like this:

```javascript
{
     bs_const: {
        DEBUG: true
    }
}
```

These don't currently update the output manifest file (because we just copy the manifest file from src), but they do impact the conditional compile items.


bsconfig.schema.json
```diff
@@ -21,6 +21,21 @@
...
         },
+        "manifest": {
+            "description": "A entry to modify manifest values",
+            "type": "object",
+            "properties": {
+                "bs_const": {
+                    "description": "A dictionary of bs_consts to change or add to the manifest. Each entry ",
+                    "type": "object",
+                    "patternProperties": {
+                        "^[A-Z0-9._%+-]+@[A-Z0-9.-]+\\.[A-Z]{2,6}$": {
```
chrisdp, line 32:
```suggestion
                        "^.+$": {
```

src/BsConfig.ts
```diff
@@ -12,6 +12,10 @@ export interface BsConfig {
      */
     project?: string;
 
+    manifest?: {
+        bs_const?: Record<string, boolean|null>;
```
TwitchBronBron, line 16:
```suggestion
        bs_const?: Record<string, boolean | null>;
```

TwitchBronBron (approved):

Issue #880, opened 2023-09-02
Adding string and number should be compile error


TwitchBronBron:
With the new type validations in v0.66.0, this should be a compile error because you can't add a string and an int. 

![image](https://github.com/rokucommunity/brighterscript/assets/2544493/0d53ba62-3b32-45a8-a5a4-936b542b1aa9)

markwpearce:
In scope validate’s ast walk, visit binaryOperator expressions, and if neither LHS or RHS are dynamic type, and the util function that finds binary operator results return dynamic type, then raise a new diagnostic… Something like “Operator $operator can not be used with types $lhs and $rhs”

markwpearce:
Solved in https://github.com/rokucommunity/brighterscript/commit/7cafaf12219dcfa73b3e0be04f7acba47b10f93a

PR #896, opened 2023-09-19
Adds validation for binary operators

markwpearce:
Solves  #880

image

### DIFF
--- a/src/globalCallables.ts
+++ b/src/globalCallables.ts
@@ -815,7 +815,7 @@ let programStatementFunctions = [
     }, {
         name: 'Pos',
         shortDescription: 'Returns a number from 0 to window width, indicating the current cursor position on the cursor. Requires a "dummy argument" (any numeric expression).',
-        type: new TypedFunctionType(new VoidType()),
+        type: new TypedFunctionType(new IntegerType()),
         file: globalFile,
         params: [{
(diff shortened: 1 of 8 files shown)


src/globalCallables.ts
```diff
@@ -815,7 +815,7 @@ let programStatementFunctions = [
     }, {
         name: 'Pos',
         shortDescription: 'Returns a number from 0 to window width, indicating the current cursor position on the cursor. Requires a "dummy argument" (any numeric expression).',
-        type: new TypedFunctionType(new VoidType()),
+        type: new TypedFunctionType(new IntegerType()),
```
markwpearce, line 818: Pos(int) retruns an int: https://developer.roku.com/en-ca/docs/references/brightscript/language/program-statements.md#print-item-list

src/parser/Statement.ts
```diff
@@ -2687,7 +2687,7 @@ export class EnumStatement extends Statement implements TypedefProvider {
 
         const resultType = new EnumType(
             this.fullName,
-            members[0]?.getType(options)
+            members[0]?.getType(options).underlyingType
```
markwpearce, line 2690: fixes issue with enums and underlying types

TwitchBronBron (approved):

PR #900, opened 2023-09-19
Interface param support

TwitchBronBron:
Adds support for defining method parameters on interfaces.


markwpearce:
Wow... I didn;t know this was even an issue haha

markwpearce (approved):

PR #901, opened 2023-09-19
Fixes issues with Interface validations, …

markwpearce:
Also adds extra details about missing or mismatches members:
image


src/types/BscType.ts
```diff
@@ -72,20 +72,33 @@ export abstract class BscType {
     }
 
 
-    checkCompatibilityBasedOnMembers(targetType: BscType, flags: SymbolTypeFlag) {
+    checkCompatibilityBasedOnMembers(targetType: BscType, flags: SymbolTypeFlag, data: TypeCompatibilityData = {}) {
```
markwpearce, line 75: I was basically checking this backwords - making sure this type had all the members of the incoming type.

PR #913, opened 2023-09-22
Add some more details to the plugins docs

josephjunker:
This adds some additional information on plugins. If the note on the diagnostics availability in early build steps is too much detail I can remove it.

### DIFF
--- a/docs/plugins.md
+++ b/docs/plugins.md
@@ -323,6 +323,8 @@ export default function () {
 };
 ```
 
+Note that creating diagnostics requires that `BscFile` objects are available. This means that diagnostics cannot be recorded in the `beforeProgramCreate` and `afterProgramCreate` steps.
+
 ## Modifying code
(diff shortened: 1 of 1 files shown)


docs/plugins.md
```diff
@@ -323,6 +323,8 @@ export default function () {
 };
 ```
 
+Note that creating diagnostics requires that `BscFile` objects are available. This means that diagnostics cannot be recorded in the `beforeProgramCreate` and `afterProgramCreate` steps.
```
TwitchBronBron, line 326: This isn't quite true. You can push diagnostics to the program itself with `program.addDiagnostics()`, but you would need to add a "fake" file. I'm doing this in `util.ts`. 
```
let diagnostic = {
    message: "Some diagnostic',
    code: 'some-code',
    file: {
        srcPath: configFilePath
    },
    range: this.getRangeFromOffsetLength(projectFileContents, err.offset, err.length)
} as BsDiagnostic;
```

But yeah, this pattern isn't overly intuitive, but I don't think this paragraph is quite correct, so I'd say let's remove it, or add my above example as an alternative.

josephjunker: That makes sense. I've removed this paragraph, because explaining it might be more confusing then not mentioning it at all.

TwitchBronBron (approved):
Thanks so much for this!

TwitchBronBron (approved):

PR #927, opened 2023-10-05
Semantic Tokes for Native Components and Type completion in Type Expressions

markwpearce:
Semantic tokens for Components Types/Interfaces:

image

Code completion in Type Cast/Type Expression:

image

### DIFF
--- a/src/parser/Parser.ts
+++ b/src/parser/Parser.ts
@@ -412,6 +412,11 @@ export class Parser {
                 });
                 //consume the statement separator
                 this.consumeStatementSeparators();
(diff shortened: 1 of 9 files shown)


src/parser/Parser.ts
```diff
@@ -412,6 +412,11 @@ export class Parser {
                 });
                 //consume the statement separator
                 this.consumeStatementSeparators();
+            } else if (this.peek().kind !== TokenKind.Identifier && !this.checkAny(...DeclarableTypes, ...AllowedTypeIdentifiers)) {
```
markwpearce, line 415: Only try to find a type if the next token can be a type identifier -- basically, don't throw an exception if there isn't the correct kind of token next.
This is important so that we can do completions inside this expression

Issue #926, opened 2023-10-04
Some Intrinsic types have overloaded methods:  👀 `roString`


markwpearce:
There may be others, but `ifStringOps` has a few:

```
Instr(substring as String) as Integer
vs
Instr(start_index as Integer, substring as String) as Integer
```
This is the worst case, because it doesn't conform to the brightscript function pattern.

Also see:
 - `Mid(start_index as Integer) as String` vs `Mid(start_index as Integer, num_chars as Integer) as String`
 - `StartsWith(matchString as String) As Boolean` vs `StartsWith(matchString as String, matchPos as Integer) As Boolean`
 - `EndsWith(matchString as String) As Boolean` vs `EndsWith(matchString as String, length as Integer) As Boolean`

In these cases, the 2nd param can basically be thought of as optional

markwpearce:
This is also a problem/issue for `ifSGNodeFields` with overloaded methods `observeField` and `observeFieldScoped`

markwpearce:
Can probably be mitigated in the `ScopeValidator` by checking function calls against UnionTypes and validating if the call does not meet wither signature...

PR #929, opened 2023-10-06
Adds `callFunc` as member method to Custom Components

markwpearce:
- Custom Components have `callFunc` member
- Adds `isVariadic` flag to `TypedFunctionType`, because CallFunc() may have any number of arguments
- CallFunc is handled in a custom way in the code, because it does not exist in a built-in interface.


src/Program.spec.ts
```diff
@@ -2278,7 +2278,7 @@ describe('Program', () => {
             expectTypeToBe(table.getSymbolType('ifDeviceInfo', opts).getMemberType('GetRandomUUID', rtOpts), TypedFunctionType);
 
             expectTypeToBe(table.getSymbolType('ifSGNodeField', opts), InterfaceType);
-            expectTypeToBe(table.getSymbolType('ifSGNodeField', opts).getMemberType('observeField', rtOpts), TypedFunctionType);
+            expectTypeToBe(table.getSymbolType('ifSGNodeField', opts).getMemberType('addFields', rtOpts), TypedFunctionType);
```
markwpearce, line 2281: Needed to change this because there `observeField` is overloaded, so it is actually a UnionType... see: https://github.com/rokucommunity/brighterscript/issues/926

src/types/ComponentType.ts
```diff
@@ -94,3 +97,10 @@ export class ComponentType extends InheritableType {
         return this.callFuncMemberTable.getSymbolType(name, options);
     }
 }
+
+export const CallFuncMemberMethod = new TypedFunctionType(DynamicType.instance);
+CallFuncMemberMethod.addParameter('functionName', StringType.instance, false);
+CallFuncMemberMethod.isVariadic = true;
+
+// Taken from: https://developer.roku.com/en-ca/docs/developer-program/core-concepts/handling-application-events.md#functional-fields
+export const CallFuncDescription = `callFunc() is a synchronized interface on roSGNode. It will always execute in the component's owning ScriptEngine and thread (by rendezvous if necessary), and it will always use the m and m.top of the owning component. Any context from the caller can be passed via one or more method parameters, which may be of any type (previously, callFunc() only supported a single associative array parameter).\n\nTo call the function, use the \`callFunc\` field with the required method signature. A return value, if any, can be an object that is similarly arbitrary. The method being called must determine how to interpret the parameters included in the \`callFunc\` field.`;
```
TwitchBronBron, line 106: Would it work to move this into the `scrape-roku-docs.ts` instead, so it shows up in the data.json? Just add a hardcoded function. That way this data flows the same as all the others.

TwitchBronBron (changes requested):

PR #944, opened 2023-11-15
Validation Performance: File level `providedSymbols` and `requiredSymbols`

markwpearce:
```
benchmark: Running benchmarks:

    validate@local   --------- 271.514 ops/sec
    validate@0.65.10 --------- 101.219 ops/sec
```

⚡ Huge improvement to validation times! ⚡ 

How it works:

- When a file changes:
  - the file is broken down into individually-validatable segments (eg. each function statement, etc.)
  - each segment is indexed for symbols it requires to compile
  - all the symbols that are available externally to the file are indexed.
  - all the symbols that file requires are indexed (this is a Set of all the required symbols for all its segments)
  - all the symbols that have incompatible types to the previously-known types are marked especially.
- When there's a program validation
  - for the files that changed, check to see:
    - if there are any "duplicate symbols" in the same scope 
    - if the required symbols are are compatible for each scope the file is in
    - if there's a required symbol that is not in any scope
  - do a scope validation, but also pass along a reference to files that have changed, and what the changes are
- When validating a scope:
  - Mark each diagnostic if it was generated for a full scope validation vs. a validation based on an AST segment
  - clear any previously generated diagnostics if they were on a full scope validation
  - for each file in the scope
     -  validate the entire file if it's changed - but only once!
     - if it hasn't changed, and it does not require any symbols that have changed since last validation, do not validate!
     - if it does require something that has changed, than only for the segments in the file that require that symbol, validate the segment.
     - do appropriate clearing of diagnostics attached to the scope based on file and segment re-validations.
 

Known issues:
~The language server still weirdly re-validates for all open files, and there seems to be a race condition somewhere. I see it most notably on Brighterscript files, and when using classes/types defined in namespaces~** these are fixed!**


src/Program.ts
```diff
@@ -537,7 +554,14 @@ export class Program {
...
             if (this.hasFile(srcPath)) {
-                this.removeFile(srcPath);
+                const existingFile = this.files[srcPath.toLowerCase()];
+                if (isBrsFile(existingFile)) {
+                    if (existingFile.fileContents === fileData.toString()) {
+                        // return;
+                    } else {
+                    }
+                }
+                this.removeFile(srcPath, true, true);
```
TwitchBronBron, line 564: Is this refactor still necessary? The commented-out `return` seems to be the only real reason for the change.

PR #946, opened 2023-11-18
Interface optional properties

markwpearce:
image


src/parser/Parser.ts
```diff
@@ -746,11 +755,40 @@ export class Parser {
...
+            }
+
+            name = this.consume(
+                DiagnosticMessages.expectedClassFieldIdentifier(),
+                TokenKind.Identifier,
+                ...AllowedProperties
+            ) as Identifier;
+        } else {
+            // no name after `optional` ... optional is the name
+            name = optionalKeyword as Identifier;
```
TwitchBronBron, line 787: Since `optional` is the name of the field, I think we need to change the optionalKeyword.kind to `Identifier` instead of just casting it in typescript land. 

Like this:
https://github.com/rokucommunity/brighterscript/blob/5a68874bd86dbf88c88db8a0a545b4b00b5db8dd/src/parser/Parser.ts#L361

PR #955, opened 2023-11-22
Add `optional` modifier for interface and class members

TwitchBronBron:
Adds the ability to mark interface fields, interface methods, and class fields as `optional`. We don't really do anything with them right now, but this is in preparation for the upcoming type tracking features.


docs/classes.md
```diff
@@ -608,3 +608,41 @@ sub main()
...
+
+Classes can include optional fields, denoted with the keyword `optional` before the name. If a class has an optional field, it signifies that the value of the field may be `invalid`. Optional fields in a class do not change how the class or its members are validated.
+
+```brighterscript
+class Video
+    url as string
+    length as float
+    optional subtitleUrl as string
+    optional rating as string
+    optional genre as string[]
```
markwpearce, line 623: can't have `string[]` in 0.65 :)

TwitchBronBron: Good catch!

markwpearce (approved):

iObject (approved):

PR #958, opened 2023-11-22
Completion performance

markwpearce:
Previously:
```
for each scope file is in
    on file text completion,
        for each scope the file is in
             get all symbols from entire symbol table
```        

Now:
```
on file text completion,
    get symbols local to the completion position
    for each scope the file was in
         get symbols available at scope level
         get symbols based on namespace inclusion
   get the symbols from global scope

remove duplicates of same completion across different projects
```


TwitchBronBron (approved):
I fixed the build issue. Everything looks great!

PR #971, opened 2023-11-30
Add-component-children-and-field-attributes

iObject:


src/bscPlugin/fileProviders/ComponentStatementProvider.ts
```diff
@@ -56,26 +56,76 @@ export class ComponentStatementProvider {
...
                 ${interfaceMembers.length > 0 ? '</interface>' : ''}
+                ${componentChildren}
             </component>
         `);
 
-
         this.event.files.push(xmlFile);
     }
 
+    private getAnnotationValue(annotations: any) {
```
TwitchBronBron, line 103:
```suggestion
    private getAnnotationValue(annotations: AnnotationExpression[]) {
```

src/bscPlugin/fileProviders/ComponentStatementProvider.ts
```diff
@@ -56,26 +56,76 @@ export class ComponentStatementProvider {
...
+                    let values = args?.token?.text.replaceAll('\"', '').replaceAll(' ', '').split(',');
+                    response = response.concat(values);
+                } else if (isTemplateStringExpression(args)) {
+                    let textOutput = '';
+                    args.quasis[0]?.expressions?.forEach((a: { token: { text: string } }) => {
+                        textOutput += a.token.text;
+                    });
+                    response.push(textOutput);
+                }
+            });
```
TwitchBronBron, line 120: Let's use a for loop here instead, since it's slightly faster at scale
```suggestion
           for (const annotation of annotations) {
                let args = annotation?.call?.args[0];
                if (isVariableExpression(args) || isDottedGetExpression(args)) {
                    response.push(args.name.text);
                } else if (isLiteralExpression(args)) {
                    let values = args?.token?.text.replaceAll('\"', '').replaceAll(' ', '').split(',');
                    response = response.concat(values);
                } else if (isTemplateStringExpression(args)) {
                    let textOutput = '';
                    args.quasis[0]?.expressions?.forEach((a: { token: { text: string } }) => {
                        textOutput += a.token.text;
                    });
                    response.push(textOutput);
                }
            });
```

TwitchBronBron (changes requested):
Looks pretty good! 

I'd like to see some unit tests to verify these features work and ensure they continue to work in the future. You can add them in `ComponentStatement.spec.ts` and follow this existing test's model: 

https://github.com/rokucommunity/brighterscript/blob/086e26710b61dae9070ec6cd78feaa3eee55527a/src/parser/tests/statement/ComponentStatement.spec.ts#L327-L348

Let's add at least one for each annotation type, but feel free to do more with variations on the annotation (like what happens if the annotation is present but missing the string for example). Maybe test a few with several annotations present at the same time, that type of thing.

You run them by calling `npm run test:nocover` or by using the `test` task in vscode.

src/parser/tests/statement/ComponentStatement.spec.ts
```diff
@@ -284,7 +284,7 @@ describe('ComponentStatement', () => {
         `);
     });
 
-    it.only('adds private member to m and creates init function if missing', async () => {
+    it.skip('adds private member to m and creates init function if missing', async () => {
```
iObject, line 287:
```suggestion
    it.only('adds private member to m and creates init function if missing', async () => {
```

PR #989, opened 2023-12-19
Namespace validation fix

georgejecook:
Fixes false diagnostics pertaining to interface and class statements for names that are a substring of a sub-part of a namespace.

### DIFF
--- a/src/Scope.spec.ts
+++ b/src/Scope.spec.ts
@@ -30,7 +30,7 @@ import { InterfaceType } from './types/InterfaceType';
 import { ComponentType } from './types/ComponentType';
 import * as path from 'path';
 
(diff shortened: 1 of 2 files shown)


src/Scope.spec.ts
```diff
@@ -30,7 +30,7 @@ import { InterfaceType } from './types/InterfaceType';
 import { ComponentType } from './types/ComponentType';
 import * as path from 'path';
 
-describe('Scope', () => {
+describe.only('Scope', () => {
```
TwitchBronBron, line 33: Remove the .only
```suggestion
describe('Scope', () => {
```

TwitchBronBron (approved):

PR #993, opened 2023-12-21
Language Server Rewrite

TwitchBronBron:
There are some significant performance issues in the language server, focused around when files are opened/edited/closed from the editor. To mitigate this, we need to refactor the language server so that it's better about throttling/debouncing ALL the relevant files instead of just one at a time.

This is a fairly significant refactor of the language server code, and breaks it down into more manageable classes. 

Task list

- [x] Port All the LSP functions to use the manager
- [x] Handle when watched files change (files not in a tab that were changed like git checkout or ropm install).
- [x] Handle when directories get added or deleted
- [x] Reload single project when bsconfig.json changes
- [x] Prevent revalidation when file is opened but contents haven't changed
- [x] Handle standalone files (files opened in a tab not included in a project)
- [x] Fix bugs with diagnostics being out of sync
- [x] Fix all the LanguageServer tests
- [x] Add new tests for `ProjectManager` and other new logic
- [x] Verify/manage when to pause or cancel validations
- [x] Clean up all unused functions in `LanguageServer.ts`
- [x] ignore changes from `.roku-deploy-staging`, .git, and other ignored folders
- [x] figure out why completions sometimes show global values instead of accurate results
- [x] handle logLevel across bsconfig.json files
- [x] handle quirks around semantic tokens (open file, find and replace all, see they didn't change)
- [x] fix the @Trace() annotation so it doesn't log entire file contents.

## Process for excluding files:
Build a global list of filters, based on:
    - globs from `files.exclude`
    - globs from `.gitignore` (maybe using [this](https://www.npmjs.com/package/gitignore-to-glob)?)
    - all files in ".git", and every workspaceFolder.outDir.
Also produce a per-project list of file globs based on the project's bsconfig (or default)

In the lsp, in onWatchedFilesChanged, filter out any files that match zero of the patterns from above (handling negated .gitignore results and files.exclude).


src/Program.ts
```diff
@@ -1472,3 +1553,10 @@ export interface FileTranspileResult {
     map: SourceMapGenerator;
     typedef: string;
 }
+
+export function test(foo) {
+    const bar = foo + 'Hello';
+
+
+    return bar;
+}
```
iBicha, line 1562: This shouldn't be here I feel

TwitchBronBron: oops! Good catch. 

https://github.com/rokucommunity/brighterscript/pull/1432

PR #996, opened 2023-12-23
Improve null safety

josephjunker:
This adds a second tsconfig which can be used to view null safety errors, and fixes a handful of them. Nothing here should change runtime behavior and I've aimed to make the changes as noninvasive as possible.

- I've been loose with test files, preferring unsafe assertions to code changes in them
- When possible, I've added annotations rather than modify runtime code

I've added explicit type annotations and intermediate interfaces for the return types of some functions and methods which were previously implicit. In some cases this helps inference go through without changes and in other cases it just made it easier for me to determine what exactly was undefined in a nested object.

I've skipped a couple of the harder errors. Most notably, there seems to be an implicit distinction between an "input BsConfig" and an "initialized BsConfig", in that many fields in the BsConfig class are optional but assumed to be present during actual processing. The potentially pedantic way to address this would be to make a second `InitializedBsConfig` type which is used internally and is produced by parsing a BsConfig, but this would be significantly more invasive than what I am aiming for here.

Please let me know if you would prefer any modifications to my approach.

### DIFF
--- a/src/CodeActionUtil.ts
+++ b/src/CodeActionUtil.ts
@@ -11,9 +11,6 @@ export class CodeActionUtil {
         for (const change of obj.changes) {
             const uri = URI.file(change.filePath).toString();
 
-            //justification: `edit` is defined above so we know that `edit.changes` is defined.
-            /* eslint-disable @typescript-eslint/no-non-null-assertion,  @typescript-eslint/no-unnecessary-type-assertion */
-
             //create the edit changes array for this uri
             if (!edit.changes![uri]) {
                 edit.changes![uri] = [];
@@ -27,8 +24,6 @@ export class CodeActionUtil {
                     TextEdit.replace(change.range, change.newText)
                 );
             }
-
-            /* eslint-enable @typescript-eslint/no-non-null-assertion,  @typescript-eslint/no-unnecessary-type-assertion */
         }
         const action = CodeAction.create(obj.title, edit, obj.kind);
         action.isPreferred = obj.isPreferred;
--- a/src/CodeActionUtil.ts
+++ b/src/CodeActionUtil.ts
@@ -11,7 +11,7 @@ export class CodeActionUtil {
         for (const change of obj.changes) {
             const uri = URI.file(change.filePath).toString();
 
-            // Justification: `edit` is defined above so we know that `edit.changes` is defined.
+            //justification: `edit` is defined above so we know that `edit.changes` is defined.
             /* eslint-disable @typescript-eslint/no-non-null-assertion,  @typescript-eslint/no-unnecessary-type-assertion */
 
             //create the edit changes array for this uri
--- a/src/CodeActionUtil.ts
+++ b/src/CodeActionUtil.ts
@@ -11,19 +11,24 @@ export class CodeActionUtil {
         for (const change of obj.changes) {
             const uri = URI.file(change.filePath).toString();
 
+            // Justification: `edit` is defined above so we know that `edit.changes` is defined.
+            /* eslint-disable @typescript-eslint/no-non-null-assertion,  @typescript-eslint/no-unnecessary-type-assertion */
+
             //create the edit changes array for this uri
-            if (!edit.changes[uri]) {
-                edit.changes[uri] = [];
+            if (!edit.changes![uri]) {
+                edit.changes![uri] = [];
             }
             if (change.type === 'insert') {
-                edit.changes[uri].push(
+                edit.changes![uri].push(
                     TextEdit.insert(change.position, change.newText)
                 );
             } else if (change.type === 'replace') {
-                edit.changes[uri].push(
+                edit.changes![uri].push(
                     TextEdit.replace(change.range, change.newText)
                 );
             }
+
+            /* eslint-enable @typescript-eslint/no-non-null-assertion,  @typescript-eslint/no-unnecessary-type-assertion */
         }
         const action = CodeAction.create(obj.title, edit, obj.kind);
         action.isPreferred = obj.isPreferred;
@@ -31,7 +36,7 @@ export class CodeActionUtil {
         return action;
     }
 
-    public serializableDiagnostics(diagnostics: Diagnostic[]) {
+    public serializableDiagnostics(diagnostics: Diagnostic[] | undefined) {
         return diagnostics?.map(({ range, severity, code, source, message, relatedInformation }) => ({
             range: range,
             severity: severity,
--- a/src/CommentFlagProcessor.ts
+++ b/src/CommentFlagProcessor.ts
@@ -179,7 +179,6 @@ export class CommentFlagProcessor {
      * @param text the text to tokenize
      */
     private tokenizeByWhitespace(text: string): Token[] {
-
         let tokens = [] as Array<Token>;
         let currentToken: Token | null = null;
 
--- a/src/CommentFlagProcessor.ts
+++ b/src/CommentFlagProcessor.ts
@@ -51,11 +51,12 @@ export class CommentFlagProcessor {
         let affectedRange: Range;
         if (tokenized.disableType === 'line') {
             affectedRange = util.createRange(range.start.line, 0, range.start.line, range.start.character);
-        } else if (tokenized.disableType === 'next-line') {
+        } else {
+            // tokenized.disableType must be 'next-line'
             affectedRange = util.createRange(range.start.line + 1, 0, range.start.line + 1, Number.MAX_SAFE_INTEGER);
         }
 
-        let commentFlag: CommentFlag;
+        let commentFlag: CommentFlag | null = null;
 
         //statement to disable EVERYTHING
         if (tokenized.codes.length === 0) {
@@ -115,10 +116,10 @@ export class CommentFlagProcessor {
     /**
      * Small tokenizer for bs:disable comments
      */
-    private tokenize(text: string, range: Range) {
+    private tokenize(text: string, range: Range): DisableToken | null {
         let lowerText = text.toLowerCase();
         let offset = 0;
-        let commentTokenText: string;
+        let commentTokenText: string | null = null;
 
         for (const starter of this.commentStarters) {
             if (text.startsWith(starter)) {
@@ -177,9 +178,11 @@ export class CommentFlagProcessor {
      * Given a string, extract each item split by whitespace
      * @param text the text to tokenize
      */
-    private tokenizeByWhitespace(text: string) {
-        let tokens = [] as Array<{ startIndex: number; text: string }>;
-        let currentToken = null;
+    private tokenizeByWhitespace(text: string): Token[] {
+
+        let tokens = [] as Array<Token>;
+        let currentToken: Token | null = null;
+
         for (let i = 0; i < text.length; i++) {
             let char = text[i];
             //if we hit whitespace
@@ -205,4 +208,18 @@ export class CommentFlagProcessor {
         }
         return tokens;
     }
+}
+
+interface Token {
+    startIndex: number;
+    text: string;
+}
+
+interface DisableToken {
+    commentTokenText: string | null;
+    disableType: 'line' | 'next-line';
+    codes: {
+        code: string;
+        range: Range;
+    }[];
 }
--- a/src/DependencyGraph.ts
+++ b/src/DependencyGraph.ts
@@ -137,17 +137,19 @@ export class Node {
     ) {
         if (dependencies.length > 0) {
             this.subscriptions = [];
-        }
-        for (let dependency of this.dependencies) {
-            let sub = this.graph.onchange(dependency, (event) => {
-                //notify the graph that we changed since one of our dependencies changed
-                this.graph.emit(this.key, event);
-            });
 
-            this.subscriptions.push(sub);
+            for (let dependency of this.dependencies) {
+                let sub = this.graph.onchange(dependency, (event) => {
+                    //notify the graph that we changed since one of our dependencies changed
+                    this.graph.emit(this.key, event);
+                });
+
+                this.subscriptions.push(sub);
+            }
         }
     }
-    private subscriptions: Array<() => void>;
+
+    private subscriptions: Array<() => void> | undefined;
 
     /**
      * Return the full list of unique dependencies for this node by traversing all descendents
@@ -161,7 +163,7 @@ export class Node {
             let dependency = dependencyStack.pop();
 
             //if this is a new dependency and we aren't supposed to skip it
-            if (!dependencyMap[dependency] && !exclude.includes(dependency)) {
+            if (dependency && !dependencyMap[dependency] && !exclude.includes(dependency)) {
                 dependencyMap[dependency] = true;
 
                 //get the node for this dependency
--- a/src/DiagnosticFilterer.ts
+++ b/src/DiagnosticFilterer.ts
@@ -112,7 +112,7 @@ export class DiagnosticFilterer {
             let fileDiagnostics = this.byFile[filePath];
             for (let i = 0; i < fileDiagnostics.length; i++) {
                 let diagnostic = fileDiagnostics[i];
-                if (diagnostic.code && filter.codes.includes(diagnostic.code)) {
+                if (filter.codes.includes(diagnostic.code!)) {
                     //remove this diagnostic
                     fileDiagnostics.splice(i, 1);
                     //repeat this loop iteration (with the new item at this index)
--- a/src/DiagnosticFilterer.ts
+++ b/src/DiagnosticFilterer.ts
@@ -6,8 +6,12 @@ import { standardizePath as s } from './util';
 
 export class DiagnosticFilterer {
     private byFile: Record<string, BsDiagnostic[]>;
-    private filters: Array<{ src?: string; codes?: (number | string)[] }>;
-    private rootDir: string;
+    private filters: Array<{ src?: string; codes?: (number | string)[] }> | undefined;
+    private rootDir: string | undefined;
+
+    constructor() {
+        this.byFile = {};
+    }
 
     /**
      * Filter a list of diagnostics based on the provided filters
@@ -24,7 +28,7 @@ export class DiagnosticFilterer {
         let result = this.getDiagnostics();
 
         //clean up
-        delete this.byFile;
+        this.byFile = {};
         delete this.rootDir;
         delete this.filters;
 
@@ -108,7 +112,7 @@ export class DiagnosticFilterer {
             let fileDiagnostics = this.byFile[filePath];
             for (let i = 0; i < fileDiagnostics.length; i++) {
                 let diagnostic = fileDiagnostics[i];
-                if (filter.codes.includes(diagnostic.code)) {
+                if (diagnostic.code && filter.codes.includes(diagnostic.code)) {
                     //remove this diagnostic
                     fileDiagnostics.splice(i, 1);
                     //repeat this loop iteration (with the new item at this index)
@@ -123,7 +127,7 @@ export class DiagnosticFilterer {
         let globalIgnoreCodes: (number | string)[] = [...config1.ignoreErrorCodes ?? []];
         let diagnosticFilters = [...config1.diagnosticFilters ?? []];
 
-        let result = [];
+        let result: Array<{ src?: string; codes: (number | string)[] }> = [];
 
         for (let filter of diagnosticFilters as any[]) {
             if (typeof filter === 'number') {
@@ -151,6 +155,6 @@ export class DiagnosticFilterer {
                 codes: globalIgnoreCodes
             });
         }
-        return result as Array<{ src?: string; codes: (number | string)[] }>;
+        return result;
     }
 }
--- a/src/DiagnosticSeverityAdjuster.ts
+++ b/src/DiagnosticSeverityAdjuster.ts
@@ -16,6 +16,11 @@ export class DiagnosticSeverityAdjuster {
 
     public createSeverityMap(diagnosticSeverityOverrides: BsConfig['diagnosticSeverityOverrides']): Map<string, DiagnosticSeverity> {
         const map = new Map<string, DiagnosticSeverity>();
+
+        if (!diagnosticSeverityOverrides) {
+            return map;
+        }
+
         Object.keys(diagnosticSeverityOverrides).forEach(key => {
             const value = diagnosticSeverityOverrides[key];
             switch (value) {
--- a/src/LanguageServer.ts
+++ b/src/LanguageServer.ts
@@ -1214,8 +1214,7 @@ export class LanguageServer {
 
             const activeSignature = signatures.length > 0 ? 0 : null;
 
-            // TODO: confirm whether this is intentional. `null >= 0` is `true`.
-            const activeParameter = activeSignature >= 0 ? signatures[activeSignature]?.index : null;
+            const activeParameter = activeSignature !== null ? signatures[activeSignature]?.index : null;
 
             let results: SignatureHelp = {
                 signatures: signatures.map((s) => s.signature),
--- a/src/LanguageServer.ts
+++ b/src/LanguageServer.ts
@@ -50,7 +50,7 @@ import type { BusyStatus } from './BusyStatusTracker';
 import { BusyStatusTracker } from './BusyStatusTracker';
 
 export class LanguageServer {
-    private connection = undefined as Connection;
+    private connection = undefined as any as Connection;
 
     public projects = [] as Project[];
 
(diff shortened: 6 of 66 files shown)


src/LanguageServer.ts
```diff
@@ -1214,6 +1214,7 @@ export class LanguageServer {
 
             const activeSignature = signatures.length > 0 ? 0 : null;
 
+            // TODO: confirm whether this is intentional. `null >= 0` is `true`.
```
josephjunker, line 1217: This seems likely to be a bug to me

src/parser/AstNode.ts
```diff
@@ -126,7 +133,7 @@ export abstract class Statement extends AstNode {
     /**
      * Annotations for this statement
      */
-    public annotations: AnnotationExpression[];
+    public annotations: AnnotationExpression[] = [];
```
josephjunker, line 136: The only place that actually changes behavior based on whether `annotations` is `undefined` is a unit test in the parser (which I've updated to check whether it's an empty array instead.) All of the runtime code which uses `annotation` sets a default value of `[]` before using it. I've left those checks in just in case but I can remove them too.

src/util.ts
```diff
@@ -789,8 +791,8 @@ export class Util {
      * @param array the array to flatMap over
      * @param callback a function that is called for every array item
      */
-    public flatMap<T, R>(array: T[], callback: (arg: T) => R) {
-        return Array.prototype.concat.apply([], array.map(callback)) as never as R;
+    public flatMap<T, R>(array: T[], callback: (arg: T) => R[]): R[] {
+        return Array.prototype.concat.apply([], array.map(callback));
```
josephjunker, line 795: This is the same type, spelled differently. In the original `R` was assumed to be an array of some other type, whereas in the new version the array-ness of the return type is explicit. This improves type inference at a usage site.

src/util.ts
```diff
@@ -994,7 +999,10 @@ export class Util {
             }
         }
         if (leftmostRange) {
-            return this.createRangeFromPositions(leftmostRange.start, rightmostRange.end);
+            //if we don't have a rightmost range, use the leftmost range for both the start and end
+            return this.createRangeFromPositions(
+                leftmostRange.start,
+                rightmostRange ? rightmostRange.end : leftmostRange.end);
```
josephjunker, line 1005: This is changed behavior; previously an error would have been thrown in this case. I don't know whether this case would occur in practice.

src/util.ts
```diff
@@ -1133,6 +1142,9 @@ export class Util {
                         // the official plugin format is a factory function that returns a new instance of a plugin.
                     } else if (typeof theExport === 'function') {
                         plugin = theExport();
+                    } else {
+                        //this should never happen; somehow an invalid plugin has made it into here
+                        throw new Error(`TILT: Encountered an invalid plugin: ${String(plugin)}`);
```
josephjunker, line 1147: I tend to sprinkle these assertions into my code, but they are often considered unidiomatic. The alternative here would be to remove the `if` condition from the `else if` above. Removing the condition could allow bad data to get farther into the program before it crashes, but so long as the type of `theExport` (`CompilerPlugin | CompilerPluginFactory`) is correct, no `if` is needed in the `else`.

src/SymbolTable.ts
```diff
@@ -157,4 +157,4 @@ export interface BscSymbol {
 /**
  * A function that returns a symbol table.
  */
-export type SymbolTableProvider = () => SymbolTable;
+export type SymbolTableProvider = () => SymbolTable | undefined;
```
josephjunker, line 160: Other parts of the program pass functions which may return `undefined` as `SymbolTableProvider`s. The place where SymbolTableProviders are read is line 43; this line will return `undefined` whether the `SymbolTableProvider` is `undefined` or whether it is a function which returns `undefined`.

src/CommentFlagProcessor.ts
```diff
@@ -51,11 +51,12 @@ export class CommentFlagProcessor {
         let affectedRange: Range;
         if (tokenized.disableType === 'line') {
             affectedRange = util.createRange(range.start.line, 0, range.start.line, range.start.character);
-        } else if (tokenized.disableType === 'next-line') {
+        } else {
+            // tokenized.disableType must be 'next-line'
```
josephjunker, line 55: The type of `tokenized.disableType` is `'line' | 'next-line'`. If this type is wrong and some third value is possible, then it would be best to not remove the `if` here, and instead either:

- Add a check for the invalid value and throw an error
- Short-circuit and return

I think that the chances of an invalid third value getting in here are low, so I removed the conditional

src/CommentFlagProcessor.ts
```diff
@@ -206,3 +208,17 @@ export class CommentFlagProcessor {
...
+}
+
+interface DisableToken {
+    commentTokenText: string | null;
+    disableType: 'line' | 'next-line';
+    codes: {
+        code: string;
+        range: Range;
+    }[];
+}
```
josephjunker, line 224: inline annotations were getting noisy above

src/DependencyGraph.ts
```diff
@@ -137,17 +137,19 @@ export class Node {
...
 
-            this.subscriptions.push(sub);
+            for (let dependency of this.dependencies) {
+                let sub = this.graph.onchange(dependency, (event) => {
+                    //notify the graph that we changed since one of our dependencies changed
+                    this.graph.emit(this.key, event);
+                });
+
+                this.subscriptions.push(sub);
+            }
```
josephjunker, line 148: The change here is to only iterate through the dependencies if their length is > 0. This isn't a behavioral change, but moving the loop inside the conditional makes it easier to tell TS that `subscriptions` has the value `[]`. The difference between `subscriptions` being `[]` vs `undefined` is important in other places in this file, so I can't do something like always giving it a default value.

src/DiagnosticFilterer.ts
```diff
@@ -24,7 +28,7 @@ export class DiagnosticFilterer {
         let result = this.getDiagnostics();
 
         //clean up
-        delete this.byFile;
+        this.byFile = {};
```
josephjunker, line 31: Nowhere checks whether `byFile` is `undefined`, but several places do iterate over its keys:

- `getDiagnostics` iterates over them without checking their presence; this is safe in the current implementation because it runs after `groupByFile` (which assigns a default value of `{}` to `this.byFile`.
- `filterAllFiles` is in the same boat as `getDiagnostics`.
- `filterFile` deletes a field off of `this.byFile` without checking whether `this.byFile` is defined, but this is also safe for the same reasons

Changing the default value to never be undefined prevents excessive checks or assertions in these places.

src/DiagnosticSeverityAdjuster.ts
```diff
@@ -16,6 +16,11 @@ export class DiagnosticSeverityAdjuster {
 
     public createSeverityMap(diagnosticSeverityOverrides: BsConfig['diagnosticSeverityOverrides']): Map<string, DiagnosticSeverity> {
         const map = new Map<string, DiagnosticSeverity>();
+
+        if (!diagnosticSeverityOverrides) {
+            return map;
+        }
+
```
josephjunker, line 23: Behavioral change; existing implementation will throw an error if `diagnosticSeverityOverrids` is undefined. This is linked to the aforementioned issue with BsConfig optionality; most likely this field will always be defined by the time this method is invoked because initialization will have run on the BsConfig.

src/LanguageServer.ts
```diff
@@ -740,7 +740,7 @@ export class LanguageServer {
         //clone the diagnostics for each code action, since certain diagnostics can have circular reference properties that kill the language server if serialized
         for (const codeAction of codeActions) {
             if (codeAction.diagnostics) {
-                codeAction.diagnostics = codeAction.diagnostics.map(x => util.toDiagnostic(x, params.textDocument.uri));
+                codeAction.diagnostics = codeAction.diagnostics?.map(x => util.toDiagnostic(x, params.textDocument.uri));
```
josephjunker, line 743: Behavioral change: If `codeActions.diagnostics` is `undefined` the changed code will silently ignore the error rather than throwing. I am unsure of whether this can happen in practice.

src/LanguageServer.ts
```diff
@@ -1244,7 +1245,7 @@ export class LanguageServer {
             await Promise.all(this.getProjects().map(project => {
                 return project.builder.program.getReferences(srcPath, position);
             })),
-            c => c
+            c => c ?? []
```
josephjunker, line 1248: Existing implementation can return `undefined` instead of an array, which is filtered out below. As long as the `filter` line is present, returning `[]` vs `undefined` have equivalent behavior. I did not remove the `filter` because I don't know whether anything else in the array could be implicitly undefined.

src/LanguageServer.ts
```diff
@@ -1260,7 +1261,7 @@ export class LanguageServer {
 
     @AddStackToErrorMessage
     @TrackBusyStatus
-    private async onFullSemanticTokens(params: SemanticTokensParams) {
+    private async onFullSemanticTokens(params: SemanticTokensParams): Promise<SemanticTokens | undefined> {
```
josephjunker, line 1264: This could always return `undefined`; the signature change is just to make it explicit.

src/LanguageServer.ts
```diff
@@ -1272,9 +1273,11 @@ export class LanguageServer {
...
             if (project.builder.program.hasFile(srcPath)) {
                 let semanticTokens = project.builder.program.getSemanticTokens(srcPath);
-                return {
-                    data: encodeSemanticTokens(semanticTokens)
-                } as SemanticTokens;
+                if (semanticTokens !== undefined) {
+                    return {
+                        data: encodeSemanticTokens(semanticTokens)
+                    } as SemanticTokens;
+                }
```
josephjunker, line 1280: I've confirmed that is `encodeSemanticTokens` is called with `undefined` that an error will be thrown. So this is another behavioral change, which causes the algorithm to silently continue rather than throw an error in an edge case which I'm not sure will ever happen.

src/Logger.ts
```diff
@@ -22,7 +22,7 @@ export class Logger {
         logLevel?: LogLevel,
         public prefix?: string
     ) {
-        this.logLevel = logLevel;
+        this.logLevel = logLevel ?? LogLevel.log;
```
josephjunker, line 25: This is a little subtle. If `undefined` is passed for `logLevel`, then the setter will be called with `undefined`. The setter defines the type as `LogLevel` and I don't want to change that signature. Internally the setter checks whether its argument is a string; if it's not it falls back to `LogLevel.log`. So this duplicates that "default log level" and sets it in the constructor.

I didn't remove the "is this a string" check from the setter because I don't know whether other code in the system might set the log level to something other than a string.

src/astUtils/reflection.ts
```diff
@@ -23,15 +23,15 @@ import type { AstNode, Expression, Statement } from '../parser/AstNode';
...
     return file?.constructor.name === 'BrsFile';
 }
 
-export function isXmlFile(file: (BscFile)): file is XmlFile {
+export function isXmlFile(file: BscFile | undefined): file is XmlFile {
     return file?.constructor.name === 'XmlFile';
 }
 
-export function isXmlScope(scope: (Scope)): scope is XmlScope {
+export function isXmlScope(scope: Scope | undefined): scope is XmlScope {
```
josephjunker, line 34: These are called with potentially `undefined` values in other places. Given that they already can handle `undefined` gracefully, I chose to expand their signature rather than sprinkle more conditional logic everywhere else.

src/parser/Expression.ts
```diff
@@ -130,7 +130,7 @@ export class FunctionExpression extends Expression implements TypedefProvider {
         super();
         if (this.returnTypeToken) {
             this.returnType = util.tokenToBscType(this.returnTypeToken);
-        } else if (this.functionType.text.toLowerCase() === 'sub') {
+        } else if (this.functionType && this.functionType.text.toLowerCase() === 'sub') {
```
josephjunker, line 133: Another case of "would previously throw an error in an edge case which will likely never happen"

src/parser/TranspileState.ts
```diff
@@ -54,12 +54,12 @@ export class TranspileState {
...
      */
-    public sourceNode(locatable: { range?: Range }, code: string | SourceNode | Array<string | SourceNode>): SourceNode | undefined {
+    public sourceNode(locatable: { range?: Range }, code: string | SourceNode | Array<string | SourceNode>): SourceNode {
         return new SourceNode(
             //convert 0-based range line to 1-based SourceNode line
-            locatable.range.start.line + 1,
+            locatable.range ? locatable.range.start.line + 1 : null,
             //range and SourceNode character are both 0-based, so no conversion necessary
-            locatable.range.start.character,
+            locatable.range ? locatable.range.start.character : null,
```
josephjunker, line 62: `SourceNode` is allowed to be passed `null` here according to the VSCode typedef file, so I chose to do that rather than risk an error being thrown if the range is undefined.

src/util.ts
```diff
@@ -78,7 +78,7 @@ export class Util {
      * Determine if this path is a directory
      */
     public isDirectorySync(dirPath: string | undefined) {
-        return fs.existsSync(dirPath) && fs.lstatSync(dirPath).isDirectory();
+        return dirPath !== undefined && fs.existsSync(dirPath) && fs.lstatSync(dirPath).isDirectory();
```
josephjunker, line 81: Another "would throw in potential edge case"

src/util.ts
```diff
@@ -510,7 +512,7 @@ export class Util {
      * |  bbb | bb   |  bbb |  b   | bbb  |    bb |  bb   |     b | a     |
      * ```
      */
-    public rangesIntersect(a: Range, b: Range) {
+    public rangesIntersect(a: Range | undefined, b: Range | undefined) {
```
josephjunker, line 515: This and `rangesIntersectOrTouch` are called with potentially `undefined` values elsewhere but handle undefinedness gracefully internally, so I've loosened their type signatures.

src/util.ts
```diff
@@ -732,7 +734,7 @@ export class Util {
             //this diagnostic is affected by this flag
             if (diagnostic.range && this.rangeContains(flag.affectedRange, diagnostic.range.start)) {
                 //if the flag acts upon this diagnostic's code
-                if (flag.codes === null || flag.codes.includes(diagnosticCode)) {
+                if (flag.codes === null || (diagnosticCode !== undefined && flag.codes.includes(diagnosticCode))) {
```
josephjunker, line 737: Not a behavioral change, just satisfies the typechecker

src/util.ts
```diff
@@ -1477,7 +1489,7 @@ export class Util {
 
     public validateTooDeepFile(file: (BrsFile | XmlFile)) {
         //find any files nested too deep
-        let pkgPath = file.pkgPath ?? file.pkgPath.toString();
+        let pkgPath = file.pkgPath ?? (file.pkgPath as any).toString();
```
josephjunker, line 1492: The type of `file.pkgPath` is `string`, so unless something unexpected has happened this `toString()` call isn't necessary. Since it seems to only be relevant to cases where the types are wrong I don't think that an `any` cast will hurt anything.

tsconfig-null-safe.json
```diff
@@ -0,0 +1,35 @@
+{
```
TwitchBronBron, line 1: would it be worth extending the existing tsconfig, and then just adding the flags you wanted to override?

TwitchBronBron (changes requested):
Looking great! I found a few things that need addressed, but overall I'm happy with these changes! Thanks a bunch.

TwitchBronBron (changes requested):
Looking great! I found a few things that need addressed, but overall I'm happy with these changes! Thanks a bunch.

src/parser/AstNode.ts
```diff
@@ -47,16 +47,23 @@ export abstract class AstNode {
             if (node.symbolTable) {
                 return node.symbolTable;
             }
-            node = node.parent;
+            //justification: we are following a chain of nodes until we get to one with a SymbolTable,
+            //and the top-level node will always have a SymbolTable
+            // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
```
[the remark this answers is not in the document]
josephjunker: In theory I like having it on in most projects, but I think it's getting more noisy as I go on. There's lots of cases where it gives false positives so I think it would be reasonable to disable the rule entirely.

src/parser/AstNode.ts
josephjunker: I rarely use getters so I'm not 100% certain, but I believe I could use a getter to ensure that the type isn't potentially `undefined` but also only create the array itself as needed. Even if it doesn't matter for performance now it doesn't hurt to keep things lean.

src/Program.spec.ts
```diff
@@ -263,11 +266,11 @@ describe('Program', () => {
         });
 
         it('allows adding diagnostics', () => {
-            const expected = [{
+            const expected: BsDiagnostic[] = [{
                 message: 'message',
                 file: undefined,
                 range: undefined
-            }];
+            }] as any;
```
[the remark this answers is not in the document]
josephjunker: I've changed it to:

```typescript
            const expected = [{
                message: 'message',
                file: undefined,
                range: undefined
}] as any as BsDiagnostic[]
```

There has to be an `any` somewhere. I could also do

```typescript
            const expected = [{
                message: 'message',
                file: undefined as any,
                range: undefined as any
            }] as BsDiagnostic[];
```
as an alternative

src/validators/ClassValidator.ts
```diff
@@ -89,7 +93,8 @@ export class BsClassValidator {
     private findNamespaceNonNamespaceCollisions() {
         for (const [className, classStatement] of this.classes) {
             //catch namespace class collision with global class
-            let nonNamespaceClass = this.classes.get(util.getTextAfterFinalDot(className).toLowerCase());
+            let nonNamespaceClassName = util.getTextAfterFinalDot(className)?.toLowerCase();
+            let nonNamespaceClass = nonNamespaceClassName ? this.classes.get(nonNamespaceClassName) : undefined;
```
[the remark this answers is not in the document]
josephjunker: Yes, it would return `undefined` in any case. TS doesn't like passing `undefined` as an argument to `Map.get()`, so the ternary is just to satisfy the type checker.

I'll change this to:
```
let nonNamespaceClass = this.classes.get(nonNamespaceClassName!);
```

src/validators/ClassValidator.ts
```diff
@@ -376,7 +393,7 @@ export class BsClassValidator {
                         relatedInformation: [{
                             location: util.createLocation(
                                 URI.file(alreadyDefinedClass.file.srcPath).toString(),
-                                this.classes.get(lowerName).range
+                                alreadyDefinedClass.range
```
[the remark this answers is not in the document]
josephjunker: It was actually to satisfy the checker, since frustratingly enough TS doesn't understand this:

```typescript
// map has type Map<string, number>
if (map.has("foo")) {
    const n: number = map.get("foo")
    bar(n);
}
```
Because `has` doesn't refine the return value of `get`, TS thinks `n` can be `undefined`. My usual refactoring is

```typescript
const n: number | undefined = map.get("foo");
if (n !== undefined) {
    bar(n);
}
```

Fortunately 3/4 of this refactoring was already done, so I just grabbed the variable that already existed.

Can't hurt to be more efficient though 😁

tsconfig-null-safe.json
josephjunker: Fixed!

src/parser/Expression.ts
josephjunker: Fixed!

src/bscPlugin/hover/HoverProcessor.ts
```diff
@@ -17,7 +17,7 @@ export class HoverProcessor {
     }
 
     public process() {
-        let hover: Hover;
+        let hover: Hover | null | undefined;
```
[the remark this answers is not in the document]
josephjunker: fixed!

src/bscPlugin/validation/BrsFileValidator.spec.ts
```diff
@@ -1,3 +1,5 @@
+/* eslint-disable @typescript-eslint/no-unnecessary-type-assertion */
```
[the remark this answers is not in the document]
josephjunker: made that change

src/preprocessor/Preprocessor.ts
```diff
@@ -17,7 +17,7 @@ export class Preprocessor implements CC.Visitor {
     /** The set of errors encountered when pre-processing conditional compilation directives. */
     public diagnostics = [] as Diagnostic[];
 
-    public processedTokens: Token[];
+    public processedTokens: Token[] | undefined;
```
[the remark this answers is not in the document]
josephjunker: fixed!

src/preprocessor/PreprocessorParser.ts
```diff
@@ -7,16 +9,16 @@ import { DiagnosticMessages } from '../DiagnosticMessages';
 
 /** * Parses `Tokens` into chunks of tokens, excluding conditional compilation directives. */
 export class PreprocessorParser {
-    public diagnostics: Diagnostic[];
+    public diagnostics: Diagnostic[] = [];
 
-    public tokens: Token[];
+    public tokens: Token[] | undefined;
```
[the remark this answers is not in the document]
josephjunker: Fixed!

src/types/FunctionType.ts
```diff
@@ -11,7 +11,7 @@ export class FunctionType implements BscType {
     /**
      * The name of the function for this type. Can be null
```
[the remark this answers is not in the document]
josephjunker: fixed!

src/types/InterfaceType.ts
```diff
@@ -11,18 +11,19 @@ export class InterfaceType implements BscType {
     /**
      * The name of the interface. Can be null.
```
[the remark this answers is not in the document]
josephjunker: fixed!

src/validators/ClassValidator.ts
```diff
@@ -21,10 +22,13 @@ export class BsClassValidator {
      */
     private classes: Map<string, AugmentedClassStatement>;
 
-    public validate(scope: Scope) {
+    public constructor(scope: Scope) {
         this.scope = scope;
         this.diagnostics = [];
+        this.classes = new Map();
```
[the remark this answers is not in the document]
josephjunker: fixed!

src/CodeActionUtil.ts
```diff
@@ -11,27 +11,32 @@ export class CodeActionUtil {
         for (const change of obj.changes) {
             const uri = URI.file(change.filePath).toString();
 
+            //justification: `edit` is defined above so we know that `edit.changes` is defined.
+            /* eslint-disable @typescript-eslint/no-non-null-assertion,  @typescript-eslint/no-unnecessary-type-assertion */
```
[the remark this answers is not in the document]
josephjunker: done!

src/DiagnosticFilterer.ts
```diff
@@ -108,7 +112,7 @@ export class DiagnosticFilterer {
             let fileDiagnostics = this.byFile[filePath];
             for (let i = 0; i < fileDiagnostics.length; i++) {
                 let diagnostic = fileDiagnostics[i];
-                if (filter.codes.includes(diagnostic.code)) {
+                if (diagnostic.code && filter.codes.includes(diagnostic.code)) {
```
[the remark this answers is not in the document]
josephjunker: fixed!

src/LanguageServer.ts
josephjunker: I fixed and added [a test](https://github.com/rokucommunity/brighterscript/pull/996/files#diff-433a120cab1aa0a63fe01648a9a73f0bc29ad7f5981d92838ec863cdff2d8244R591-R604). I don't have a lot of context here, but it was possible to trigger the condition by invoking `onSignatureHelp` on an XML file. Once I had the test in place I saw that the "bug" wasn't anything scary, and that it's just a field being set to `undefined` when it should have been set to `null`.

src/preprocessor/PreprocessorParser.ts
```diff
@@ -251,10 +251,10 @@ export class PreprocessorParser {
     }
 
     private peek() {
-        return this.tokens[this.current];
+        return this.tokens![this.current];
     }
 
     private previous() {
-        return this.tokens[this.current - 1];
+        return this.tokens![this.current - 1];
```
TwitchBronBron, line 258: Why do you need the `!` here when it's initialized at the top of the file during construction?

TwitchBronBron (changes requested):
Looks great! One question, and then I think this is ready to go!

src/preprocessor/PreprocessorParser.ts
josephjunker: I'm assuming that I added the assertions before adding the default value and then forgot to go back to clean up. Nice catch!

TwitchBronBron:
Looks great! Thanks for this.

Issue #981, opened 2023-12-07
Empty files should be excluded from the project


TwitchBronBron:
If a file becomes empty after transpile, we should remove it from the output, and remove any xml script imports.

TwitchBronBron:
Implemented in #997

PR #997, opened 2023-12-28
Prevent publishing of empty files

MikeAlMartinez:
This PR seeks to fix Issue #981.

The PR adds a config variable `pruneEmptyCodeFiles` which defaults to `false`. When set to `true`, brightscript files that are considered empty won't be published during transpilation. It also adds a `canBePruned` property to the Brs and XML files.

Currently Brs files are considered empty if they don't contain a `FunctionStatement`, `MethodStatement`, or a `ClassStatement`. I might have missed something here.

I can't really think of any reason why an xml file would be considered empty so I've defaulted the value to true for now. One thing I have implemented here for XML files is, imports of empty scripts are removed. If a brightscript file that an xml files references has their `canBePruned` field return false, the associated import will be removed too.

On a large internal project, this resulted in significant compile-time speedups.

- **ultra 4800x (12.5):**
    - pruning off: 7250ms
    - pruning on:  4965ms

- **stick4k 3800x (12.5):**
    - pruning off: 5789ms
    - pruning on:  4342ms

- **express 3960x (11.5):**
    - pruning off: 8232ms
    - pruning on:  6111ms

### DIFF
--- a/src/files/BrsFile.ts
+++ b/src/files/BrsFile.ts
@@ -78,6 +78,27 @@ export class BrsFile {
     }
 
     /**
+     * Should this file be written to disk
+     */
+    public get isPublishable() {
+        let isPublishable = false;
+        this.ast.walk(createVisitor({
+            FunctionStatement: () => {
+                isPublishable = true;
+            },
+            MethodStatement: () => {
+                isPublishable = true;
+            },
+            ClassStatement: () => {
+                isPublishable = true;
+            }
+        }), {
+            walkMode: WalkMode.visitStatementsRecursive
+        });
+        return isPublishable;
+    }
+
+    /**
      * The parseMode used for the parser for this file
      */
     public parseMode = ParseMode.BrightScript;
--- a/src/files/BrsFile.ts
+++ b/src/files/BrsFile.ts
@@ -86,14 +86,11 @@ export class BrsFile {
             FunctionStatement: () => {
                 isPublishable = true;
             },
-            MethodStatement: () => {
-                isPublishable = true;
-            },
             ClassStatement: () => {
                 isPublishable = true;
             }
         }), {
-            walkMode: WalkMode.visitStatementsRecursive
+            walkMode: WalkMode.visitStatements
         });
         return isPublishable;
     }
--- a/src/files/BrsFile.ts
+++ b/src/files/BrsFile.ts
@@ -78,7 +78,7 @@ export class BrsFile {
     }
 
     /**
-     * Should this file be written to disk
+     * Will this file result in only comment or whitespace output? If so, it can be excluded from the output if that bsconfig setting is enabled.
      */
     public get canBePruned() {
         let canPrune = true;
--- a/src/files/BrsFile.ts
+++ b/src/files/BrsFile.ts
@@ -80,19 +80,19 @@ export class BrsFile {
     /**
      * Should this file be written to disk
      */
-    public get isPublishable() {
-        let isPublishable = false;
+    public get canBePruned() {
+        let canPrune = true;
         this.ast.walk(createVisitor({
             FunctionStatement: () => {
-                isPublishable = true;
+                canPrune = false;
             },
             ClassStatement: () => {
-                isPublishable = true;
+                canPrune = false;
             }
         }), {
             walkMode: WalkMode.visitStatements
         });
-        return isPublishable;
+        return canPrune;
     }
 
     /**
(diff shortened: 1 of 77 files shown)


src/files/BrsFile.ts
```diff
@@ -77,6 +77,27 @@ export class BrsFile {
...
+    /**
+     * Should this file be written to disk
+     */
+    public get isPublishable() {
+        let isPublishable = false;
+        this.ast.walk(createVisitor({
+            FunctionStatement: () => {
+                isPublishable = true;
+            },
+            MethodStatement: () => {
```
TwitchBronBron, line 89: Methods can't exist on their own, so we can probably remove this one.

src/files/BrsFile.ts
```diff
@@ -77,6 +77,27 @@ export class BrsFile {
...
+                isPublishable = true;
+            },
+            ClassStatement: () => {
+                isPublishable = true;
+            }
+        }), {
+            walkMode: WalkMode.visitStatementsRecursive
```
TwitchBronBron, line 96: You probably don't need the recursive part (that steps into function bodies which we don't care about for this)

MikeAlMartinez: Yeah, I wasn't sure if perhaps something like a namespace could contain functions so perhaps I needed the recursive version? Happy to change this and the methods version though. 👍

src/files/BrsFile.ts
MikeAlMartinez: removed 👍

src/files/BrsFile.ts
MikeAlMartinez: Done!

TwitchBronBron:
Okay, so I've been toying with a different name for this option. I'm open to other names, but here's what I came up with:

Let's rename `publishEmptyFiles` to `pruneEmptyCodeFiles` and invert the logic and default to `false`. It's a bit more specific to the action being performed. I also like that the default is to _not_ prune them, and then when you enable `pruneEmptyCodeFiles`, bsc will start pruning them.  Also, the term `publish` is a bit loaded because sometimes that could mean actually copying to the device (which we aren't doing with this prop). 

Along with that, I'm thinking we should rename `isPublishable` to `canBePruned`.

Thoughts?

src/files/XmlFile.ts
```diff
@@ -458,27 +463,54 @@ export class XmlFile {
         this.program.logger.debug('XmlFile', chalk.green(this.pkgPath), ...args);
     }
 
+    private determinePkgMapPath(uri) {
+        if (/^pkg:\/.+/.test(uri)) {
+            return uri.replace(/^pkg:\//, '');
+        }
+        return util.getPkgPathFromTarget(this.pkgPath, uri);
+    }
```
TwitchBronBron, line 471: I think you can completely eliminate this function and just call `util.getPkgPathFromTarget` from `checkScriptsForPublishableImports`. We do a similar `pkg:/` check in `getPkgPathFromTarget`.

MikeAlMartinez:
> Okay, so I've been toying with a different name for this option. I'm open to other names, but here's what I came up with:
> 
> Let's rename `publishEmptyFiles` to `pruneEmptyCodeFiles` and invert the logic and default to `false`. It's a bit more specific to the action being performed. I also like that the default is to _not_ prune them, and then when you enable `pruneEmptyCodeFiles`, bsc will start pruning them. Also, the term `publish` is a bit loaded because sometimes that could mean actually copying to the device (which we aren't doing with this prop).
> 
> Along with that, I'm thinking we should rename `isPublishable` to `canBePruned`.
> 
> Thoughts?

That makes sense to me. I wasn't really sold on publish either so will update. 👍

src/files/BrsFile.ts
```diff
@@ -77,6 +77,24 @@ export class BrsFile {
         this.srcPath = value;
     }
 
+    /**
+     * Should this file be written to disk
```
TwitchBronBron, line 81:
```suggestion
     * Will this file result in only comment or whitespace output? If so, it can be excluded from the output if that bsconfig setting is enabled.
```

src/files/XmlFile.ts
```diff
@@ -70,6 +70,11 @@ export class XmlFile {
 
     public commentFlags = [] as CommentFlag[];
 
+    /**
+     * should this file be written to disk
+     */
```
TwitchBronBron, line 75:
```suggestion
    /**
     * Will this file result in only comment or whitespace output? If so, it can be excluded from the output if that bsconfig setting is enabled.
     */
```

TwitchBronBron:
This is fantastic work! Thank you so much.

PR #1000, opened 2024-01-07
Improving null safety: Add FinalizedBsConfig and tweak plugin events

josephjunker:
This only reduces the total number of null check errors by 30, but it pushes the errors a little farther towards the edges. I've resolved most of the errors in `Program` and `ProgramBuilder`. I believe this may be a breaking change, if plugins are allowed to call `new Program()`, because the interface to `Program`'s constructor is now more strict.

I'm a little unsure of what to do about `FinalizedBsConfig.rootDir` and `FinalizedBsConfig.stagingDir`. I'd like to make them required, because a number of places assume that they exist. AFAICT their default values come from roku-deploy though, rather than anywhere in `bsc`, and if I add the default values from roku-deploy to `normalizeConfig`, tons of tests break. I'm figuring that it's better to make incremental progress by making the other fields required than it is to risk breakage by making a more significant change to runtime behavior.

I've added a unit test that asserts that the return value of `normalizeConfig` deeply matches an exact object. I added this test and confirmed that it was passing before making any changes to `normalizeConfig`. This is to give some level of confidence that behavior is unchanged, even though I've had to update the configs in a lot of other tests to satisfy the more strict types.

The changes to the plugin interface are as I mentioned in Slack, and shouldn't be more restrictive than what already exists. I made the name a little prettier than what I originally used in the Slack thread.

### DIFF
--- a/src/Program.ts
+++ b/src/Program.ts
@@ -3,7 +3,7 @@ import * as fsExtra from 'fs-extra';
 import * as path from 'path';
 import type { CodeAction, CompletionItem, Position, Range, SignatureInformation, Location } from 'vscode-languageserver';
 import { CompletionItemKind } from 'vscode-languageserver';
-import type { BsConfig } from './BsConfig';
+import type { FinalizedBsConfig } from './BsConfig';
 import { Scope } from './Scope';
 import { DiagnosticMessages } from './DiagnosticMessages';
 import { BrsFile } from './files/BrsFile';
@@ -60,7 +60,7 @@ export class Program {
         /**
          * The root directory for this program
          */
-        public options: BsConfig,
+        public options: FinalizedBsConfig,
         logger?: Logger,
         plugins?: PluginInterface
     ) {
@@ -74,12 +74,12 @@ export class Program {
         //normalize the root dir path
         this.options.rootDir = util.getRootDir(this.options);
 
-        this.createGlobalScope();
+        this.globalScope = this.createGlobalScope();
     }
 
     public logger: Logger;
 
-    private createGlobalScope() {
+    private createGlobalScope(): Scope {
         //create the 'global' scope
         this.globalScope = new Scope('global', this, 'scope:global');
         this.globalScope.attachDependencyGraph(this.dependencyGraph);
@@ -91,6 +91,8 @@ export class Program {
         this.globalScope.getDiagnostics = () => [];
         //TODO we might need to fix this because the isValidated clears stuff now
         (this.globalScope as any).isValidated = true;
+
+        return this.globalScope;
     }
 
     /**
@@ -320,7 +322,7 @@ export class Program {
     /**
      * roku filesystem is case INsensitive, so find the scope by key case insensitive
      */
-    public getScopeByName(scopeName: string) {
+    public getScopeByName(scopeName: string): Scope | undefined {
         if (!scopeName) {
             return undefined;
         }
@@ -328,7 +330,7 @@ export class Program {
         //so it's safe to run the standardizePkgPath method
         scopeName = s`${scopeName}`;
         let key = Object.keys(this.scopes).find(x => x.toLowerCase() === scopeName.toLowerCase());
-        return this.scopes[key];
+        return this.scopes[key!];
     }
 
     /**
@@ -495,8 +497,8 @@ export class Program {
      * @param rootDir must be a pre-normalized path
      */
     private getPaths(fileParam: string | FileObj | { srcPath?: string; pkgPath?: string }, rootDir: string) {
-        let srcPath: string;
-        let pkgPath: string;
+        let srcPath: string | undefined;
+        let pkgPath: string | undefined;
 
         assert.ok(fileParam, 'fileParam is required');
 
@@ -631,7 +633,7 @@ export class Program {
                 this.plugins.emit('beforeScopeDispose', scope);
                 scope.dispose();
                 //notify dependencies of this scope that it has been removed
-                this.dependencyGraph.remove(scope.dependencyGraphKey);
+                this.dependencyGraph.remove(scope.dependencyGraphKey!);
                 delete this.scopes[file.pkgPath];
                 this.plugins.emit('afterScopeDispose', scope);
             }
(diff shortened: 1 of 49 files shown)


src/Program.ts
```diff
@@ -60,7 +60,7 @@ export class Program {
         /**
          * The root directory for this program
          */
-        public options: BsConfig,
+        public options: FinalizedBsConfig,
```
josephjunker, line 63: If you'd like this to be a backwards-compatible change we could let `BsConfig` be passed here, and then apply `normalizeConfig` inside of the constructor. This feels a little weird to me, because it's letting potentially invalid data get pretty far into the program before getting normalized, but I can't think of any concrete cases where it would be harmful.

src/ProgramBuilder.ts
```diff
@@ -68,7 +68,10 @@ export class ProgramBuilder {
     private staticDiagnostics = [] as BsDiagnostic[];
 
     public addDiagnostic(srcPath: string, diagnostic: Partial<BsDiagnostic>) {
-        let file: BscFile = this.program.getFile(srcPath);
+        if (!this.program) {
+            throw new Error('Cannot call `ProgramBuilder.addDiagnostic` before `ProgramBuilder.run()`');
+        }
```
josephjunker, line 73: An error would be thrown anyway in this case, but I a) wanted to satisfy TS and b) hit this error and was confused by it while writing a plugin in the past, so I thought an explicit error message might be nice.

src/ProgramBuilder.ts
```diff
@@ -33,10 +33,10 @@ export class ProgramBuilder {
      */
     public allowConsoleClearing = true;
 
-    public options: BsConfig;
+    public options: FinalizedBsConfig = util.normalizeConfig({});
     private isRunning = false;
-    private watcher: Watcher;
-    public program: Program;
+    private watcher: Watcher | undefined;
+    public program: Program | undefined;
```
josephjunker, line 39: This may also be a breaking change, but I don't see any good way around that

src/Program.ts
TwitchBronBron: Yeah. Let's just declare `options` as a field down below, and let the incoming type be `BsConfig`. We're already doing a `normalizeConfig` in the constructor anyway.
image.png

src/Program.ts
```diff
@@ -74,12 +74,13 @@ export class Program {
         //normalize the root dir path
         this.options.rootDir = util.getRootDir(this.options);
 
-        this.createGlobalScope();
+        this.globalScope = this.createGlobalScope();
```
TwitchBronBron, line 77: What was the purpose of changing this to an assignment? The global scope was created _in_ `createGlobalScope`.

src/ProgramBuilder.ts
TwitchBronBron: yeah this is fine.

src/ProgramBuilder.ts
```diff
@@ -237,6 +243,9 @@ export class ProgramBuilder {
      * The rootDir for this program.
      */
     public get rootDir() {
+        if (!this.program) {
+            throw new Error('Cannot call `ProgramBuilder.rootDir` until after `ProgramBuilder.run()`');
```
TwitchBronBron, line 247:
```suggestion
            throw new Error('Cannot access `ProgramBuilder.rootDir` until after `ProgramBuilder.run()`');
```

src/ProgramBuilder.ts
```diff
@@ -210,6 +213,9 @@ export class ProgramBuilder {
 
         //on any file watcher event
         this.watcher.on('all', async (event: string, thePath: string) => { //eslint-disable-line @typescript-eslint/no-misused-promises
+            if (!this.program) {
+                throw new Error('TILT: somehow file watcher ran before `ProgramBuilder.run()`');
```
TwitchBronBron, line 217: what does "TILT" mean?

src/ProgramBuilder.ts
```diff
@@ -464,7 +476,7 @@ export class ProgramBuilder {
     /**
      * Parse and load the AST for every file in the project
      */
-    private async loadAllFilesAST() {
+    private async loadAllFilesAST(program: Program) {
```
TwitchBronBron, line 479: I think I'd prefer seeing the `this.program!` down below instead of making it a parameter. That changes the flow less. (people shouldn't be calling this function...but you never know)

TwitchBronBron (changes requested):
Got a few suggestions, but overall looking good!

TwitchBronBron (changes requested):
Got a few suggestions, but overall looking good!

src/Program.ts
TwitchBronBron: Let's keep this PR as-is, and open new ones with other stuff. Feel free to open those other PRs targeting this one so you don't have to keep remembering which one is next. They'll get back in sync as we merge.

src/ProgramBuilder.ts
TwitchBronBron: I think this is fine. It's technically true.

src/ProgramBuilder.ts
josephjunker: I don't remember where I picked that up-- it's the message that a pinball machine displays before it locks up when it detects a player is cheating. Maybe "Internal invariant violation error" would be better?

src/Program.ts
josephjunker: This was to make it easier to convince TS that this was indeed created in the constructor. As an alternative I could declare the scope property on the class as `public globalScope: Scope = undefined as any`, which might be less disruptive.

josephjunker: Updated to `public globalScope: Scope  = undefined as any`

src/ProgramBuilder.ts
josephjunker: fixed!

src/ProgramBuilder.ts
josephjunker: Fixed!

src/ProgramBuilder.ts
josephjunker: I renamed this to hopefully make it more clear.

TwitchBronBron (approved):
Everything looks great. Thanks for this!

src/ProgramBuilder.ts
TwitchBronBron: Yep, much better!

Issue #967, opened 2023-11-29
Flag assignment to undefined class member


TwitchBronBron:
Line 18 should be a diagnostic because the class didn't explicitly define the `aa` property. 

![image](https://github.com/rokucommunity/brighterscript/assets/2544493/da236fbd-1d4f-4f07-a0cd-99cc0566bb50)

Issue #966, opened 2023-11-29
Flag assignments to namespaced entities


TwitchBronBron:
We should not allow assigning values to types on the left hand side. 

![image](https://github.com/rokucommunity/brighterscript/assets/2544493/2bd8597a-ce21-4cd1-8ddb-978648787aa8)

Issue #918, opened 2023-09-23
Handle class private members better


markwpearce:
Only have private class members in code completion when being accessed from `m`
Cause validation error when they are accessed externally

PR #1004, opened 2024-01-09
Adds Diagnostics for Member Accessibility 

markwpearce:
- Adds diagnostics for setting unknown fields of classes

image
image

- Adds diagnostics for accessing `private` or `protected` class members when you shouldn't have access to them.

Private members are available in the class that defined them
Protected members are available in the class that defined them, and all sub classes.

https://github.com/rokucommunity/brighterscript/assets/810290/bcb37181-ea2f-4f7f-8892-47e4105ca37a

Solves #966 
Solves #967 
Solves #918


TwitchBronBron (approved):
Looks great! Nice work.

PR #1008, opened 2024-01-10
Fix cross namespace collision detection

TwitchBronBron:
Fixes a bug where a const and a function are not allowed to have the same name across different namespaces.

### DIFF
--- a/src/Scope.ts
+++ b/src/Scope.ts
@@ -264,10 +264,13 @@ export class Scope {
                 links.push({ item: nsContainersWithStatement?.namespaceStatements?.[0], file: nsContainersWithStatement?.file as BrsFile });
             }
         }
-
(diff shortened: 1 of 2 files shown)


src/Scope.ts
```diff
@@ -264,10 +264,13 @@ export class Scope {
                 links.push({ item: nsContainersWithStatement?.namespaceStatements?.[0], file: nsContainersWithStatement?.file as BrsFile });
             }
         }
-
+        const fullNameLower = (containingNamespace ? `${containingNamespace}.${name}` : name).toLowerCase();
         const callable = this.getCallableByName(name);
```
markwpearce, line 268: Get callable by name will find callables regardless of namespaces... we need to check the namespace after the fact

PR #1017, opened 2024-01-15
adds support for libpkg prefix

georgejecook:
We found that component libraries that use libpkg:/path/file.brs in the script imports would result in files not being found.

This pr adds support for both pkg types.

### DIFF
--- a/src/Program.spec.ts
+++ b/src/Program.spec.ts
@@ -712,7 +712,7 @@ describe('Program', () => {
             program.setFile('components/component1.brs', '');
 
             program.validate();
-            expect(program.getDiagnostics()).to.be.empty;
+            expectZeroDiagnostics(program);
         });
 
         it('accepts libpkg .bs script reference', () => {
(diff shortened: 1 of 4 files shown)


src/Program.spec.ts
```diff
@@ -702,6 +702,60 @@ describe('Program', () => {
...
+                    <script type="text/brightscript" uri="libpkg:/components/component1.brs" />
+                </component>
+            `);
+            program.setFile('components/component1.brs', '');
+
+            program.validate();
+            expect(program.getDiagnostics()).to.be.empty;
+        });
+
+        it('does not set function not found diagnostic for function in libpkg referenced script reference', () => {
```
TwitchBronBron, line 718: Can you add some tests that verify these types of imports also work in .bs files.in import statements.

TwitchBronBron: Btw, I meant:

```
import "libpkg:/something.bs"
```

Issue #1006, opened 2024-01-10
Remove the `.references` item from parser


TwitchBronBron:
The `.references` item is a holdover from before we had proper AST walking. As such, since it's a breaking change, we should remove it before the v1 release.

markwpearce:
Should there be any caching here, or should all references be replaced with a `findChildren()` call ?

PR #1021, opened 2024-01-17
Remove `Parser.references`

markwpearce:
Basically, removes all the hard-coded/static lists in `Parser.references` and instead uses cached AST walks, that are invalidated during BRS File Validation time... At that point in time, all changes from plugins should be done, and we can trust the walk for finding quick lookups for various types of expressions/statements.


markwpearce:
Fixes #1006

src/files/BrsFile.ts
```diff
@@ -105,6 +108,8 @@ export class BrsFile implements BscFile {
 
     public program: Program;
 
+    public cachedLookups: CachedLookups;
```
TwitchBronBron, line 111: Don't forget to mark this as `private` so that plugins don't find this as a replacement for `.parser.references`

TwitchBronBron (approved):

Issue #1018, opened 2024-01-16
Eliminate transpiling `invalid` in callfuncs


TwitchBronBron:
Historically, callfuncs would fail if you didn't pass at least 1 argument (even if that argument was `invalid`). BrighterScript supported this by making the callfunc operator always inject an `invalid` as the first argument if you specified no arguments. 

However, in recent years that problem seems to have been fixed in the Roku OS, so this feature is no longer necessary.

Actions:
 - by default, do NOT inject the `invalid` as the first argument to callfunc operator statements.
 - add a bsconfig option to re-enable this feature (for legacy purposes) since this can result in code not running properly.

PR #1043, opened 2024-01-30
Changed adding invalid as arg to empty callfunc invocations by default

markwpearce:
New `bsconfig.json` option:

```
        "legacyCallfuncHandling": {
            "description": "Legacy RokuOS versions required at least a single argument in callfunc() invocations. Previous brighterscript versions handled this by inserting invalid as an argument when no other args are present. This is not necessary in modern RokuOS versions.",
            "type": "boolean",
            "default": false
        }
```

addresses #1018

### DIFF
--- a/bsconfig.schema.json
+++ b/bsconfig.schema.json
@@ -283,7 +283,7 @@
             "default": false
         },
         "legacyCallfuncHandling": {
-            "description": "Legacy RokuOS versions required at least a single argument in callfunc() invocations. Previous brighterscript versions handled this by inserting invalid as an argument when no other args are present. This is not necessary in modern RokuOS versions.",
+            "description": "Legacy RokuOS versions required at least one argument in callfunc() invocations. Previous brighterscript versions handled this by inserting invalid as an argument when no other args are present. This is not necessary in modern RokuOS versions.",
             "type": "boolean",
             "default": false
         }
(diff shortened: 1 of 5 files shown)


bsconfig.schema.json
```diff
@@ -281,6 +281,11 @@
             "description": "Allow brighterscript features (classes, interfaces, etc...) to be included in BrightScript (`.brs`) files, and force those files to be transpiled.",
             "type": "boolean",
             "default": false
+        },
+        "legacyCallfuncHandling": {
+            "description": "Legacy RokuOS versions required at least a single argument in callfunc() invocations. Previous brighterscript versions handled this by inserting invalid as an argument when no other args are present. This is not necessary in modern RokuOS versions.",
```
TwitchBronBron, line 286:
```suggestion
            "description": "Legacy RokuOS versions required at least one argument in callfunc() invocations. Previous brighterscript versions handled this by inserting invalid as an argument when no other args are present. This is not necessary in modern RokuOS versions.",
```

src/BsConfig.ts
```diff
@@ -195,6 +195,12 @@ export interface BsConfig {
      * scripts inside `source` that depend on bslib.brs.  Defaults to `source`.
      */
     bslibDestinationDir?: string;
+
+    /* Legacy RokuOS versions required at least a single argument in callfunc() invocations.
```
TwitchBronBron, line 199:
```suggestion
    /* Legacy RokuOS versions required at least one argument in callfunc() invocations.
```

TwitchBronBron (changes requested):
Just one small tweak to the wording and then this is good to go!

PR #1059, opened 2024-02-02
Allow Additional v1 Syntax:

markwpearce:
Allows without diagnostic, but no deeper validation:

- built-in types for class member types
- type declarations on LHS of assignments

eg. this is now ok:

```
class Foo
    node as roSGNode
end class

sub someFunc()
    myStr as string = "hello"
end sub
```


TwitchBronBron (approved):
This is great. Thanks!

Issue #1047, opened 2024-01-31
All AST props should be readonly


TwitchBronBron:
All AST properties should be marked as readonly, since the only official way to manipulate AST is by using the AstEditor. This will help guide developers towards using the ASTEditor. 

We should probably verify that the AstEditor will allow us to edit readonly properties....

markwpearce:
Fixed in #1069

PR #1069, opened 2024-02-12
Make All AST Props Readonly

markwpearce:
- Makes all the AST Props `readonly` - both tokens and any expressions referenced.
- Removes `get BrsFile.functionCalls()` - was not used... also removed tests associated with it
- Re-worked some parsing code so it gathers all tokens/expressions before building the node -- this has the consequence that there may be more diagnostics in some situations, because it doesn't end parsing as early as before (eg. try catch blocks)
- A few more small AST tweaks (like renaming `ImportStatement.tokens.filePath` - > `ImportStatment.tokens.path`


markwpearce:
Addresses: #1047

src/parser/Parser.ts
```diff
@@ -877,49 +886,44 @@ export class Parser {
...
             //if the parser was unable to produce a block, make an empty one so the AST makes some sense...
-            if (!func.body) {
-                func.body = new Block({
-                    statements: [],
-                    startingRange: util.createRangeFromPositions(func.range.start, func.range.start)
-                });
-            }
-            func.body.symbolTable = new SymbolTable(`Block: Function '${name?.text ?? ''}'`, () => func.getSymbolTable());
-
-            if (!func.body) {
```
markwpearce, line 902: This code was never called --- if `this.block()` failed, a new Block was created automatically.

src/parser/Parser.ts
```diff
@@ -523,19 +525,19 @@ export class Parser {
...
+        const nameToken = this.tryIdentifier(...this.allowedLocalIdentifiers);
 
-        result.tokens.name = this.tryIdentifier(...this.allowedLocalIdentifiers);
+        this.warnIfNotBrighterScriptMode('enum declarations');
+
+        const parentAnnotations = this.enterAnnotationBlock();
 
         this.consumeStatementSeparators();
+
+        const body: (EnumMemberStatement | CommentStatement)[] = [];
```
TwitchBronBron, line 540: The grouping syntax here resembles arrow functions so I got a little confused until I read the whole line. Maybe use the `Array` type instead?
```suggestion
        const body: Array<EnumMemberStatement | CommentStatement> = [];
```

src/parser/Parser.ts
```diff
@@ -877,49 +886,44 @@ export class Parser {
...
+
+            if (!body) {
+                body = new Block({
+                    statements: [],
+                    startingRange: util.createBoundingRange(
+                        functionType, name, leftParen, ...params, rightParen, asToken, typeExpression, endFunctionType)
+                });
+            }
+
+            body.symbolTable = new SymbolTable(`Block: Function '${name?.text ?? ''}'`, () => func.getSymbolTable());
```
TwitchBronBron, line 915: This doesn't _need_ to be addressed now, but it might be nice to figure out how Plugins could generate this symbol table if they are creating new `Block`s themselves. Or, I guess if they're creating FunctionExpressions, maybe the inner FunctionExpression would create a SymbolTable for the block if there isn't one yet? idk, just something to consider as we get closer to the final release

src/parser/Parser.ts
```diff
@@ -2855,7 +2864,7 @@ export class Parser {
                 while (this.matchAny(TokenKind.Comma, TokenKind.Newline, TokenKind.Colon, TokenKind.Comment)) {
                     // collect comma at end of expression
                     if (lastAAMember && this.checkPrevious(TokenKind.Comma)) {
-                        lastAAMember.tokens.comma = this.previous();
+                        (lastAAMember.tokens as any).comma = this.previous();
```
TwitchBronBron, line 2867: Can we cast this to a more specific type? perhaps `AAMemberExpression`?

TwitchBronBron (commented):
Just a few small comments, then this is good to go. Great work!

TwitchBronBron (commented):
Just a few small comments, then this is good to go. Great work!

src/parser/Parser.ts
markwpearce: I *think* the only block that truly makes it own symbols is a function expression.... so I'll move the code that updates the symbol table to the FuncExpression constructor...

likewise for setting the `functionStatement` prop on functionExpression - I'll move it to the FunctionStatement constructor.

Issue #1046, opened 2024-01-31
Make SG AST Constructors look/behave like new AST Constructors


markwpearce:
eg. Constructors should accept an options object instead of ordered parameters.

PR #1070, opened 2024-02-12
Modifies all SG AST constructors to use named properties in objects

markwpearce:
Addresses: #1046

### DIFF
--- a/src/parser/SGTypes.ts
+++ b/src/parser/SGTypes.ts
@@ -2,12 +2,6 @@ import { SourceNode } from 'source-map';
 import type { Range } from 'vscode-languageserver';
 import { createSGAttribute, createSGInterface, createSGInterfaceField, createSGInterfaceFunction } from '../astUtils/creators';
 import type { FileReference } from '../interfaces';
-import { BooleanType } from '../types/BooleanType';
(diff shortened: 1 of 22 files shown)


src/parser/SGTypes.ts
```diff
@@ -861,25 +859,32 @@ export interface SGReferences {
...
+        component?: SGComponent;
+    } = {}) {
+        this.prolog = options.prolog;
+        this.root = options.root;
+        this.component = options.component;
     }
 
+    public readonly prolog?: SGProlog;
+    public readonly root?: SGElement;
+    public readonly component?: SGComponent;
```
TwitchBronBron, line 874: We actually intentially chose the names `*Element` in #818 to align with the way the HTML dom references their items. Let's keep these original names.

src/parser/SGTypes.ts
```diff
@@ -488,32 +480,6 @@ export enum SGFieldType {
 }
 export const SGFieldTypes = Object.keys(SGFieldType);
 
-export function getBscTypeFromSGFieldType(sgFieldType: string) {
```
markwpearce, line 491: unused function .... better version in `utils.getNodeFieldType`

PR #1075, opened 2024-02-14
Adds validation for calling `new` on a non-class constructor

markwpearce:
image

image

Previously, the check for this was in `ClassValidator` and only checked for `new` with callables. It was moved to `ScopeValidator`, and made more general, so it catches ALL uses of `new` on non-classes

As a result, `CachedLookups.newExpressions` was not used anymore, so it was removed.


TwitchBronBron (commented):
This looks great!

TwitchBronBron (approved):

TwitchBronBron (approved):

PR #1078, opened 2024-02-18
Allow negative patterns in diagnostic filters

josephjunker:
This feature lets users prefix a glob with `!` to make it a negative pattern which un-ignores certain errors. My use case for this is one which I've added to the documentation; I want to ignore errors from all dependencies in a project _except_ for one specific dependency, and I don't want to have to maintain a list of all other dependencies any time one is added or moved.

Using file paths to indicate un-ignores means that it's not possible to un-ignore a code without specifying a file path. A workaround for this is to use the glob `!*/**/*`, matching everything in the rootDir. I've added this to the documentation and the tests.

The implementation is such that it's possible to un-un-un-un-ignore patterns to any depth. This is likely overkill and would be inadvisable to do in practice, but I think the semantics of this are actually simpler and more familiar than it would be to have things bottom out at a specific level.

### DIFF
--- a/docs/bsconfig.md
+++ b/docs/bsconfig.md
@@ -75,7 +75,7 @@ Type: `Array<string | number | {src: string; codes: number[]}`
 A list of filters used to hide diagnostics.
 
 - A `string` value should be a relative-to-root-dir or absolute file path or glob pattern of the files that should be excluded. Any file matching this pattern will have all diagnostics supressed. These file paths refer to the location of the source files pre-compilation and are relative to [`rootDir`](#rootdir). Absolute file paths may be used as well.
-    - A file glob may be previxed with `!` to make it a negative pattern which "un-ignores" the files it matches. (See examples below).
+    - A file glob may be prefixed with `!` to make it a negative pattern which "un-ignores" the files it matches. (See examples below).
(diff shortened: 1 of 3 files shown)


TwitchBronBron (approved):
This looks great!

docs/bsconfig.md
```diff
@@ -74,7 +74,8 @@ Type: `Array<string | number | {src: string; codes: number[]}`
 
 A list of filters used to hide diagnostics.
 
-- A `string` value should be a relative-to-root-dir or absolute file path or glob pattern of the files that should be excluded. Any file matching this pattern will have all diagnostics supressed.
+- A `string` value should be a relative-to-root-dir or absolute file path or glob pattern of the files that should be excluded. Any file matching this pattern will have all diagnostics supressed. These file paths refer to the location of the source files pre-compilation and are relative to [`rootDir`](#rootdir). Absolute file paths may be used as well.
+    - A file glob may be previxed with `!` to make it a negative pattern which "un-ignores" the files it matches. (See examples below).
```
TwitchBronBron, line 78:
```suggestion
    - A file glob may be prefixed with `!` to make it a negative pattern which "un-ignores" the files it matches. (See examples below).
```

TwitchBronBron (approved):
This is great. Thanks so much!

PR #1082, opened 2024-02-28
Empty interfaces break the parser

fumer-fubotv:
Fixes a bug where empty interfaces would break the parser

image


TwitchBronBron (approved):
Nice simple fix. Looks good to me. Thanks!

Issue #1042, opened 2024-01-30
Exclude trivia tokens from lexer output array


TwitchBronBron:
Since tokens now have `leadingTrivia`, we should probably just exclude them from the lexer output array to make it consistent. The trivia tokens will exist only on the leadingTrivia collection.

Issue #1031, opened 2024-01-26
Remove CommentStatement and CommentExpression


TwitchBronBron:
CommentStatement and Comment expression should be removed from the language, and not emitted from the lexer. 

These are considered trivia and can be found in the leadingTrivia of a given token, expression, or statement.

PR #1083, opened 2024-03-03
Removes `CommentStatement` and all comment tokens from Lexer output

markwpearce:
- Comments now reside SOLELY in leading trivia.
- `TranspileState.transpileToken()` will automatically transpile comments (if they exist) ahead of the token
- To still indent comments before a "block closing token" (eg. `}`, `end sub`, etc), use TranspileState.transpileEndBlockToken()`
- Modified getSignatureHelp, getDocumentation, etc. to use leading trivia only

### DIFF
--- a/src/parser/Parser.ts
+++ b/src/parser/Parser.ts
@@ -112,7 +112,12 @@ export class Parser {
      */
     public ast = new Body({ statements: [] });
 
-    public eofToken: Token;
+    public get eofToken(): Token {
+        const lastToken = this.tokens[this.tokens.length - 1];
+        if (lastToken?.kind === TokenKind.Eof) {
+            return lastToken;
+        }
+    }
 
     public get statements() {
         return this.ast.statements;
@@ -195,7 +200,6 @@ export class Parser {
         this.pendingAnnotations = [];
 
         this.ast = this.body();
-        this.eofToken = this.isAtEnd() ? this.peek() : undefined;
         //now that we've built the AST, link every node to its parent
         this.ast.link();
         return this;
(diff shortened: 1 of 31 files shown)


markwpearce:
Addresses #1042 
Addresses #1031

src/parser/Parser.ts
```diff
@@ -113,6 +112,8 @@ export class Parser {
      */
     public ast = new Body({ statements: [] });
 
+    public eofToken: Token;
```
TwitchBronBron, line 115: Let's make this a getter that accesses the final token of `this.tokens[]`

src/parser/SGTypes.ts
```diff
@@ -291,8 +291,8 @@ export class SGElement {
 
     public transpile(state: TranspileState) {
         return new SourceNode(null, null, null, [
-            state.transpileToken(this.tokens.startTagOpen, '<'), // <
-            state.transpileToken(this.tokens.startTagName),
+            ...state.transpileToken(this.tokens.startTagOpen, '<'), // <
```
TwitchBronBron, line 294: the `source-map` library accepts infinitely recursive arrays of arrays of arrays.... so we don't need to do the spread on any of these. I'd rather not do the spread, because `source-map` runs in web-assembly so it'll actually be more efficient for it to traverse the nested arrays. 

So let's undo all the spreads for `state.transpileToken()` and just pust the resulting array there

TwitchBronBron (changes requested):

TwitchBronBron (changes requested):
Had a few suggestions, but overall looks good.

PR #1096, opened 2024-03-08
Fix classname as variable for completions and validations

markwpearce:
https://github.com/rokucommunity/brighterscript/assets/810290/96313697-7232-4e9e-bdc3-60876fc83783

When a Class constructor function is used outside of `new` it is treated as a function type, allowing code completion and validation based on that.

This PR also adds member (and method) validation on primitive types in general

### DIFF
--- a/src/parser/tests/expression/TemplateStringExpression.spec.ts
+++ b/src/parser/tests/expression/TemplateStringExpression.spec.ts
@@ -149,11 +149,11 @@ describe('TemplateStringExpression', () => {
         it('properly transpiles one line template string with expressions', async () => {
             await testTranspile(`
                 sub main()
-                    a = \`hello \${LINE_NUM.text} world \${"template" + "".getChars()} test\`
+                    a = \`hello \${LINE_NUM.text} world \${"template" + "".trim()} test\`
                 end sub
             `, `
                 sub main()
-                    a = ("hello " + bslib_toString(LINE_NUM.text) + " world " + bslib_toString("template" + "".getChars()) + " test")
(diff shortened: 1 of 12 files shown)


src/parser/tests/expression/TemplateStringExpression.spec.ts
```diff
@@ -220,12 +220,12 @@ describe('TemplateStringExpression', () => {
...
         it('properly transpiles more complex multiline template string', async () => {
             await testTranspile(`
-                sub main()
-                    a = \`I am multiline\n\${a.isRunning()}\nmore\`
+                sub main(data)
+                    a = \`I am multiline\n\${data.isRunning()}\nmore\`
                 end sub
             `, `
-                sub main()
-                    a = ("I am multiline" + chr(10) + bslib_toString(a.isRunning()) + chr(10) + "more")
```
markwpearce, line 228: `a` doesn't exist when `a.isRunning()` is called

src/parser/tests/expression/TemplateStringExpression.spec.ts
```diff
@@ -149,11 +149,11 @@ describe('TemplateStringExpression', () => {
         it('properly transpiles one line template string with expressions', async () => {
             await testTranspile(`
                 sub main()
-                    a = \`hello \${LINE_NUM.text} world \${"template" + "".getChars()} test\`
```
markwpearce, line 152: string.getChars() is not a function

TwitchBronBron (approved):

Issue #1091, opened 2024-03-07
Using enums in ternary if statements has unused variable warnings when compiled


agohof:
When using enums to get values in a ternary if statement, the transpiled code will set the enums as params but it won't use them to get the results. Instead, it inserts the actual values directly in the return statements.

### Example

Using the following enums:
```brightscript
enum ThumbButtonSize
    Width = 40
    Height = 37
end enum

enum ClosedCaptionsButtonSize
    Width = 38
    Height = 30
end enum
```

This ternary operation:
```brightscript
buttonBaseWidth = (nodeId = "closedCaptions") ? ClosedCaptionsButtonSize.Width : ThumbButtonSize.Width
```

Will be transpiled into the following code:
```brightscript
buttonBaseWidth = (function(__bsCondition, ClosedCaptionsButtonSize, ThumbButtonSize)
            if __bsCondition then
                return 38
            else
                return 40
            end if
        end function)((nodeId = "closedCaptions"), ClosedCaptionsButtonSize, ThumbButtonSize)
```

When compiled, there will be unused variable warnings because of the `ClosedCaptionsButtonSize` and `ThumbButtonSize` params.

TwitchBronBron:
Yeah, we need to improve the transpiled output to exclude those enums. Good catch.

TwitchBronBron:
Any chance you'd be interested in trying to fix this? I think `util.getExpressionInfo()` needs updated to exclude enums, constants, and interfaces.

https://github.com/rokucommunity/brighterscript/blob/aa92cecad83ce852a6934b0397f6efde6ba941f0/src/parser/Expression.ts#L1381

https://github.com/rokucommunity/brighterscript/blob/aa92cecad83ce852a6934b0397f6efde6ba941f0/src/util.ts#L1189-L1214

PR #1101, opened 2024-03-11
Prevent unused variable warnings on ternary and null coalescence expressions

agohof:
Prevents unused variable warnings when transpiling ternary and null coalescence expressions if they include enums and constants.

Addresses: #1091

### DIFF
--- a/src/parser/Expression.ts
+++ b/src/parser/Expression.ts
@@ -1393,8 +1393,9 @@ export class TernaryExpression extends Expression {
 
     transpile(state: BrsTranspileState) {
         let result = [] as TranspileResult;
-        let consequentInfo = util.getExpressionInfo(this.consequent!);
-        let alternateInfo = util.getExpressionInfo(this.alternate!);
+        const file = state.file;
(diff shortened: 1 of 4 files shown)


src/parser/Expression.ts
```diff
@@ -1476,8 +1477,9 @@ export class NullCoalescingExpression extends Expression {
 
     transpile(state: BrsTranspileState) {
         let result = [] as TranspileResult;
-        let consequentInfo = util.getExpressionInfo(this.consequent);
-        let alternateInfo = util.getExpressionInfo(this.alternate);
+        const file = state.file;
+        let consequentInfo = util.getExpressionInfo(this.consequent, file);
+        let alternateInfo = util.getExpressionInfo(this.alternate, file);
```
TwitchBronBron, line 1482: No need to make a separate variable here, just reference it from state.

```suggestion
        let consequentInfo = util.getExpressionInfo(this.consequent, state.file);
        let alternateInfo = util.getExpressionInfo(this.alternate, state.file);
```

TwitchBronBron (approved):
Looks great, and it's nice and simple. Thanks!

PR #1105, opened 2024-03-11
Move scope.validate() into BscPlugin

TwitchBronBron:
Move the `scope.validate()` call into the BscPlugin


markwpearce (approved):
Nice

src/bscPlugin/validation/ScopeValidator.ts
```diff
@@ -57,6 +74,13 @@ export class ScopeValidator {
 
     private walkFiles() {
 
+        //do many per-file checks for every file in this (and parent) scopes
+        this.event.scope.enumerateBrsFiles((file) => {
```
TwitchBronBron, line 78: This feels wasteful to loop all the files when we're doing a file loop below. Might be worth considering in the future some type of combined "all files in scope" loop that also includes some flags for "is own file" and typedef filtering as well.

But not for this PR

PR #1110, opened 2024-03-14
More validation performance

markwpearce:
Refactors local variable checking to only check un-changed `AstSegment`s for function shadowing if those segments make an assignment to a symbol that changed.

Before:
```
validate@local   ---------- 51.368 ops/sec
validate@0.65.26 --------- 101.435 ops/sec
```

After: 
```
validate@local   --------- 286.331 ops/sec
validate@0.65.26 --------- 102.877 ops/sec
```

### DIFF
--- a/benchmarks/package-lock.json
+++ b/benchmarks/package-lock.json
@@ -10,8 +10,8 @@
             "license": "MIT",
             "dependencies": {
                 "benchmark": "^2.1.4",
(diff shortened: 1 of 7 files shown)


TwitchBronBron:
Woah! That's some serious performance improvements!!! I'll test this out in our huge project to make sure it didn't break anything, but I'm super excited about this

TwitchBronBron (approved):

PR #1111, opened 2024-03-14
Added ArrayDefaultTypeReferenceType for when array symbol is not defined yet

markwpearce:
Needed to add a whole new ReferenceType for this use case!

Because the for-each-loop target was defined on a namespace, it wasn't defined yet at File-validation time (Namespaces are only defined at scope linking), so the default type was unknown.

Now that we have a reference type for this particular case, the array symbol can be defined in the future, and the for-each item will be automatically known.

image

### DIFF
--- a/src/Scope.spec.ts
+++ b/src/Scope.spec.ts
@@ -3167,7 +3167,7 @@ describe('Scope', () => {
             });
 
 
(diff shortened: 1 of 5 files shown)


TwitchBronBron (approved):
Tested and this works great. Thanks!

PR #1126, opened 2024-03-28
Fix bug in range and position caching

fumer-fubotv:
Fixes an issue in `util.createRange()` and `util.createPosition()` where integers over 13 bits long would sometimes generate an incorrect cache key (thus returning the wrong cached value and wrong position/range values).


TwitchBronBron (approved):
Great find! Thanks for this.

Issue #1104, opened 2024-03-11
bs-disable doesn't work for 5-level nested function call


TwitchBronBron:
The `bs:disable-line` comment isn't working for this specific use case.

![bad-interface](https://github.com/rokucommunity/brighterscript/assets/2544493/bd3b921b-572b-474a-83d1-aaa92e47a09b)

What's strange, is deleting 1 character from the function call makes it work again. No idea why.

```brighterscript
namespace http
	function request() as object
		callback = sub()
			if true then
				response = {}
				handleInterceptedScreenData(response._interceptedScreenData) 'bs:disable-line 1001
			end if
		end sub
	end function
end namespace
```

PR #1127, opened 2024-03-29
Fix bug in range and position caching

fumer-fubotv:
https://github.com/rokucommunity/brighterscript/issues/1104

### DIFF
--- a/src/util.ts
+++ b/src/util.ts
@@ -1070,7 +1070,7 @@ export class Util {
      * A cache of `Range` objects. The key is a 52bit integer created from the 4 range integers and leveraging bitshifting.
      * The whole point of this cache is to reduce garbage collection churn, so we didn't want to use string concatenation for the key
      */
-    private rangeCache = new Map<number, Range>();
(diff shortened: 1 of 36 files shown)


src/util.ts
```diff
@@ -1080,22 +1080,15 @@ export class Util {
...
-        if ((startLine > 8191) || (startCharacter > 8191) || (endLine > 8191) || (endCharacter > 8191)) {
-            return {
-                start: this.createPosition(startLine, startCharacter),
-                end: this.createPosition(endLine, endCharacter)
-            };
-        }
-
         // eslint-disable-next-line no-bitwise
-        const key = (startLine << 52) + (startCharacter << 39) + (endLine << 26) + (endCharacter << 13);
+        const key = (endCharacter << 24) + (endLine << 16) + (startLine << 8) + startCharacter;
```
TwitchBronBron, line 1084: Using 8 bits per number only gives us a max of 255 lines in a file or 255 characters in a line before this key overflows.

fumer-fubotv: updated

PR #1135, opened 2024-04-12
Support multi-scope diagnostics

TwitchBronBron:
Adds some diagnostics as multi-scope diagnostics so we don't clutter the diagnostics list with the same diagnostic over and over


src/CrossScopeValidator.ts
```diff
@@ -236,4 +239,40 @@ export class CrossScopeValidator {
...
+                message: `In scope '${scope.name}'`,
+                location: util.createLocation(
+                    URI.file(diagnostic.file.srcPath).toString(),
+                    diagnostic.range
+                )
+            });
+        }
+    }
+
+    private multiScopeCache = new Cache<string, BsDiagnosticWithOrigin>();
```
TwitchBronBron, line 277: We will need to clear this cache between each scope validation (similar to what we do in ScopeValidator)

Issue #1036, opened 2024-01-29
Declaring the type of `m` in a function


markwpearce:
In order to give type hints for `m` in component functions, callbacks, etc, it will be useful to be able to specify/declare its type

the proposal for this currently is to add `type m as _______` after the function signature....:

eg, for this  interface declaration:
```brighterscript
interface MyComponentObject
   top as roSGNodeMyComponent
   isEnabled as boolean
   isSelected as boolean
end interface
```

you can specify `m` as that type in this way:

```brighterscript
sub printDetails() type m as MyComponentObject
   print m.top.text, m.isEnabled
end sub
```

```brighterscript
function handleButtonPress() as boolean, type m as MyComponentObject
   ' we know m is MyComponentObject, m.top is roSGNodeMyComponent, etc.
   if m.isEnabled 
       m.isSelected = true
       m.top.text = "Clicked"
   end if
   return m.isEnabled
end function
```

```brighterscript
sub callOtherFunc()
   doSomethingElse(function() type m as MyComponentObject
           ' this is a call back
           print m.top.text
       end function, otherParam)
end sub
```

TwitchBronBron:
Here's the latest proposal. I think we should move forward with this.

Syntax:
```brighterscript
type m as SomeInterface
```
**Things to consider:**
 - class methods (and functions on AALiterals) retain their own `m` type. But anon functions inside of them would inherit the closest m type override
 - the variable being type-overridden must already exist (i.e. we are not declaring _new_ types)
 - only runtime variables can be type overridden. (i.e. can't do this for classes, interfaces, enums, etc). This basically means it only really applies to `m`, but if closures every show up.
 - These statements _must_ be declared at the _top_ of their enclosing scope (top of file, top of namespace, top of function). They can be intermixed with import statements at top of file.

markwpearce:
fixed with #1140

PR #1140, opened 2024-04-22
Adds `typecast m as <Type>` statement

markwpearce:
Adds ability to add a single statement to beginning of file or function, and every usage of `m` in that or lower scopes will be typecast as the type given.

Syntax: 

```
typecast m as <type>
```

eg:
```
import "types.bs"
typecast m as Thing1

sub func1()
    x = m.value ' x is type of Thing1.value
    print x
end sub

sub func2()
    typecast m as Thing2 
    x = m.value  ' x is type of Thing2.value
    print x
end sub

sub func3()
    aa = {
        innerFunc: sub()
            typecast m as Thing3
            x = m.value  ' x is type of Thing3.value
            print x
        end sub
    }
end sub

class TestKlass
    value as string

    sub method1()
        x = m.value ' x is type of TestKlass.value
        print x
    end sub

    sub method2()
        typecast m as Thing1
        x = m.value  ' x is type of Thing1.value
        print x
    end sub
end class

```

### DIFF
--- a/docs/typecasts.md
+++ b/docs/typecasts.md
@@ -1,0 +1,111 @@
+# BrighterScript Typecasts
+
+In BrighterScript, the inferred types of expressions can be explicitly changed through typecasts.
+
+There are two ways of doing this - through inline typecasts, or through typecast statements.
+
+Any expressions that has been typecast will be assumed to be that type.
+
+## Inline typecasts
+
+To invoke an inline typecast, use the `as <type>` syntax after an expression. The whole typecast can also be enclosed in parentheses as well.
+
+In this example, the `node` parameter is `dynamic`, but is cast to the built-in type `roSgNode`. `roSgNode` has an `id` property that is a `string`.
+Due to the assignment, `nodeId` is also known to be a string.
+
+Because of the typecast, the intellisense code completion also knows the type, so at the `.` after the typecast, members of `roSgNode` will be suggested.
+
+Example:
+
+```BrighterScript
+function getPrefixedId(node, prefix as string) as string
+    nodeId = (node as roSgNode).id
+    return prefix + nodeId
+end function
+```
+
+## Typecast Statments
+
+In order to specify all usages of an expression are explicity set to a certian type, you can use a `typecast` statement.
+
+Syntax:
+
+`typecast m as <type>`
+
+Rules:
+
+- `Typecast` statements are only allowed at the top of files (along with `import` and `library` statements), and as the first statement in a function or namespace.
+- There can be a maximum of one `typecast` statement per block
+- Only runtime symbols can be typecast
+- Currently only `m` is supported
+
+This is very usefuly for components, so that the actual type of Associative Array `m` can narrowed to provide better validation.
+
+**pkg:/components/Widget.xml**
+
+```xml
+<?xml version="1.0" encoding="utf-8" ?>
+<component name="Widget" extends="Group">
+    <script uri="Widget.bs" />
+    <script uri="Widget.interface.bs" />
+    <interface>
+        <field id="enabled" type="boolean" value="false" />
+        <field id="state" type="string" />
+    </interface>
+    <children>
+        <label id="myLabel" />
+    </children>
+</component>
+```
+
+**pkg:/components/Widget.interface.xml**
+
+```BrighterScript
+interface WidgetObject
+    top as roSgNodeWidget
+    label as roSgNodeLabel
+    value as string
+end interface
+```
+
+**pkg:/components/Widget.bs**
+
+```BrighterScript
+typecast m as WidgetObject
+
+sub init
+    m.label = m.top.findNode("myLabel")
+    m.value = "Test"
+    m.top.state = "ok"
+    printDetails()
+end sub
+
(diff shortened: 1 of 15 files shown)


markwpearce:
addresses: #1036

src/parser/Parser.ts
```diff
@@ -1030,6 +1031,10 @@ export class Parser {
             return this.importStatement();
         }
 
+        if (this.check(TokenKind.TypeCast) && this.checkNextText(TokenKind.Identifier, 'm')) {
```
TwitchBronBron, line 1034: Let's allow any variable in this location during parsing, not just `m`. Then we will add a diagnostic during validation  saying "only `m` is allowed here". That makes the AST slightly more resilient to recognize this syntax no matter what variable shows up, while still restricting to `m` downstream.

src/parser/Statement.ts
```diff
@@ -3415,3 +3415,68 @@ export class ContinueStatement extends Statement {
...
+            this.tokens.typeCast,
+            this.tokens.m,
+            this.tokens.as,
+            this.typeExpression
+        );
+    }
+
+    public readonly tokens: {
+        readonly typeCast?: Token;
+        readonly m?: Token;
```
TwitchBronBron, line 3445: Let's make this an `Expression`. In this PR we only support `VariableExpression`, but it could expand to others the future so let's keep it as broad as possible to reduce the chance of breaking changes (or needless duplicate AST nodes). 

Not sure what to name the thing. `item`? `element`? `subject`? `obj`?
```suggestion
        readonly item?: Expression;
```

src/parser/Statement.ts
```diff
@@ -3415,3 +3415,68 @@ export class ContinueStatement extends Statement {
         return this.tokens.continue?.leadingTrivia ?? [];
     }
 }
+
+
+export class TypeCastMStatement extends Statement {
```
TwitchBronBron, line 3420: Let's call this `TypecastStatement`. While the current implementation is restricted to `m`, we might expand this in the future (casting global functions, consts, etc...), so I want to leave the door open to support those without breaking AST changes.

src/astUtils/CachedLookups.ts
```diff
@@ -53,6 +53,10 @@ export class CachedLookups {
         return this.getFromCache<Array<ImportStatement>>('importStatements');
     }
 
+    get typeCastMStatements(): TypeCastMStatement[] {
```
TwitchBronBron, line 56:
```suggestion
    get typecastStatements(): TypecastStatement[] {
```

TwitchBronBron (changes requested):
Looks pretty good. I have a few high-level changes:
 - let's treat typecast as 1 word (i.e. `typecast`, `typecastStatements`, `TypecastStatement`) instead of uppercasing the `c`. 
 - With respect to the _naming_ of the AST nodes and variables, I'd like this concept to remain generic and not focused on exclusively `m` (So call it `TypecastStatement` instead of `TypecastMStatement`). This PR will remain focused exclusively on only supporting `m`, but we might expand in the future so I want to be able to reuse the same AST nodes for that. 
 - if it's not too much extra work, we should support `typecast` being used at the top of namespace statement bodies as well, which would apply to only that namespace block (and its enclosed children blocks in that same file). Might be complicated, so it's fine to push off to a future PR after v1 lands.

src/parser/Statement.ts
markwpearce: done

src/parser/Statement.ts
markwpearce: Using `TypecastExpression` here, like I always should have.

src/bscPlugin/validation/BrsFileValidator.ts
```diff
@@ -352,6 +358,28 @@ export class BrsFileValidator {
...
+                                file: this.event.file
+                            });
+                        }
+                    } else {
+                        this.event.file.diagnostics.push({
+                            ...DiagnosticMessages.typeCastMStatementMustBeDeclaredAtStart(),
+                            range: result.range,
+                            file: this.event.file
+                        });
+                    }
```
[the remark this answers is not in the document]
markwpearce: now using single function `validateTypecastStatements()` to do specific validation

src/DiagnosticMessages.ts
```diff
@@ -100,8 +100,8 @@ export let DiagnosticMessages = {
         code: 1015,
         severity: DiagnosticSeverity.Error
     }),
-    expectedIdentifierAfterKeyword: (keywordText: string) => ({
-        message: `Expected identifier after '${keywordText}' keyword`,
+    expectedIdentifierAfterKeyword: (keywordText: string, expectedText = '') => ({
```
TwitchBronBron, line 103: Do we need the `expectedText` anymore? I don't see it being used anywhere now.

TwitchBronBron (approved):
Looks great!

docs/typecasts.md
```diff
@@ -0,0 +1,111 @@
+# BrighterScript Typecasts
```
TwitchBronBron, line 1: Nice!

PR #1145, opened 2024-04-24
Fixes CompletionItemKinds for typed variables

### LINKED ISSUE #1133: Wrong completion type for variable of type interface
Variables should never be of type interface. In the below example, runResult is a local variable that is an interface type. So that symbol should show the `variable` completion type. only actual interface names should show that interface symbol.

image

markwpearce:
Fixes #1133 

All variables that are Classes, Interfaces, Enum, functions, etc. will be displayed as variables.


src/bscPlugin/completions/CompletionsProcessor.ts
```diff
@@ -371,21 +371,32 @@ export class CompletionsProcessor {
     private getCompletionKindFromSymbol(symbol: BscSymbol, areMembers = false) {
         const type = symbol?.type;
         const extraData = symbol?.data;
+        // eslint-disable-next-line no-bitwise
```
TwitchBronBron, line 374: Where's the bitwise?

Issue #1023, opened 2024-01-19
Conditional compile AST


TwitchBronBron:
The AST is currently missing the conditional compile AST items. They get parsed before the primary parser, and the the `false` evaluated blocks get stripped before the parser is invoked. 

- We need to add proper AST nodes for the conditional compile blocks, and eliminate the `preprocessor` concept entirely.
- transpile needs to output the code (right now it gets stripped)
- - We need to consider how AST walking will work. because we will now have extra `#if` blocks. Perhaps another `WalkMode` flag that represents "step into false conditional compile blocks", and it would be disabled by default

PR #1149, opened 2024-04-29
Adds Conditional Compilation Statements to the AST

markwpearce:
Conditional compile directives (`#if`, `#const`, `#error`, etc.) are now part of the AST, and as such, the branches of the condition compiles can be walked.

There is one concession with this: `#if` blocks are treated as any other block, and everything in them must be self contained, or will cause a diagnostic.

That means that even though this is allowed in Brightscript:
```
sub foo()
# if true
end sub
#end if
```

it will cause some diagnostics, because the `sub foo()` is partially inside and outside the `#if` statement

- [x] TODO - Add some walk mode options that allow walking the tree based on true/false const values.

- [x] Support `#if not <bs_const>` #1158 

Addresses #1023


markwpearce:
I suppose we should support `#if not <bs_const>` as well? #1158

TwitchBronBron:
> I suppose we should support `#if not <bs_const>` as well? #1158

Yes, if we can!

src/parser/Statement.ts
```diff
@@ -3536,3 +3576,188 @@ export class AliasStatement extends Statement {
...
+
+        //end if
+        results.push(...state.transpileEndBlockToken(this.elseBranch ?? this.thenBranch, this.tokens.hashEndIf, '#end if'));
+
+        return results;
+    }
+
+    walk(visitor: WalkVisitor, options: WalkOptions) {
+        if (options.walkMode & InternalWalkMode.walkStatements) {
+            let conditionTrue = options.bsConsts?.get(this.tokens.condition.text.toLowerCase());
```
TwitchBronBron, line 3696: Shouldn't we fall back to the parent bsConsts if not provided?
```typescript
const bsConsts = options.bsConsts ?? this.getBsConsts();
```
That way it supports calling .walk on any ast node.

TwitchBronBron (changes requested):

TwitchBronBron (approved):

Issue #1113, opened 2024-03-14
Add alias statement


TwitchBronBron:
Add an `alias` statement to the language. This will provide a workaround for shadowed items. 

Consider this example. Due to the namespace-relative functionality, the `get` method found on line 22 is the outer `get` function on line 21. However, we _wanted_ to call `get.aa()` from line 27. We currently have no way to work around this.

![image](https://github.com/rokucommunity/brighterscript/assets/2544493/fbdc7c28-0200-4cd0-b5f1-79a1730b4fd4)

That's where the `alias` statement comes in. It allows declaring a new name for an existing symbol. 
```brighterscript
alias get2 = get

namespace http
    'Do an HTTP request
    sub get()
        print get2.aa() 'using `get2` aliased symbol here. it's now clear which item we intended to use
    end sub
end namespace

namespace get
    function aa()
    end function
end namespace
```

**Criteria:**
- aliases are scoped to the current file, and will not leak into the outer component or source scopes. 
- at transpile time, all referenced items will be transpiled using the _original_ name (i.e. `get2.aa()` => `get_aa()`)
- alias name may not have the same name as an existing symbol.

PR #1151, opened 2024-05-02
Adds `Alias` Statement

markwpearce:
This is pretty cool!

`alias` provides a way to work around variable name shadowing... but as a side effect, it also give a way to shorthand just about any property/function you want.

```BrighterScript
alias get2 = get

namespace http
    'Do an HTTP request
    sub get()
        print get2.aa().data 'using `get2` aliased symbol here. it's now clear which item we intended to use
    end sub
end namespace

namespace get
    function aa()
        return {
            data: "abc"
        }
    end function
end namespace
```

transpiles to

```BrightScript
sub http_get()
    print get_aa().data
end sub

sub get_aa()
    return {
        data: "abc"
    }
end sub
```

it also allows this ... which is cool?:

```
alias lc = lcase
alias co = createObject

sub getModelTypeLower()
   lc(co("roDeviceInfo").getModelType())
end sub
```

it also allows this ... which is also cool?:

```
alias abgdz = alpha.beta.gamma.delta.zeta

sub foo()
   print abgdz.someConst
end sub
```


markwpearce:
addresses #1113

src/files/BrsFile.spec.ts
```diff
@@ -2877,6 +2877,215 @@ describe('BrsFile', () => {
...
+                    end function
+
+                    sub alpha_printDir()
+                        print "North"
+                        print "North"
+                    end sub
+                `);
+            });
+
+            it('can deep alias a namespaced thing', async () => {
```
TwitchBronBron, line 3005: Can you add a test for aliased class? And make sure we can `new` the class using the alias name? And that it transpiles properly?
i.e.:
```
alias Person2 = Person
sub test()
     dude = new Person2()
end sub
```

TwitchBronBron (changes requested):
Looks great! Just a few small things

markwpearce:
Updated to fixes review comments.

TwitchBronBron (approved):

PR #1154, opened 2024-05-03
`CreateObject` function call implies correct return type

markwpearce:
Previously, everything returned from `createObject` was `ObjectType`

Now, it looks at the arguments, and if passed string literals, it is able to infer either the built-in Component/Object or a custom component:

```brighterscript
dateTime = createObject("roDatetime") ' this is the InterfaceType `roDateTime`
dateTime.getHours() ' this is an integer


button = createObject("roSgnode", "Poster") '  this is the ComponentType `roSgNodePoster`
button.uri ' this is a string
```

if the args are not string literals, it still just returns ObjectType....

```brighterscript
function getObject(objectName as string)
    return createObject(objectName) ' this is just ObjectType
end function
```


src/files/tests/optionalChaning.spec.ts
```diff
@@ -80,12 +80,12 @@ describe('optional chaining', () => {
                 print arr?.value
                 print obj?.[0]
                 print obj?.getName()?.first?.second
-                print createObject("roByteArray")?.value
+                print createObject("roByteArray")?.Count
```
markwpearce, line 83: I love it when adding more type info makes the tests fail because they were wrong :)

markwpearce:
One cool side effect of this is that it opens the door to generics - the return type can be based on the specifics of the call expression....

src/types/TypedFunctionType.ts
```diff
@@ -5,10 +5,12 @@ import { BscTypeKind } from './BscTypeKind';
 import { isUnionTypeCompatible } from './helpers';
 import { BuiltInInterfaceAdder } from './BuiltInInterfaceAdder';
 import type { TypeCompatibilityData } from '../interfaces';
+import type { CallExpression } from '../parser/Expression';
 
 export class TypedFunctionType extends BaseFunctionType {
     constructor(
-        public returnType: BscType
+        public returnType: BscType,
+        public returnTypeLookup?: (callExpr: CallExpression) => BscType
```
TwitchBronBron, line 13: Hmm. I'm not crazy about having this callback function.

markwpearce:
Ok. Hmmm.   I don’t know about making support for overloading functions yet, so maybe I’ll change this so it’s just a special case, and does not change the interface/constructors for anything yet.

TwitchBronBron (approved):
The latest changes feel much better, protects us a bit from breaking changes. Thanks!

PR #1159, opened 2024-05-08
Adds `roSgNodeContentNode` fields

markwpearce:
Also:

- adds missing fields to `Video` node
- fixes descriptions when the first line is `Available Since Roku OS ..."
- better infers types coming from Roku docs, especially when a field is another Component type


scripts/create-roku-types-data-json.ts
```diff
@@ -0,0 +1,36 @@
+import * as fsExtra from 'fs-extra';
```
TwitchBronBron, line 1: What's the point of this entire script? I feel like you could just call if `(!fsExtra.existsSync(dataJsonPath)) {   fsExtra.outputJSONSync(dataJsonPath, {}); }` at the beginning of the `scrape-roku-docs.ts` script.

TwitchBronBron (changes requested):

scripts/create-roku-types-data-json.ts
markwpearce: scrape-roku-docs.ts will not run if `./src/roku-types/data.json` does not exist, because Brighterscript itself needs that file to exist, and brighterscript is required in the doc scraper (for parsing, types, etc)

in `roku-types/index.ts` there's the line:
```
import data = require('./data.json');
```

markwpearce: this is mostly a convenience script, so if the scrapper script failed last time or was halted through the debugger, I can re-run it without doing another step

markwpearce: I'll get rid of it, and just not delete `data.json` at the beginning of the  scraper script

TwitchBronBron: We could also change `roku-types/index.ts` to load it a different way. Perhaps something like this?

```typescript
import * as fsExtra from 'fs-extra';
let rokuTypes = {} as Whatever;
if(fsExtra.pathExistsSync('./path/to/roku-types.json')){
    rokuTypes = fsExtra.readFileSync('...path');
}
export rokuTypes;
```

anyway, just a thought. not deleting the file would probably work too. your call.

TwitchBronBron (approved):

PR #1169, opened 2024-05-09
Move function calls to separate diagnostic

TwitchBronBron:
Moves "cannot find name" diagnostics for function calls into their own diagnostic code. This will help dev teams more slowly migrate to "best practices" in their projects by ignoring this specific diagnostic when functions are imported into only some of the scopes the script is used in. 

**Before:**
image

**After:**
image

### DIFF
--- a/src/DiagnosticMessages.ts
+++ b/src/DiagnosticMessages.ts
@@ -724,6 +724,20 @@ export let DiagnosticMessages = {
         message: `Optional chaining may not be used in the left-hand side of an assignment`,
         code: 1139,
         severity: DiagnosticSeverity.Error
+    }),
+    /**
+     *
+     * @param name for function calls where we can't find the name of the function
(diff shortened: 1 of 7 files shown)


src/DiagnosticMessages.ts
```diff
@@ -724,6 +724,20 @@ export let DiagnosticMessages = {
...
         severity: DiagnosticSeverity.Error
+    }),
+    /**
+     *
+     * @param name for function calls where we can't find the name of the function
+     * @param fullName if a namespaced name, this is the full name `alpha.beta.charlie`, otherwise it's the same as `name`
+     */
+    cannotFindFunction: (name: string, fullName?: string) => ({
+        message: `Cannot find function '${name}'`,
+        code: 1140,
```
markwpearce, line 735: Can we make this code 1149 - in v1 we have up to 1148 so far

chrisdp (approved):

Issue #1167, opened 2024-05-09
Missing parameter in completions


TwitchBronBron:
![image](https://github.com/rokucommunity/brighterscript/assets/2544493/6cf10c71-4e3a-460b-97bf-8c9df6979693)

PR #1170, opened 2024-05-09
Fixes union type issues with unknown members

markwpearce:
- Adds extra detail to `cannotFindName` diagnostic to say what it cannot find the name in
- If a union type is in a Brighterscript file, it will act like a Dynamic type
- Removes functionExpression body symbol table, in favor of single symbol table that also includes the params. (I think this solves #1167)

### DIFF
--- a/src/bscPlugin/validation/ScopeValidator.ts
+++ b/src/bscPlugin/validation/ScopeValidator.ts
@@ -5,7 +5,7 @@ import { Cache } from '../../Cache';
 import type { DiagnosticInfo } from '../../DiagnosticMessages';
 import { DiagnosticMessages } from '../../DiagnosticMessages';
 import type { BrsFile } from '../../files/BrsFile';
-import type { BsDiagnostic, CallableContainer, ExtraSymbolData, FileReference, OnScopeValidateEvent, TypeChainEntry, TypeCompatibilityData } from '../../interfaces';
+import type { BsDiagnostic, CallableContainer, ExtraSymbolData, FileReference, GetTypeOptions, OnScopeValidateEvent, TypeChainEntry, TypeChainProcessResult, TypeCompatibilityData } from '../../interfaces';
 import { SymbolTypeFlag } from '../../SymbolTypeFlag';
 import type { AssignmentStatement, ClassStatement, DottedSetStatement, EnumStatement, NamespaceStatement, ReturnStatement } from '../../parser/Statement';
 import util from '../../util';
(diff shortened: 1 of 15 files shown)


src/parser/Expression.ts
```diff
@@ -225,14 +225,6 @@ export class FunctionExpression extends Expression implements TypedefProvider {
...
         this.returnTypeExpression = options.returnTypeExpression;
 
-        //if there's a body, and it doesn't have a SymbolTable, assign one
-        if (this.body) {
-            if (!this.body.symbolTable) {
-                this.body.symbolTable = new SymbolTable(`Block`, () => this.getSymbolTable());
-            } else {
-                this.body.symbolTable.pushParentProvider(() => this.getSymbolTable());
-            }
-        }
```
TwitchBronBron, line 235: So for anonymous functions, they don't have a parent FunctionStatement. I don't think we can eliminate the symbolTable in here. For example:

```brightscript
sub test()
    doSomething = sub()
        'i have no immediate parent FunctionStatement
    end sub
end sub
```

src/bscPlugin/validation/ScopeValidator.ts
```diff
@@ -1358,6 +1360,34 @@ export class ScopeValidator {
         }
     }
 
+
+    private getNodeTypeWrapper(file: BrsFile, node: AstNode, getTypeOpts: GetTypeOptions) {
```
TwitchBronBron, line 1364: Add a jsdoc comment explaining what this is for.

src/interfaces.ts
```diff
@@ -902,10 +903,18 @@ export interface TypeChainProcessResult {
      * The name of the last item in the chain, OR the first unresolved item in the chain
      */
     itemName: string;
+    /**
+     * The type kind of  the item of `itemName`
```
TwitchBronBron, line 907:
```suggestion
     * The TypeKind of the item of `itemName`
```

src/interfaces.ts
```diff
@@ -902,10 +903,18 @@ export interface TypeChainProcessResult {
...
+     */
+    itemTypeKind: BscTypeKind | string;
     /**
      * The name of the parent of the item of `itemName`
      */
     itemParentTypeName: string;
+    /**
+     * The type kind of  the parent of the item of `itemName`
```
TwitchBronBron, line 915:
```suggestion
     * The TypeKind of the parent of the item of `itemName`
```

TwitchBronBron (changes requested):

src/parser/Expression.ts
TwitchBronBron: Nevermind. I misunderstood what was happening here. FunctionExpression is still getting a symbolTable, we've just eliminated the one in `Body` in this specific situation

TwitchBronBron (approved):

PR #1175, opened 2024-05-10
Fix several lsp crashes for v1

TwitchBronBron:
Fix several crashes in the language server related to missing range or AST properties.

### DIFF
--- a/src/bscPlugin/completions/CompletionsProcessor.ts
+++ b/src/bscPlugin/completions/CompletionsProcessor.ts
@@ -190,16 +190,16 @@ export class CompletionsProcessor {
             expression = file.getClosestExpression(this.event.position);
         }
 
-        if (!expression) {
-            return [];
-        }
-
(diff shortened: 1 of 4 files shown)


markwpearce (approved):

src/bscPlugin/completions/CompletionsProcessor.ts
```diff
@@ -190,16 +190,16 @@ export class CompletionsProcessor {
...
-            return [];
-        }
-
         if (isFunctionExpression(expression)) {
             // if completion is the last character of the function, use the completions of the body of the function
             expression = expression.body;
         }
 
-        const tokenBefore = file.getTokenBefore(file.getClosestToken(expression.range.start));
+        if (!expression) {
```
TwitchBronBron, line 198: Moved this after the if statement above, because there was a small chance of expression being `null`, this catches that.

PR #1177, opened 2024-05-10
Merge master into v1

TwitchBronBron:
The master ->release-1.0.0 merge was a bit busy, so this is a dedicated PR to help with reviewing the changes to ensure that everything was done properly. 

**Extra changes:**
- added `astNode` to the TypeChain so we could better support `cannotFindFunction` diagnostics introduced recently in master

TODO:

- [x] fix a few more broken unit tests


src/interfaces.ts
```diff
@@ -877,6 +937,7 @@ export class TypeChainEntry {
         range: Range;
         separatorToken?: Token;
         kind?: AstNodeKind;
```
markwpearce, line 939: If we’re going to add AstNode, we can get rid of range and kind…

markwpearce (changes requested):
Looking forward to getting parity with master again. 
We should cleanup the the TypeChainEntry class now that it has astNode.

PR #1187, opened 2024-05-15
Adds `isInstance` flag

markwpearce:
Adds the `isInstance` flag on symbol extra data for when a symbol represents an instance of a type.

This is useful for walking all variable expressions, and being able to check if that expression is an instance of the type or a reference to the actual type (or class constructor).

eg:

```
class Klass
end class

sub test(k as Klass)  ' k is an instance, Klass is not, but both are of ClassType("klass")
end sub
```


src/bscPlugin/completions/CompletionsProcessor.ts
```diff
@@ -374,14 +374,15 @@ export class CompletionsProcessor {
     private getCompletionKindFromSymbol(symbol: BscSymbol, areMembers = false) {
         const type = symbol?.type;
         const extraData = symbol?.data;
-        const finalTypeNameLower = type?.toString().split('.').pop().toLowerCase();
-        const symbolNameLower = symbol?.name.toLowerCase();
-        let nameMatchesType = symbolNameLower === finalTypeNameLower;
+        const isInstance = extraData?.isInstance;
```
markwpearce, line 377: Example of use of the isInstance flag

TwitchBronBron (approved):

Issue #1186, opened 2024-05-14
Change += and other operators to ExpressionStatement


TwitchBronBron:
Currently expressions like this: `value += 2` are parsed as `AssignmentStatements`. This is strange because the AST ends up with 2 different pointers to the `value` token (in the assignment AND in the `value += value`). The TypeScript AST represents this as an `ExpressionStatement`, which seems to make more sense in this context, so we should consider moving to a similar pattern.

This would be a breaking change.

![image](https://github.com/rokucommunity/brighterscript/assets/2544493/37daa690-362e-413e-8fa9-fb3cc4f6bed0)

PR #1196, opened 2024-05-21
Adds AugmentedAssignmentStatement

markwpearce:
code like this:

```
sub foo(t)
   t += 1
end sub
```

used to be an assignment statement, with a binary Expression... eg:

```
Assignment (t, '=') -> 
     BinaryExpression (t, '+=', 1)
```

which included "t" twice

This is changed to a single `AugmentedAssignmentStatement`

```
AugmentedAssignment (t, '+=', 1)
```

Additionally, this PR makes sure the parser only builds Assignments (including typed Assignments) when it is allowed to.

I also added tests for `AugmentedAssignmentStatement` and `IncrementStatement`

Addresses #1186


src/bscPlugin/validation/ScopeValidator.ts
```diff
@@ -150,7 +150,12 @@ export class ScopeValidator {
                             nameRange: assignStmt.tokens.name.range
                         });
                     },
-                    NewExpression: (newExpr) => {
+                    AugmentedAssignmentStatement: (binaryExpr) => {
+                        this.validateBinaryExpression(file, binaryExpr);
+                    },
+                    IncrementStatement: (stmt) => {
+                        this.validateIncrementStatement(file, stmt);
+                    }, NewExpression: (newExpr) => {
```
TwitchBronBron, line 158:
```suggestion
                    },
                    NewExpression: (newExpr) => {
```

src/bscPlugin/validation/ScopeValidator.ts
```diff
@@ -672,6 +681,33 @@ export class ScopeValidator {
...
+        let rightType = this.getNodeTypeWrapper(file, incStmt.value, getTypeOpts);
+
+        if (!rightType.isResolvable()) {
+            // Can not find the type. error handled elsewhere
+            return;
+        }
+
+        if (isUnionType(rightType)) {
+            // TODO: it is possible to validate based on innerTypes, but more complicated
+            // Because you need to verify each combination of types
```
TwitchBronBron, line 696:
```suggestion
            // because you need to verify each combination of types
```

src/parser/Parser.ts
```diff
@@ -992,36 +994,61 @@ export class Parser {
...
-            ...allowedAssignmentOperators
+            DiagnosticMessages.expectedOperatorAfterIdentifier([TokenKind.Equal], name.text),
+            ...[TokenKind.Equal]
         );
         let value = this.expression();
 
-        let result: AssignmentStatement;
+        let result = new AssignmentStatement({ equals: operator, name: name, value: value, as: asToken, typeExpression: typeExpression });
+
+        /*AssignmentStatement | AugmentedAssignmentStatement;
```
TwitchBronBron, line 1014: Do we still need this commented-out code?

src/parser/Parser.ts
```diff
@@ -992,36 +994,61 @@ export class Parser {
...
+                operator: operator,
+                value: value
             });
         }
+*/
+        return result;
+    }
+
+    private augmentedAssignment(): AugmentedAssignmentStatement {
+        let name = this.advance() as Identifier;
```
TwitchBronBron, line 1030: What about dotted get and indexed get? Shouldn't it be `this.expression()`?

```
array[0] += 1
alpha.beta.charlie += 2
```

TwitchBronBron (changes requested):

src/parser/Parser.ts
markwpearce: I'll change it so all uses of `+=`, '-=`, etc... will be Augmented Assignments

markwpearce: done

TwitchBronBron (approved):

Issue #1204, opened 2024-05-27
bsconfig.schema.json is not type valid against jsonschema draft-07 target


bartvandenende-wm:
## Summary

bsconfig.schema.json is not (strict) type valid against jsonschema draft-07 target

## Repro steps

1. Install `npm install -g ajv-cli`
2. comile schema with `ajv compile -s 'bsconfig.schema.json' `

  **Expected result:** 
no compile errors

  **Actual result:** 
  
  ```
  schema bsconfig.schema.json is invalid
error: strict mode: unknown keyword: "deprecationMessage"
  ```
```
strict mode: missing type "array" for keyword "items" at "bsconfig.schema.json#/properties/files" (strictTypes)
```

## Details

Proposal would be to:
- remove the `"deprecationMessage"` attribute from "ignoreErrorCodes"
- add `"type": "array"` to the "files".

PR #1205, opened 2024-05-27
fix: conform bsconfig.schema.json to strict types

bartvandenende-wm:
## Summary

bsconfig.schema.json is not (strict) type valid against jsonschema draft-07 target

fixes https://github.com/rokucommunity/brighterscript/issues/1204

## Details

- remove the `"deprecationMessage"` attribute from "ignoreErrorCodes"
- add `"type": "array"` to the "files".

## Validation steps

1. Install `npm install -g ajv-cli`
2. compile schema with `ajv compile -s bsconfig.schema.json --allow-union-types=true`
```
schema bsconfig.schema.json is valid
```


TwitchBronBron:
This is great! Would you be willing to add a unit test that runs the validation steps? That way we can ensure it remains in compliance with the json schema in the future.

bartvandenende-wm:
> This is great! Would you be willing to add a unit test that runs the validation steps? That way we can ensure it remains in compliance with the json schema in the future.

I could, but that likely would require us to move the `bsconfig.schema.json` into the TS `src` folder for the json to be importable for the unit test, which would be a breaking change.

TwitchBronBron:
So vscode supports the "deprecationMessage" option, and it even shows up as deprecated. I'm not sure we should remove that property as it provides useful error messaging in the editor. I'm sure JSON schema allows for having other properties? Or perhaps ajv-cli would have an option for ignoring that error?

image

TwitchBronBron (approved):

PR #1210, opened 2024-06-03
Document the deprecated `ifMessagePort.setPort()` method.

TwitchBronBron:
Adds info about the deprecated `ifMessagePort.setPort()` method.

image

TODO - it would be nice to show strikethrough on the deprecated method somehow in the editor. 

Also, based on [this discussion](https://rokudevelopers.slack.com/archives/CF419UU80/p1715179435039329), should this actually be added to the roUrlTransfer object instead of ifUrlTransfer? Those don't currently support having their own methods in our scraper, but it could?


markwpearce (approved):

markwpearce:
Yes, the deprecated `setPort` method should only be on the object itself. 
The scrapper/BuiltInInterfaceAdder might not support that currently, but it’s not that big a deal to add

PR #1211, opened 2024-06-04
Refines ScopeValidator Diagnostics to use more specific Diagnostic tags

### LINKED ISSUE #1209: Diagnostics show up and disappear in LSP mode depending on which file was edited last
Found an issue with diagnostics going missing when editing a file.

```typescript
it.only('lsp bug', () => {
    program.options.autoImportComponentScript = true;
    program.setFile('components/playerInterfaces.bs', `
        interface MediaObject
            optional url as string
        end interface
    `);
    program.setFile('components/player.xml', `
        <component name="Player" extends="Group">
        </component>
    `);
    program.setFile('components/player.bs', `
        import "playerInterfaces.bs"
        import "playerUtils.bs"
        sub test()
            media = {} as MediaObject
            print media.missingBool1
        end sub
    `);
    program.setFile('components/playerUtils.bs', `
        import "playerInterfaces.bs"
        function test1(media as MediaObject) as boolean
            print media.missingBool2
        end function
    `);
    program.validate();
    //we have both diagnostics
    expectDiagnostics(program, [
        DiagnosticMessages.cannotFindName('missingBool1', undefined, 'MediaObject').message,
        DiagnosticMessages.cannotFindName('missingBool2', undefined, 'MediaObject').message
    ]);

    //add the last file again thus "opening" it.
    program.setFile('components/playerUtils.bs', `
        import "playerInterfaces.bs"
        function test2(media as MediaObject) as boolean
            print media.missingBool2
        end function
    `);
    program.validate();
    //we STILL have both diagnostics
    expectDiagnostics(program, [
        DiagnosticMessages.cannotFindName('missingBool1', undefined, 'MediaObject').message,
        DiagnosticMessages.cannotFindName('missingBool2', undefined, 'MediaObject').message
    ]);
});
```

markwpearce:
Fixes an issue where more diagnostics would be cleared on file changes than should have been.

Fixes #1209


src/bscPlugin/validation/ScopeValidator.ts
```diff
@@ -60,9 +60,10 @@ export class ScopeValidator {
         if (this.event.program.globalScope === this.event.scope) {
             return;
         }
-        this.event.program.diagnostics.clearByFilter({ scope: this.event.scope, tag: ScopeValidatorDiagnosticTag });
+        this.event.program.diagnostics.clearByFilter({ scope: this.event.scope, segment: null, tag: ScopeValidatorDiagnosticTag });
```
TwitchBronBron, line 63: Setting these to null is a bit unintuitive, especially combined with the `!== undefined` above. Is this intentional? Throughout the codebase, we've tried to basically treat `null` and `undefined` as the same thing. 

What does setting it to `null` accomplish? Does it say "don't match on segements"? Or "mach all segments"? I'm wondering if there's a more clear way to handle this, since plugin authors will also be calling these functions.

src/DiagnosticManager.ts
```diff
@@ -197,14 +197,15 @@ export class DiagnosticManager {
...
 
         const needToMatch = {
-            tag: !!filter.tag,
-            scope: !!filter.scope,
-            file: !!filter.file,
-            segment: !!filter.segment
+            tag: !!(filter.tag),
+            scope: !!(filter.scope),
+            file: !!(filter.file),
+            segment: !!(filter.segment)
```
TwitchBronBron, line 208: Are the parentheses needed now?

markwpearce:
Changed tack here... Instead of being laisez-faire about the diagnostics clearing, instead i've changed it to be very specific to the what is cleared.. so each validation function can decide best on how to clear existing diagnostics

TwitchBronBron (approved):

src/DiagnosticManager.ts
markwpearce: nope .. i'll remove them

PR #1223, opened 2024-06-07
Fixes issue with setting `roSGNodeTargetSet.targetRects`

### LINKED ISSUE #1219: Issue with setting fields that are of type `rect2DArray`
image

image

AssociativeArrays with {x:float, y:float, width:float, height:float} *should* be assignable to "rect2d" types...

Specifically a problem with `roSGNodeTargetSet.targetRects`

markwpearce:
Fixes #1219

Reason this was a problem is because the scraper did not understand `array of rectangles` as a type.

I wish Roku would use actual types here! The interface type should have been `rect2dArray` ... oh well..

I also updated the diagnostic message for type incompatibility to be able to add a override for the extended message which is useful if assigning Arrays of different default types, and the inner type is incompatible.

Example:
```brighterscript
targetSet = createObject("roSGNode", "TargetSet")
targets = ["hello", "world"]
targetSet.targetRects = targets ' targetRects is an array of rectangles, which are AA's with x,y,width,height
```

Before

```
"Type 'Array<string>' is not compatible with type 'Array<roAssociativeArray>'
    Type 'Array<string>' is missing the following members: height, width, x, y, ifAssociativeArray, AddReplace, Append, Clear, Count, Delete, DoesExist, Items, Keys, Lookup, LookupCI, ...and 5 more"
```

Now:

```
"Type 'Array<string>' is not compatible with type 'Array<roAssociativeArray>'
    Type 'string' is missing the following members of type 'roAssociativeArray': height, width, x, y, ifAssociativeArray, AddReplace, Append, Clear, Count, Delete, DoesExist, Items, Keys, Lookup, LookupCI, ...and 5 more"
```

`

### DIFF
--- a/src/DiagnosticMessages.ts
+++ b/src/DiagnosticMessages.ts
@@ -815,8 +815,11 @@ export const defaultMaximumTruncationLength = 160;
 
 export function typeCompatibilityMessage(actualTypeString: string, expectedTypeString: string, data: TypeCompatibilityData) {
     let message = '';
+    actualTypeString = data?.actualType?.toString() ?? actualTypeString;
+    expectedTypeString = data?.expectedType?.toString() ?? expectedTypeString;
+
     if (data?.missingFields?.length > 0) {
-        message = `\n    Type '${actualTypeString}' is missing the following members: ` + util.truncate({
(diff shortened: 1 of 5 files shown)


TwitchBronBron:
Nice!

PR #1224, opened 2024-06-10
Remove block starting range

markwpearce:
Removes the `startingRange` property on Blocks.

If a block is not empty, the range of a Block is the range of its statements

If the block is empty, the Block looks at its parent to see what the range should be, with specific logic for all the different types of nodes that can contain a block.

### DIFF
--- a/src/parser/Statement.ts
+++ b/src/parser/Statement.ts
@@ -273,7 +273,7 @@ export class Block extends Statement {
             lastBitBefore = util.createBoundingRange(
                 this.parent.tokens.functionType,
                 this.parent.tokens.leftParen,
-                ...this.parent.parameters,
+                ...(this.parent.parameters ?? [])
                 this.parent.tokens.rightParen,
                 this.parent.tokens.as,
                 this.parent.returnTypeExpression
--- a/src/parser/Statement.ts
+++ b/src/parser/Statement.ts
@@ -259,7 +259,6 @@ export class Block extends Statement {
     }
 
     public readonly statements: Statement[];
-    public readonly startingRange?: Range;
 
     public readonly kind = AstNodeKind.Block;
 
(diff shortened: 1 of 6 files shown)


src/parser/Statement.ts
```diff
@@ -253,23 +253,105 @@ export class AugmentedAssignmentStatement extends Statement {
...
 
     public readonly statements: Statement[];
-    public readonly startingRange?: Range;
 
     public readonly kind = AstNodeKind.Block;
 
-    public readonly range: Range | undefined;
+    get range(): Range {
+        if (this.statements.length > 0) {
+            return util.createBoundingRange(...(this.statements));
```
TwitchBronBron, line 267: Don't need these extra parens
```suggestion
            return util.createBoundingRange(...this.statements);
```

src/parser/Statement.ts
```diff
@@ -253,23 +253,105 @@ export class AugmentedAssignmentStatement extends Statement {
...
+        }
+        let lastBitBefore: Position;
+        let firstBitAfter: Position;
+
+        if (isFunctionExpression(this.parent)) {
+            lastBitBefore = util.createBoundingRange(
+                this.parent.tokens.functionType,
+                this.parent.tokens.leftParen,
+                ...this.parent.parameters,
```
TwitchBronBron, line 276: Is there ever a situation where .parameters is null? If so, maybe do something more like this?

```suggestion
                ...(this.parent.parameters ?? [])
```

src/util.ts
```diff
@@ -1088,7 +1088,12 @@ export class Util {
     /**
      * Create a `Range` from two `Position`s
      */
-    public createRangeFromPositions(startPosition: Position, endPosition: Position): Range {
+    public createRangeFromPositions(startPosition: Position, endPosition: Position): Range | undefined {
+        startPosition = startPosition ?? endPosition;
```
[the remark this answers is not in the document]
TwitchBronBron: Could you add a unit test to ensure this keeps working as expected?

TwitchBronBron (changes requested):
Just a few small tweaks. Looks great!

PR #1227, opened 2024-06-11
Removed `isInline` property from `IfStatement` 

markwpearce:
- It was changed to a getter, so that proper diagnostic could still be added if there was a problem.
- Added tests for inline conditional statements
- Added support for nested inline conditionals. (People must have a deathwish!)

### DIFF
--- a/src/parser/Statement.ts
+++ b/src/parser/Statement.ts
@@ -634,8 +634,8 @@ export class IfStatement extends Statement {
             ...(this.tokens.endIf?.leadingTrivia ?? [])
         ];
 
-        const hasNewlineOrColon = allLeadingTrivia.find(t => t.kind === TokenKind.Newline);
-        return !hasNewlineOrColon;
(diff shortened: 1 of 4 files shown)


src/parser/Statement.ts
```diff
@@ -619,12 +621,23 @@ export class IfStatement extends Statement {
     public readonly condition: Expression;
     public readonly thenBranch: Block;
     public readonly elseBranch?: IfStatement | Block;
-    public readonly isInline?: boolean;
 
     public readonly kind = AstNodeKind.IfStatement;
 
     public readonly range: Range | undefined;
 
+    get isInline() {
```
markwpearce, line 629: This could potentially be a `utils` function that is more general - i.e. it could work for all AstNodes, by iterating all tokens and properties that are also AstNodes

TwitchBronBron (approved):

PR #1236, opened 2024-06-17
Convenience export Location

markwpearce:
Export `Location` so it can be used directly in plugins
Also, ran `npm audit fix`


TwitchBronBron:
Nice

PR #1244, opened 2024-07-08
Sets a hard limit of number of scopes to check for completions

### LINKED ISSUE #1240: Performance issues with `CompletionsProcessor`
To get completions in v1,  the `CompletionsProcessor` currently does a scope linking and symbol table lookup for EVERY scope, then checks to see if any resultant completions are valid for every scope. 

For a file in many scopes, this can take a bit of time (sometimes greater than 500 ms).

There are probably ways to speed this up, perhaps by creating a temporary scope that is the intersection of all scopes for a file, then linking and doing symbol table lookups once for this temporary scope.

markwpearce:
Sets a hard limit (of 5) of the number of scopes to check for a file when processing completions.

In Jellyfin, completion processing went from ~500ms to ~9ms

Fixes #1240

Based off code found in this commit: https://github.com/rokucommunity/brighterscript/commit/bac0ae90bcc00df2689371e451983c33478c4695

### DIFF
--- a/src/bscPlugin/completions/CompletionsProcessor.ts
+++ b/src/bscPlugin/completions/CompletionsProcessor.ts
@@ -92,7 +92,7 @@ export class CompletionsProcessor {
         // always include the source scope if applicable to this file
         let sourceScope = scopesForFile.find(x => x.name === 'source');
         if (sourceScope && !scopesToProcess.includes(sourceScope)) {
-            //replace the first scope with the source scope so we always process exactly 3 scopes
+            //replace the first scope with the source scope so we process a consistent number of scopes if possible
             scopesToProcess[0] = sourceScope;
         }
         return scopesToProcess;
--- a/src/bscPlugin/completions/CompletionsProcessor.ts
+++ b/src/bscPlugin/completions/CompletionsProcessor.ts
@@ -24,6 +24,9 @@ import { createIdentifier } from '../../astUtils/creators';
 import type { FunctionExpression } from '../../parser/Expression';
 import { LogLevel } from '../../Logger';
 
+
+const SCOPES_FOR_COMPLETION = 5;
+
 export class CompletionsProcessor {
     constructor(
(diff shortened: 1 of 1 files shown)


src/bscPlugin/completions/CompletionsProcessor.ts
```diff
@@ -67,13 +69,35 @@ export class CompletionsProcessor {
...
+        }
+
+        // Only process the first few scopes. This might result in missing completions,
+        // but it's better than wasting TONS of cycles building essentially the same completions over and over
+        const scopesToProcess = scopesForFile.slice(0, SCOPES_FOR_COMPLETION);
+
+        // always include the source scope if applicable to this file
+        let sourceScope = scopesForFile.find(x => x.name === 'source');
+        if (sourceScope && !scopesToProcess.includes(sourceScope)) {
+            //replace the first scope with the source scope so we always process exactly 3 scopes
```
TwitchBronBron, line 95:
```suggestion
            //replace the first scope with the source scope so we process a consistent number of scopes if possible
```

src/bscPlugin/completions/CompletionsProcessor.ts
```diff
@@ -24,6 +24,9 @@ import { createIdentifier } from '../../astUtils/creators';
 import type { FunctionExpression } from '../../parser/Expression';
 import { LogLevel } from '../../Logger';
 
+
+const SCOPES_FOR_COMPLETION = 5;
```
TwitchBronBron, line 28: Is there a reason you went with 5 instead of 3 like we have in https://github.com/rokucommunity/brighterscript/commit/bac0ae90bcc00df2689371e451983c33478c4695 ?

markwpearce: no reason! I'll change it back

TwitchBronBron (approved):

Issue #1155, opened 2024-05-03
Remove `.file` prop from `BsDiagnostic`


TwitchBronBron:
When #1136 lands, we should consider removing `file` from the `BsDiagnostic` interface. It doesn't feel right being attached to a diagnostic, and we could easily update `DiagnosticsManager` to support returning a new object that includes the scope/file information (i.e. the file) for a diagnostic) if we need that information.

This would be a breaking change, so likely should be done as part of v1.

markwpearce:
Not really sure how this works. How do we associate with a diagnostic with a file? Do we use the `Location` property on the `Diagnostic.relatedInformation`?

TwitchBronBron:
All diagnostics flow through the diagnosticManager now. So `file` would be a required property when registering a diagnostic. Something like this:

```ts
program.diagnostics.register({
    ...DiagnosticMessages.cannotFindName('someVar'),
}, {
    file: theFile,
});
```

Then if we need to know what file that diagnostic comes from, that would be an attribute of the return value from the diagnostic manager. Perhaps something like this?:

```ts
diagnosticManager.getDiagnosticsByFile(); // Map<BscFile, Diagnostic[]>; 
```

markwpearce:
We decided that we should replace `.file` and `.range` with `.location` to mirror the changes in the AST Nodes.

PR #1248, opened 2024-07-12
Remove `file` prop from `BsDiagnostic`

markwpearce:
All diagnostics internal to Brighterscript now use a `location: Location` prop instead of `file: BscFile`

This has a few good points:
 - Diagnostics no longer have a reference to a file, so no circular references to remove
 - Aligns better with the change to use `Location` for Tokens and AstNodes

One thing to note is that sometimes we need to back-convert a URI to a path for finding the files that a diagnostic applies to. 
There might be some inconsistencies between Operating Systems (Mac vs. Windows, etc) that *should* be dealt with the `URI` namespace from `vscode-uri`, but I'm not entirely sure.

Addresses: #1155

### DIFF
--- a/src/CrossScopeValidator.ts
+++ b/src/CrossScopeValidator.ts
@@ -16,7 +16,6 @@ import type { BscType } from './types/BscType';
 import type { BscFile } from './files/BscFile';
 import type { ClassStatement, ConstStatement, EnumMemberStatement, EnumStatement, InterfaceStatement, NamespaceStatement } from './parser/Statement';
 import { ParseMode } from './parser/Parser';
-import { URI } from 'vscode-uri';
 import { globalFile } from './globalCallables';
 import type { DottedGetExpression, VariableExpression } from './parser/Expression';
 import type { InheritableType } from './types';
@@ -582,15 +581,11 @@ export class CrossScopeValidator {
 
                                 const relatedInformation = thatNameRange ? [{
                                     message: `${thatNodeKindName} declared here`,
-                                    location: util.createLocationFromRange(
-                                        URI.file(otherDupe.file?.srcPath).toString(),
-                                        thatNameRange
-                                    )
+                                    location: util.createLocationFromFileRange(otherDupe.file, thatNameRange)
                                 }] : undefined;
                                 this.program.diagnostics.register({
                                     ...DiagnosticMessages.nameCollision(thisNodeKindName, thatNodeKindName, thatName),
-                                    file: dupe.file,
-                                    range: thisNameRange,
+                                    location: util.createLocationFromFileRange(dupe.file, thisNameRange),
                                     relatedInformation: relatedInformation
                                 }, {
                                     scope: scope,
@@ -618,16 +613,14 @@ export class CrossScopeValidator {
         }
 
         // Check each file for missing symbols - if they are missing in SOME scopes, add diagnostic
-        for (let [file, missingSymbolPerFile] of missingSymbolInScope.entries()) {
+        for (let [_, missingSymbolPerFile] of missingSymbolInScope.entries()) {
             for (const [symbol, scopeList] of missingSymbolPerFile) {
                 const typeChainResult = util.processTypeChain(symbol.typeChain);
 
                 for (const scope of scopeList) {
                     this.program.diagnostics.register({
-
                         ...this.getCannotFindDiagnostic(scope, typeChainResult),
-                        file: file,
-                        range: typeChainResult.location?.range
+                        location: typeChainResult.location
                     }, {
                         scope: scope,
                         tags: [CrossScopeValidatorDiagnosticTag]
@@ -644,8 +637,7 @@ export class CrossScopeValidator {
                 const scopeListName = [...incompatibleScopes.values()].map(s => s.name).join(', ');
                 this.program.diagnostics.register({
                     ...DiagnosticMessages.incompatibleSymbolDefinition(typeChainResult.fullChainName, scopeListName),
-                    file: symbol.file,
-                    range: typeChainResult.location?.range
+                    location: typeChainResult.location
                 }, {
                     tags: [CrossScopeValidatorDiagnosticTag]
                 });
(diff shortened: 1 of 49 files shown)


markwpearce:
@TwitchBronBron This fails in windows (I *think* due to somewhere the URI or `srcPath` has an uppercase drive letter?) ..
when you get a chance, can you look into that?

src/CrossScopeValidator.ts
```diff
@@ -618,16 +613,14 @@ export class CrossScopeValidator {
         }
 
         // Check each file for missing symbols - if they are missing in SOME scopes, add diagnostic
-        for (let [file, missingSymbolPerFile] of missingSymbolInScope.entries()) {
+        for (let [_, missingSymbolPerFile] of missingSymbolInScope.entries()) {
```
TwitchBronBron, line 616: Could we use `.values()` here now instead, since we aren't using the `file` item anymore? Should `missingSymbolInScope` structure be changed if we're not using file anymore? (I don't have enough context on what's being done here)

src/DiagnosticFilterer.ts
```diff
@@ -100,35 +101,36 @@ export class DiagnosticFilterer {
...
         }
     }
 
-    private filterFile(filter: NormalizedFilter, filePath: string) {
+    private filterFile(filter: NormalizedFilter, fileUri: string) {
         //if the filter is negative, we're turning diagnostics on
         //if the filter is not negative we're turning diagnostics off
         const isSuppressing = !filter.isNegative;
-
+        const lowerFileUri = fileUri.toLowerCase();
```
TwitchBronBron, line 124: This _probably_ isn't a problem, but it feels strange to call .toLowerCase() on a URI. For example, consider this URI from windows:
`"file:///c%3A/projects/roku/brighterscript1/rootdir/source/common1.brs"`

The `%3A` is some type of hex code representing a colon (for `C:/projects...`). By running .`toLowerCase()`, that `%3a` represents a different value. It feels expensive to parse the URI to get the file path in this situation, so I'm not sure we need to change it, but it just stood out as an interesting problem.

src/LanguageServer.ts
```diff
@@ -1305,12 +1305,12 @@ export class LanguageServer {
             //Get only the changes to diagnostics since the last time we sent them to the client
             const patch = this.diagnosticCollection.getPatch(this.projects);
 
-            for (let filePath in patch) {
-                const uri = URI.file(filePath).toString();
-                const diagnostics = patch[filePath].map(d => util.toDiagnostic(d, uri));
+            for (let fileUri in patch) {
+                // const uri = URI.file(filePath).toString();
```
TwitchBronBron, line 1309: Can this be deleted?
```suggestion
```

src/ProgramBuilder.ts
```diff
@@ -300,41 +294,42 @@ export class ProgramBuilder {
...
                 return (
-                    (a.range?.start.line ?? -1) - (b.range?.start.line ?? -1) ||
-                    (a.range?.start.character ?? -1) - (b.range?.start.character ?? -1)
+                    (a.location?.range?.start.line ?? -1) - (b.location?.range?.start.line ?? -1) ||
+                    (a.location?.range?.start.character ?? -1) - (b.location?.range?.start.character ?? -1)
                 );
             });
 
-            let filePath = srcPath;
+            let filePath = URI.parse(fileUri).fsPath;
```
TwitchBronBron, line 319: Can we use the `util.uriToPath()` function here instead (for consistency).

src/files/XmlFile.ts
```diff
@@ -259,11 +259,11 @@ export class XmlFile implements BscFile {
             srcPath: this.srcPath,
             destPath: this.destPath
         });
-        const diagnostics = this.parser.diagnostics.map(diagnostic => ({
+        /*const diagnostics = this.parser.diagnostics.map(diagnostic => ({
             ...diagnostic,
             file: this
-        }));
-        this.program?.diagnostics.register(diagnostics);
+        }));*/
```
TwitchBronBron, line 265: Can this block comment be deleted?

src/util.ts
```diff
@@ -744,10 +745,17 @@ export class Util {
...
      */
-    private isUriLike(filePath: string) {
+    public isUriLike(filePath: string) {
         return filePath?.indexOf('file:') === 0;// eslint-disable-line @typescript-eslint/prefer-string-starts-ends-with
     }
 
+    /**
+     * Gets a URI string for a file's srcPath
+     */
+    public getFileUri(file: BscFile) {
```
TwitchBronBron, line 755: Do we need this function? Isn't it almost just as easy to call `util.pathToUri(file?.srcPath)`? Seems like an extra function hop for fairly low benefit.

TwitchBronBron (changes requested):
Just a few comments and suggestions, but overall this looks great!

src/DiagnosticFilterer.ts
TwitchBronBron: I just tested the vscode URI parser, seems it doesn't care about the case of the `a` in `%3A` or `%3a`, so this is most likely fine. 

image

src/Program.ts
```diff
@@ -1038,6 +1035,11 @@ export class Program {
             return this.files[
                 (normalizePath ? util.standardizePath(filePath) : filePath).toLowerCase()
             ] as T;
+        } else if (util.isUriLike(filePath)) {
```
TwitchBronBron, line 1038: I'm wondering if we could eliminate this separate condition, in favor of doing something like:

```typescript
else {
           filePath = util.uriToPath(filePath);
            return this.destMap.get(
                (normalizePath ? util.standardizePath(filePath) : filePath).toLowerCase()
            ) as T;
        }

markwpearce: What if a file's destPath is different from its srcPath and the lookup is the uri of the srcPath?

TwitchBronBron: yeah, good point. I think I looked at the wrong condition. I think it technically should merge with the condition ABOVE this one, not the `else` below. But now that regex is getting a bit busy. Yeah, just leave it as is. we can fix it later if there are performance issues.

Issue #1249, opened 2024-07-15
Add `setMessagePort` / `getMessagePort` to `IfSystemLog`


markwpearce:
Another Roku docs failure.

Very similar to `roSteamSocket` --- the message port getter/setter are implemented in `ifSystemLog` and, and they aren't mentioned in the docs.

Need to update the docs scraper to add them in, like we did here: https://github.com/rokucommunity/brighterscript/blob/release-1.0.0/scripts/scrape-roku-docs.ts#L1242

PR #1250, opened 2024-07-15
Fixes ifSystemLog and ifArraySlice docs

markwpearce:
Addresses #1249 

I also found an issue with `ifArraySlice` that is fixed here.

The problem with ifArraySlice is that the docs were not parsed correctly -- instead of messing with the parsing, I just added it an an override.

This is how it looks on the Roku docs site - I think the double square brackets are the issue:
 
image

### DIFF
--- a/scripts/scrape-roku-docs.ts
+++ b/scripts/scrape-roku-docs.ts
@@ -1268,7 +1268,6 @@ class Runner {
                 },
                 ifsystemlog: {
                     methods: [{
-
                         description: 'Returns the message port (if any) currently associated with the object',
                         name: 'GetMessagePort',
                         params: [],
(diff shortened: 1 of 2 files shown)


TwitchBronBron (approved):
Nice!

scripts/scrape-roku-docs.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable import/no-extraneous-dependencies */
```
markwpearce, line 1: Added this because te linter was complaining that these dependencies should be `dependencies` instead of `devDependencies`

TwitchBronBron: yeah that's fine, this script is only run in dev mode anyway...

PR #1281, opened 2024-08-13
Ast node clone

TwitchBronBron:
Add `.clone()` method to AstNode and BscType objects. This will make for much easier duplication of template code for plugins. 

- [x] add `.clone()` method to `Statement`s
- [x] add `.clone()` method to `Expression`s
- [x] add `.clone()` method to `BscType`s
- [x] add unit tests


TwitchBronBron:
Whew! That was quite a lot of work adding the tests. I found a significant number of bugs and crashes due to adding the tests, so I'm very glad I added them.

Issue #1285, opened 2024-08-15
Merge `ExitWhileStatement` and `ExitForStatement`


TwitchBronBron:
I'm not sure why there are two separate statements here, for `exit while` and `exit for`. I think `exit` on its own is also _technically_ valid syntax.

We should merge them into a single `ExitStatement` that optionally takes a tokenKind for the thing that comes after it.

markwpearce:
`exit` on its own is a syntax error:

```
  for i = 1 to 10
    if i mod 4 = 0
      exit
    end if
    print i
  end for
```

produces:

```
Syntax Error. (compile error &h02 (ERR_SYNTAX): syntax error ) in main.main(12)
=================================================================
An error occurred while attempting to compile the application's components:
-------> Compilation Failed.
pkg:/source/main.brs(NaN)
Syntax Error. (compile error &h02 (ERR_SYNTAX): syntax error ) in main.main(12)
```

TwitchBronBron:
Regardless of the standalone `exit` not being valid, is there any reason to have separate `exit for` and `exit while` statements? We have a single `ContinueStatement`, I think `exit <something>` makes sense to be a single statement type as well.

https://github.com/rokucommunity/brighterscript/blob/353f4638c8a8eed7aa003b1db0438a84e4cbc560/src/parser/Statement.ts#L3601

markwpearce:
Good point... i'll merge them

PR #1289, opened 2024-08-20
Merges `ExitForStatement` and `ExitWhileStatement`

markwpearce:
There's a bit of weirdness here because `exitwhile` is an accepted statement, but `exitfor` is not.

Addresses #1285


src/parser/Parser.ts
```diff
@@ -1235,10 +1230,24 @@ export class Parser {
...
+    private exitStatement(): ExitStatement {
+        const exitToken = this.advance();
+        if (exitToken.kind === TokenKind.ExitWhile) {
+            // `exitwhile` is allowed in code, and means `exit while`
+            // use an ExitStatement that is nicer to work with
 
-        return new ExitWhileStatement({ exitWhile: keyword });
+            return new ExitStatement({
+                exit: createToken(TokenKind.Exit, 'exit', exitToken.location),
+                loopType: createToken(TokenKind.While, 'while', exitToken.location)
```
TwitchBronBron, line 1241: We need to more accurately represent the tokens as they were originally defined. So that means grabbing the raw text from the tokens instead of hardcoding `'exit'` and `'while'`, as well as their locations. Something like

```
exit: createToken(TokenKind.Exit, exitToken.substring(0, 4), /* first 4 chars of the range */),
loopType: createToken(TokenKind.While, exitToken.substring(4, 9), /* last 5 chars of the range */)

**Review by TwitchBronBron — CHANGES_REQUESTED** [review]

**Review by TwitchBronBron — APPROVED** [review]

File: src/parser/Parser.ts [hunk 2]
```diff
@@ -1235,10 +1230,40 @@ export class Parser {
...
+                originalStart.character,
+                originalStart.line,
+                originalStart.character + 4);
+            const whileRange = util.createRange(
+                originalStart.line,
+                originalStart.character + 4,
+                originalStart.line,
+                originalStart.character + exitToken.text.length);
+
+            return new ExitStatement({
```
TwitchBronBron, line 1255: OH! One more request. Since we're creating new tokens, we should probably replace the `exitfor` token in the `.tokens` array with these two new tokens, that way external callers have the correct list of tokens after the parse is finished.

TwitchBronBron (changes requested):

PR #1290, opened 2024-08-22
Add support for resolving sourceRoot at time of config load

TwitchBronBron:
Add bsconfig value for resolving the `sourceRoot` at time of loading the config.


markwpearce:
@TwitchBronBron This works really well for our use case :) Thanks!

PR #1293, opened 2024-09-03
Adds support for types declared in comments

TwitchBronBron:
Adds ability to define variable types in comments:
 - `@param {type}` for function params
 - `@return {type}` for function return type
 - `@type {type}` and `@type {type} varName` for variables in functions

### DIFF
--- a/docs/type-checking-without-transpile.md
+++ b/docs/type-checking-without-transpile.md
@@ -1,0 +1,50 @@
+# Typechecking without transpiling
+If you want the benefits of BrighterScript's type system but don't want to use the transpiler, you can use comments to declare variable types to get much richer editor experience and type validation.
+
+## Params
+Declare the type of a specific parameter:
+
+```brighterscript
+'@param {roSGNodeRectangle} node
+function resize(node, width as integer, height as string)
+    'yay, we know that `node` is a `roSGNodeRectangle` type
+    node.width = width
+    node.height = height
+end function
+```
+
+## Return type
+
+Declare the return type of a function
+
+```brighterscript
+'@return {roSGNodeRectangle}
+function createRectangle()
+    return createObject("roSGNode", "Rectangle")
+end function
+
+function test()
+    'yay, we know that `rectangle` is a `roSGNodeRectangle` type
+    rectangle = createRectangle()
+end function
+```
+
+## Inline variable type
+You can override the type of a variable in brs files by starting a comment with `@type` . This will cast the variable below with the given type.
+
+```brighterscript
+' @type {integer}
+runtime = getRuntimeFromServer()
+```
+
+## Function-level variable type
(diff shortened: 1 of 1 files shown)


markwpearce:
Perhaps we should have a `BrightScriptDocParser` that should be used to query the comments for a function... maybe it could be used like:

```
const parsedBsDoc = bsDocParser.parseNode(funcStmt); // this should internally call `util.getNodeDocumentation()`
const docParams = parsedBsDoc.getParams();
const docReturn = parsedBsDoc.getReturns();
```

markwpearce:
We should also support `@returns` as a synonym of `@return` https://jsdoc.app/tags-returns

markwpearce:
the `@type` tag in jsdocs is supposed to declare only the type of the immediate thing https://jsdoc.app/tags-type ... I don't think supporting `' @type {typename} var` in the comments for a function is good because it breaks the usage in jsdocs

You'll have to use it directly above an assignment:

```
function getModel()
    ' @type {roDeviceInfo}
    info = getDeviceInfo()
    return info.getModel()
end function
```

TwitchBronBron:
> the `@type` tag in jsdocs is supposed to declare only the type of the immediate thing https://jsdoc.app/tags-type ... I don't think supporting `' @type {typename} var` in the comments for a function is good because it breaks the usage in jsdocs
> 
> You'll have to use it directly above an assignment:
> 
> ```
> function getModel()
>     ' @type {roDeviceInfo}
>     info = getDeviceInfo()
>     return info.getModel()
> end function
> ```

The problem is that assignment can get quite messy sometimes. Consider this code:

```
if m.top.isTrue
    someDate = getDate()
else
    someDate = m.date2
end if

if m.someProp
    someDate = m.someProp.date
end if
```

Which assignment ? Unfortunately it would probably need to be above the assignment in _both_ the `if` and `else` blocks. 
In javascript you cannot assign to a variable without first declaring it (var | let |const). However, in brs, you just willy-nilly create variables by using them in assignments.

I modeled this @type after the `@var` comment in phpdoc.  https://docs.phpdoc.org/guide/references/phpdoc/tags/var.html

Would using `@var` feel better than `@type` to you.

markwpearce:
regarding 

```
if m.top.isTrue
    someDate = getDate()
else
    someDate = m.date2
end if

if m.someProp
    someDate = m.someProp.date
end if
```

This is a little awkward, because we have no way of saying a variable is of a certain type and it is that type for the whole function.

if `getDate()` returns `dynamic` , then `someDate` will be `string or dynamic` which resolves to `dynamic`

markwpearce:
image

markwpearce:
With the latest commit, this supports union types in `@param` tags  too.

markwpearce:
image

Refined diagnostic so the underline is under the type in a comment

PR #1299, opened 2024-09-10
Adds diagnostic if no return statement found in function

markwpearce:
image

If a function explicitly says it will return something, ensure there is a return statement.

This could be improved by checking all branches.

I know there's a BSLint option that basically checks the same thing, but if you don't have a return statement when you say you're going to return a string or an integer, it will crash.

### DIFF
--- a/src/bscPlugin/validation/ScopeValidator.spec.ts
+++ b/src/bscPlugin/validation/ScopeValidator.spec.ts
@@ -2264,6 +2264,51 @@ describe('ScopeValidator', () => {
                 }).message
             ]);
         });
+
+        it('allows function with no return types with void return value', () => {
+            program.setFile('source/util.bs', `
+                function doSomething()
+                    return
+                end function
+            `);
+            program.validate();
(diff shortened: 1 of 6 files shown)


src/bscPlugin/validation/ScopeValidator.spec.ts
```diff
@@ -2239,6 +2239,95 @@ describe('ScopeValidator', () => {
         });
     });
 
+    describe('expectedReturnStatement', () => {
```
TwitchBronBron, line 2242: We need tests for `sub as <returnType>` as well. This is valid code and works:
```brighterscript
sub getName() as string
  return "bob"
end sub
```

Missing return in `sub as <returnType>`  crashes like function.

image

image

TwitchBronBron (changes requested):
Hey, I added some comments about a few more edge cases we need to test for.

TwitchBronBron (approved):
I agree, this belongs in bsc because it has runtime implications when these are mismatched.

PR #1325, opened 2024-10-11
Fix issues with the .gitignore loading for pathFilterer

TwitchBronBron:
Fixes bugs with the .gitignore entries in PathFilterer for the language server. There were several bugs in there:
 - the `PathFilterer` wasn't registering the exclusion callbacks
 - the gitignore function (called `ignore`) requires relative paths with no `../` or absolute paths, so we need to call `path.relative()` and exclude any paths not matched

### DIFF
--- a/src/LanguageServer.spec.ts
+++ b/src/LanguageServer.spec.ts
@@ -22,6 +22,7 @@ import type { Project } from './lsp/Project';
 import { LogLevel, Logger, createLogger } from './logging';
 import { DiagnosticMessages } from './DiagnosticMessages';
 import { standardizePath } from 'roku-deploy';
+import undent from 'undent';
 
 const sinon = createSandbox();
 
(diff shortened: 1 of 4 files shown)


src/LanguageServer.spec.ts
```diff
@@ -578,6 +579,133 @@ describe('LanguageServer', () => {
         });
     });
 
+    describe.only('rebuildPathFilterer', () => {
```
chrisdp, line 582:
```suggestion
    describe('rebuildPathFilterer', () => {
```

src/lsp/PathFilterer.spec.ts
```diff
@@ -155,4 +161,20 @@ describe('PathFilterer', () => {
         ).to.eql([]);
     });
 
+    describe.only('registerExcludeMatcher', () => {
```
chrisdp, line 164:
```suggestion
    describe('registerExcludeMatcher', () => {
```

PR #1332, opened 2024-10-19
Quiet down grouping expressions in cached lookups

markwpearce:
Code like this would cause a lot of logging:

```
  (1 + 2).toStr()
```

Because the grouping expression was not handled in the call handler in CachedLookups

Also added a test for a specific use case (editing file list through plugin) that was in the docs.

### DIFF
--- a/src/ProgramBuilder.spec.ts
+++ b/src/ProgramBuilder.spec.ts
@@ -14,7 +14,7 @@ import type { BsConfig } from './BsConfig';
 import type { BscFile } from './files/BscFile';
 import { tempDir, rootDir, stagingDir } from './testHelpers.spec';
 import { Deferred } from './deferred';
-import type { BsDiagnostic } from './interfaces';
+import type { AfterProgramCreateEvent, BsDiagnostic } from './interfaces';
 
 describe('ProgramBuilder', () => {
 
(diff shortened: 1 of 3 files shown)


src/ProgramBuilder.spec.ts
```diff
@@ -76,6 +76,26 @@ describe('ProgramBuilder', () => {
         ).to.exist;
     });
 
+
+    it('can edit files array in afterProgramCreate event', async () => {
```
markwpearce, line 80: This specific case is mentioned in the docs, so I thought I should verify

src/astUtils/CachedLookups.ts
```diff
@@ -217,14 +231,13 @@ export class CachedLookups {
                         break;
 
                         //when we hit a variable expression, we're definitely at the leftmost expression so stop
-                    } else if (isVariableExpression(node) || isLiteralExpression(node)) {
+                    } else if (isVariableExpression(node) || isLiteralExpression(node) || isGroupingExpression(node)) {
```
markwpearce, line 234: handle group expression here

src/astUtils/CachedLookups.ts
```diff
@@ -217,14 +231,13 @@ export class CachedLookups {
...
                         break;
-
                     } else if (isDottedGetExpression(node) || isIndexedGetExpression(node)) {
                         node = node.obj;
                     } else {
                         //some expression we don't understand. log it and quit the loop
-                        this.file.program.logger.info('Encountered unknown expression while calculating function expression chain', node);
+                        this.file.program.logger.debug('Encountered unknown expression while calculating function expression chain', node.kind, node.location);
```
markwpearce, line 240: change to a debug message, and just log the kind and location

TwitchBronBron (approved):

Issue #1335, opened 2024-10-23
.BRS files allows declaring function  params `as <customType>`


markwpearce:
eg. This is not showing as an error in a `.brs` file:

```brightscript
function foo(x as roAssoctiativeArray)
    return x["test"
end function
````

PR #1336, opened 2024-10-24
Adds Parser error when custom type encountered in `brs` file

markwpearce:
This PR changes the parser so any custom type (eg. Class names, interfaces, built in types, etc) will result in a diagnostic:

`BrighterScript feature 'custom types' is not supported in standard BrightScript files`

Fixes: #1335

### DIFF
--- a/src/XmlScope.spec.ts
+++ b/src/XmlScope.spec.ts
@@ -306,7 +306,7 @@ describe('XmlScope', () => {
                     print input
                 end sub
             `);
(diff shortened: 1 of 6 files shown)


TwitchBronBron (approved):
Nice work. Looks good to me

TwitchBronBron:
@markwpearce does this also need to be fixed in v0 since we back-ported some custom type logic over there too?

PR #1339, opened 2024-10-29
Fix `Vector2DFieldInterpolator.keyValue` type

markwpearce:
Fix Vector2DFieldInterpolator.keyValue type. Also rescraped docs, and fixed places where 'int' was used instead of 'integer' and `bool` was used instead of `boolean`


scripts/scrape-roku-docs.ts
```diff
@@ -815,7 +816,13 @@ class Runner {
...
 
     private fixFunctionParams(text: string): string {
-        return text.replace(/to as /ig, 'toValue as ');
+        return text.replace(/to as /ig, 'toValue as ')
+            .replace(/as int$/ig, 'as integer')
+            .replace(/as int,/ig, 'as integer,')
+            .replace(/as int\)/ig, 'as integer)')
+            .replace(/as bool$/ig, 'as boolean')
+            .replace(/as bool,/ig, 'as boolean,')
+            .replace(/as bool\)/ig, 'as boolean)');
```
TwitchBronBron, line 825: It might be worth handling whitespace between these keywords as well. Perhaps something like this?

```suggestion
        return text.replace(/to as /ig, 'toValue as ')
            .replace(/as\s+int\s*$/ig, 'as integer')
            .replace(/as\s+int\s*,/ig, 'as integer,')
            .replace(/as\s+int\s*\)\s*/ig, 'as integer)')
            .replace(/as\s+bool\s*$/ig, 'as boolean')
            .replace(/as\s+bool\s*,/ig, 'as boolean,')
            .replace(/as\s+bool\s*\)/ig, 'as boolean)');
```

TwitchBronBron (changes requested):

TwitchBronBron (approved):

PR #1340, opened 2024-10-30
Allows more missing tokens in Transpile

markwpearce:
For some expressions and statements, we can do a better job of assuming what some missing tokens will be.

This is useful for constructing AST nodes via code, so we don't need to specify tokens for open and close parens in a function call, for example.

Includes fixes for `CallExpression`, `PrintStatement` and `FunctionExpression`

Basically, anywhere there is a token where it HAS to be a certain character or string, we might as well put in a default for transpile time, so you don't need to include EVERY token in a constructor.


TwitchBronBron (approved):
Nice!

PR #1341, opened 2024-10-31
Optimize ternary transpilation for assignments

TwitchBronBron:
Whenever a ternary expression is on the right-hand-side of an assignment, dottedSet, or indexedSet statement, we can transpile the entire thing to an `if` statement instead which improves performance significantly (should be identical to handwritten non-ternary code). 

This PR adds this `if` statement transpiling for all of the following situations:
- assignments: `a = true ? 1 : 2`
- dotted set: `m.a = true ? 1 : 2`
- indexed set: `m["a"] = true ? 1 : 2`
- complex assignment operators (i.e. `+=`, `-=`, etc...) `size += true ? 1 : 2`
- grouped ternaries: `a = (((true ? 1 : 2)))`
- nested ternaries `a = true ? (false ? 1 : 2) : 3`

These are explicit cases, so any situation other than these will be transpiled according to the existing rules ([basic](https://github.com/rokucommunity/brighterscript/blob/master/docs/ternary-operator.md#basic-usage) and [scope safe](https://github.com/rokucommunity/brighterscript/blob/master/docs/ternary-operator.md#scope-protection))

**Simple example:**
```brighterscript
a = true ? 1 : 2
```

becomes this:
```brighterscript
if true then
    a = 1
else
    a = 2
end if
```

**Nested example:**
```brighterscript
result = true ? (false ? "one" : "two") : "three"
```

becomes this:
```brighterscript
if true then
    if false then
        result = "one"
    else
        result = "two"
    end if
else
    result = "three"
end if
```

### DIFF
--- a/docs/plugins.md
+++ b/docs/plugins.md
@@ -227,6 +227,38 @@ export interface CompilerPlugin {
     afterProvideReferences?(event: AfterProvideReferencesEvent): any;
 
 
+    /**
+     * Called before the `provideDocumentSymbols` hook
+     */
+    beforeProvideDocumentSymbols?(event: BeforeProvideDocumentSymbolsEvent): any;
+    /**
+     * Provide all of the `DocumentSymbol`s for the given file
+     * @param event
+     */
+    provideDocumentSymbols?(event: ProvideDocumentSymbolsEvent): any;
+    /**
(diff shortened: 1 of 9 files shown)


jtuckerfubo (approved):
lgtm

PR #1347, opened 2024-11-08
Fix issues with the ast walkArray function

TwitchBronBron:
Fix a bug that doesn't properly recover when the array being walked changes.

The core change in logic is as follows:
 - during an AST walk, if the visitor returns a new node, we will not walk the original node's children, and instead walk the new node's children
 - if a visitor changes an array (like `statements`), we will do a simple change detection method: `array[i] !== valuePassedToVisitor`. If we detect that it has changed, we'll restart the walk over that array (but skip any nodes we've already visited). This _can_ result in elements being visited out of order, but the use case for this is to help during transpile, so the order typically matters less.

### DIFF
--- a/src/astUtils/visitors.spec.ts
+++ b/src/astUtils/visitors.spec.ts
@@ -9,8 +9,8 @@ import type { FunctionStatement } from '../parser/Statement';
 import { PrintStatement, Block, ReturnStatement, ExpressionStatement } from '../parser/Statement';
 import { TokenKind } from '../lexer/TokenKind';
 import { createVisitor, walkArray, WalkMode, walkStatements } from './visitors';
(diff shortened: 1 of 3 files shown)


TwitchBronBron:
> I wonder if those arrays should be replaced with an object with getAt, insertAt, removeAt

Yeah, in v1 we could consider revamping the AST editing functionality. However, for v0 that would be a huge breaking change, so it's probably not safe to do that at the moment, unfortunately. 

I think I have something that works for now, it'll ensure that all changed nodes are properly walked (even if a little out of order, which is okay for walking during transpile). It fixed the bugs I was seeing in some of the plugins. 

We can definitely revisit some of this for v1, I have a few ideas based on your suggestion of looking at what Babel does (with the `path` object, enter/exit events, etc.

PR #1352, opened 2024-11-21
Adds Callfunc Completion and Validation

### LINKED ISSUE #1309: Changing XmlFile field does not revalidate dependent files
Adding or removing a field on a component does not trigger revalidation of a file using that component type, but it should. I currently need to edit the brs file to trigger it to revalidate.

Here's a quick demo of the issue. 

https://github.com/user-attachments/assets/7a8d1767-6b64-4d75-ab3b-66ea93ad8564

markwpearce:
Gives `callFunc` a whole lotta love!

- Fixes an issue with completions on `@.` where ALL possible callfuncs would appear. This was a result of the parser having an exception on incomplete `@.` invocations
- Adds function documentation to hovers on `@.` callfunc invocations
- Stores all resolved types associated with callfuncs for a component in a symbolTable that is used a sibiling table when the callfunc is used in another scope
- Return type of `Callfunc` and `@.` invocations are used to set types in assignments 
- Validates arg counts and type mismatches for `callfunc` and `@.`

While I was it, I also added validation and return typing for `<component>.createChild()`

Additional validation enhancements:
- if a type depends on another type (eg, has a member of another type), and that member type changes, then all segments/ASTNodes that reference the wider type are also re-validated. This means that the type-safety flows down and things get appropriately re-validated when there are changes.
- Better handling of `typecast` statements when setting up file segments for revalidation
- Better handling of when component types change, and re-validation because of that.

TODO:
- [x] Make validation updates work better. Currently, you need to edit an XML file of the component for things to revalidate if the exported function type changes
- [x] Deal with how adding sibling tables messes up other validations. Like if the resolved version stay around when they actually should be unresolved
- [x] Do not have validation errors if the component type is the generic `roSGNode` 

To be honest, I think that if we properly re-build the Component type when a member file changes, that would fix both problems.

image

image

Also fixes #1309

### DIFF
--- a/src/AstValidationSegmenter.ts
+++ b/src/AstValidationSegmenter.ts
@@ -1,5 +1,5 @@
 import type { DottedGetExpression, TypeExpression, VariableExpression } from './parser/Expression';
-import { isAliasStatement, isBinaryExpression, isBlock, isBody, isClassStatement, isConditionalCompileStatement, isDottedGetExpression, isInterfaceStatement, isNamespaceStatement, isTypeExpression, isVariableExpression } from './astUtils/reflection';
+import { isAliasStatement, isBinaryExpression, isBlock, isBody, isClassStatement, isConditionalCompileStatement, isDottedGetExpression, isInterfaceStatement, isNamespaceStatement, isTypecastStatement, isTypeExpression, isVariableExpression } from './astUtils/reflection';
 import { ChildrenSkipper, WalkMode, createVisitor } from './astUtils/visitors';
 import type { ExtraSymbolData, GetTypeOptions, TypeChainEntry } from './interfaces';
 import type { AstNode, Expression } from './parser/AstNode';
@@ -73,12 +73,14 @@ export class AstValidationSegmenter {
             return this.checkExpressionForUnresolved(segment, expression.expression.left as VariableExpression, assignedSymbolsNames) ||
                 this.checkExpressionForUnresolved(segment, expression.expression.right as VariableExpression, assignedSymbolsNames);
         }
+        return this.addUnresolvedSymbol(segment, expression, assignedSymbolsNames);
+    }
 
+    private addUnresolvedSymbol(segment: AstNode, expression: Expression, assignedSymbolsNames?: Set<string>) {
         const flag = util.isInTypeExpression(expression) ? SymbolTypeFlag.typetime : SymbolTypeFlag.runtime;
         let typeChain: TypeChainEntry[] = [];
         const extraData = {} as ExtraSymbolData;
         const options: GetTypeOptions = { flags: flag, onlyCacheResolvedTypes: true, typeChain: typeChain, data: extraData };
-
         const nodeType = expression.getType(options);
         if (!nodeType?.isResolvable()) {
             let symbolsSet: Set<UnresolvedSymbol>;
@@ -126,6 +128,8 @@ export class AstValidationSegmenter {
 
     private currentNamespaceStatement: NamespaceStatement;
     private currentClassStatement: ClassStatement;
+    private unresolvedTypeCastTypeExpressions: TypeExpression[] = [];
+
 
     checkSegmentWalk(segment: AstNode) {
         if (isNamespaceStatement(segment) || isBody(segment)) {
@@ -161,6 +165,7 @@ export class AstValidationSegmenter {
             return;
         }
 
+
         this.segmentsForValidation.push(segment);
         this.validatedSegments.set(segment, false);
         let foundUnresolvedInSegment = false;
@@ -168,6 +173,16 @@ export class AstValidationSegmenter {
         const assignedSymbols = new Set<AssignedSymbol>();
         const assignedSymbolsNames = new Set<string>();
         this.currentClassStatement = segment.findAncestor(isClassStatement);
+
+        if (isTypecastStatement(segment)) {
+            if (this.checkExpressionForUnresolved(segment, segment.typecastExpression.typeExpression)) {
+                this.unresolvedTypeCastTypeExpressions.push(segment.typecastExpression.typeExpression);
+            }
+        }
+        let unresolvedTypeCastTypeExpression: TypeExpression;
+        if (this.unresolvedTypeCastTypeExpressions.length > 0) {
+            unresolvedTypeCastTypeExpression = this.unresolvedTypeCastTypeExpressions[this.unresolvedTypeCastTypeExpressions.length - 1];
+        }
 
         segment.walk(createVisitor({
             AssignmentStatement: (stmt) => {
@@ -186,7 +201,11 @@ export class AstValidationSegmenter {
                 assignedSymbolsNames.add(stmt.tokens.item.text.toLowerCase());
             },
             VariableExpression: (expr) => {
-                if (!assignedSymbolsNames.has(expr.tokens.name.text.toLowerCase())) {
+                const hasUnresolvedTypecastedM = unresolvedTypeCastTypeExpression && expr.tokens.name.text.toLowerCase() === 'm';
+                if (hasUnresolvedTypecastedM) {
+                    this.addUnresolvedSymbol(segment, unresolvedTypeCastTypeExpression);
+
+                } else if (!assignedSymbolsNames.has(expr.tokens.name.text.toLowerCase())) {
                     const expressionIsUnresolved = this.checkExpressionForUnresolved(segment, expr, assignedSymbolsNames);
                     foundUnresolvedInSegment = expressionIsUnresolved || foundUnresolvedInSegment;
                 }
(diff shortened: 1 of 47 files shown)


src/AstValidationSegmenter.ts
```diff
@@ -169,6 +174,16 @@ export class AstValidationSegmenter {
         const assignedSymbolsNames = new Set<string>();
         this.currentClassStatement = segment.findAncestor(isClassStatement);
 
+        if (isTypecastStatement(segment)) {
```
markwpearce, line 177: Changes in this file re related to typecast statements... if there's a top level typecast, then if that type changes, everything referencing `m` should be revalidated

src/DiagnosticManager.ts
```diff
@@ -234,7 +234,7 @@ export class DiagnosticManager {
                     isMatch = !!context.tags?.includes(filter.tag);
                 }
                 if (isMatch && needToMatch.scope) {
-                    isMatch = context.scope === filter.scope;
+                    isMatch = context.scope?.name === filter.scope.name;
```
markwpearce, line 237: if an xml file changes, then there is a new scope created, so a check on `context.scope === filter.scope` wouldn't work to remove old diagnostics

src/types/BscType.ts
```diff
@@ -10,6 +10,7 @@ export abstract class BscType {
     public readonly memberTable: SymbolTable;
     protected __identifier: string;
     protected hasAddedBuiltInInterfaces = false;
+    public isBuiltIn = false;
```
markwpearce, line 13: Added a flag at the type level to basically say "this is a type we don't need to mess with"

src/Program.ts
```diff
@@ -385,16 +402,20 @@ export class Program {
      * @param xmlFile XML file with <component> tag
      */
     private addDeferredComponentTypeSymbolCreation(xmlFile: XmlFile) {
-        this.componentSymbolsToUpdate.add({ componentKey: this.getComponentKey(xmlFile), componentName: xmlFile.componentName?.text });
-
+        const componentKey = this.getComponentKey(xmlFile);
```
markwpearce, line 405: changed this to a map with unique keys, because there multiple places where it can get updated

src/Program.ts
```diff
@@ -970,11 +1107,11 @@ export class Program {
         this.logger.info(`Validation Metrics: ${logs.join(', ')}`);
     }
 
-    private getScopesForCrossScopeValidation() {
+    private getScopesForCrossScopeValidation(someComponentTypeChanged = false) {
         const scopesForCrossScopeValidation = [];
         for (let scopeName of this.getSortedScopeNames()) {
             let scope = this.scopes[scopeName];
-            if (this.globalScope !== scope && !scope.isValidated) {
+            if (this.globalScope !== scope && (someComponentTypeChanged || !scope.isValidated)) {
```
markwpearce, line 1114: If a component type changes, don't try to limit the number of scopes to check.
In reality, this hardly changes the actual validation time anyway

src/SymbolTable.ts
```diff
@@ -303,6 +306,23 @@ export class SymbolTable implements SymbolTypeGetter {
         return symbols;
     }
 
+
+    private cachedCircularReferenceCheck: null | boolean = null;
+
+    private hasCircularReferenceWithAncestor() {
```
markwpearce, line 312: ran into some problems with circular references in symbol tables --- best way to deal with that is to check once and cache

src/bscPlugin/validation/ScopeValidator.ts
```diff
@@ -288,18 +293,8 @@ export class ScopeValidator {
         //if this is a `createObject('roSGNode'` call, only support known sg node types
         if (firstParamStringValueLower === 'rosgnode' && isLiteralExpression(call?.args[1])) {
             const componentName: Token = call?.args[1]?.tokens.value;
-            //don't validate any components with a colon in their name (probably component libraries, but regular components can have them too).
```
markwpearce, line 291: refactored some of this code to make checking specific KINDs of function calls easier - eg. callfunc args are offset by 1 vs. real function call args

src/files/BrsFile.ts
```diff
@@ -46,6 +46,7 @@ import { DefinitionProvider } from '../bscPlugin/definition/DefinitionProvider';
 export interface ProvidedSymbol {
     symbol: BscSymbol;
     duplicates: BscSymbol[];
+    requiredSymbolNames?: Set<string>;
```
markwpearce, line 49: These are the required symbol names this symbol provided from the file needs

src/parser/Parser.ts
```diff
@@ -2861,19 +2861,24 @@ export class Parser {
     private callfunc(callee: Expression): Expression {
         this.warnIfNotBrighterScriptMode('callfunc operator');
         let operator = this.previous();
-        let methodName = this.consume(DiagnosticMessages.expectedIdentifier(), TokenKind.Identifier, ...AllowedProperties);
-        // force it into an identifier so the AST makes some sense
-        methodName.kind = TokenKind.Identifier;
-        let openParen = this.consume(DiagnosticMessages.expectedOpenParenToFollowCallfuncIdentifier(), TokenKind.LeftParen);
-        let call = this.finishCall(openParen, callee, false);
-
+        let methodName = this.tryConsume(DiagnosticMessages.expectedIdentifier(), TokenKind.Identifier, ...AllowedProperties);
```
markwpearce, line 2864: this change makes it so an exception is thrown on a bad `@.` line, so that proper completions are provided

src/types/ComponentType.ts
```diff
@@ -39,8 +40,20 @@ export class ComponentType extends InheritableType {
...
-        return isComponentType(targetType) && this.name.toLowerCase() === targetType.name.toLowerCase();
+    isEqual(targetType: BscType, data: TypeCompatibilityData = {}): boolean {
+        if (isReferenceType(targetType) && targetType.isResolvable()) {
+            targetType = targetType.getTarget();
+        }
+        if (this === targetType) {
+            return true;
+        }
+
+        return isComponentType(targetType) && this.name.toLowerCase() === targetType.name.toLowerCase() &&
```
markwpearce, line 51: A better job of checking equality of Component Types, so that we can check if a change was substantial enough to re-validate

src/types/ComponentType.ts
```diff
@@ -60,9 +73,11 @@ export class ComponentType extends InheritableType {
         }
     }
 
+    private hasStartedAddingBuiltInInterfaces = false;
```
markwpearce, line 76: needed to add this incase there were some circular references -- there would be stack overflows

src/types/ComponentType.ts
```diff
@@ -85,17 +103,99 @@ export class ComponentType extends InheritableType {
...
 
     getCallFuncTable() {
         return this.callFuncMemberTable;
     }
 
     getCallFuncType(name: string, options: GetSymbolTypeOptions) {
-        return this.callFuncMemberTable.getSymbolType(name, options);
+        const callFuncType = this.callFuncMemberTable.getSymbolType(name, options);
+
+        const addAssociatedTypesTableAsSiblingToMemberTable = (type: BscType) => {
```
markwpearce, line 173: if you find a callFunc for this a given name, adds a sibling table of the types required for that callFunc to the type.
This makes it so the required types for the callfunc are available in a different context

src/bscPlugin/validation/ScopeValidator.spec.ts
```diff
@@ -3638,4 +3890,395 @@ describe('ScopeValidator', () => {
...
+                sub someFunc(widget as roSGNodeWidget)
+                    otherNode = widget@.getOther("3.14")
+                    print otherNode.myValue
+                end sub
+            `);
+
+            program.validate();
+            expectZeroDiagnostics(program);
+        });
+
```
TwitchBronBron, line 3940: Add a test for doing callfunc on a function call. Looks like right now that doesn't work.

image

TwitchBronBron (changes requested):
Overall, it seems to be working fairly well. I've added a few comments for some bugs I found. Looking forward to testing it again!

src/bscPlugin/validation/ScopeValidator.spec.ts
```diff
@@ -3638,4 +3890,395 @@ describe('ScopeValidator', () => {
         });
     });
 
+    describe('callFunc', () => {
```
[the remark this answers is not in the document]
markwpearce: Fixed a few issues with nodes from component libraries - the types were coming back as unresolved, but this has been changed so they are generic `roSGNode`

src/bscPlugin/validation/ScopeValidator.spec.ts
markwpearce: fixed -- I wasn't properly looking a the *result* of the function call

TwitchBronBron (approved):
This is fantastic! I'm pumped about this feature, and it works great in the projects I've tested it in. Nice work.

PR #1353, opened 2024-11-22
Language Server Crash Fix

addison-adler:
First, apparently this property is deprecated now
image

## Additional Work Remaining
Second, if you do something super dumb like:
```
<component name="MyNode" extends="MyNode">
```
2. ???
3. ~Profit~ crashes!

This causes an infinite loop and a lot of lang. server crashes. I'll look further at this later to see where we can maybe prevent this from crashing and flag the error to the user. Right now there's no feedback besides a ton of crashes


chrisdp:
What does the device do if you try to extend yourself? I assume a compile error?

TwitchBronBron (approved):
Nice! I'll go ahead and merge this for now, you can feel free to open another PR with the additional fixes if you find the time to work on them. :)

addison-adler:
> What does the device do if you try to extend yourself? I assume a compile error?

I'll get that info and open as an issue (or a PR if I can identify a fix)!

PR #1369, opened 2024-12-13
Ensures diagnostic on using Variable from parent function in an inner function

### LINKED ISSUE #1331: Missing diagnostics for use of variables from parent function bodies
BrightScript does not support closures, so we need to properly warn when referencing outer variables from inner functions. 

In `v1.0.0-alpha.39` there is no diagnostic for referencing `outer` in this example, but there should be.

Image

```brighterscript
function test(outer = 1)
    test2 = function(inner = 2)
        print outer + inner
    end function
end function
```

markwpearce:
image

Fixes #1331


TwitchBronBron (approved):
Nice!

PR #1370, opened 2024-12-13
Changes `expectedReturnStatement` to `returnTypeCoercionMismatch` 

### LINKED ISSUE #1334: Handle `function does not return something` edge case
I think https://github.com/rokucommunity/brighterscript/pull/1299 needs to be refactored. This function is `sub as object`, but does not return anything. However, when I print it, it returns `0`. So it should not have an error in brighterscript because it does not cause a runtime error. 

```brightscript
sub RedLines_AddLine(id, position, coords, node, childMap) as object
  line = CreateObject("roSGNode", "Rectangle")
  line.setField("id", id)
end sub
```

From what I can tell, functions `as <something>` will return `invalid`, and the runtime crash happens when it cannot convert `invalid` to the actual return data type:
- `as object` is fine because it converts `invalid` to `0`
- `as integer`, `as longinteger`, `as float`, `as double` are all fine because they convert `invalid` to 0
- `as boolean` is fine because it converts `invalid` to `false`

So really, #1299 should only show an error when the return type of `invalid` cannot be coerced into the return type, and that appears to exclusively be the `string` return type. 

I'm thinking that diagnostic should actually be renamed to "Function will return `invalid` which cannot be coerced into <returnType>`.

markwpearce:
For everything but `string`, a function with no return statement will be coerced into the specified type.

Fixes #1334

### DIFF
--- a/src/DiagnosticMessages.ts
+++ b/src/DiagnosticMessages.ts
@@ -818,8 +818,8 @@ export let DiagnosticMessages = {
         },
         severity: DiagnosticSeverity.Error
     }),
-    expectedReturnStatement: () => ({
-        message: `Expected return statement in function`,
+    returnTypeCoercionMismatch: (returnType = 'string') => ({
+        message: `Function has no return statement and will return 'invalid': '${returnType}' cannot be coerced into 'invalid'`,
         code: 1155,
         severity: DiagnosticSeverity.Error
     })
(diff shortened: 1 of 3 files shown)


TwitchBronBron (approved):

TwitchBronBron:
Looks good. Just fix the merge conflicts and happy to merge this.

TwitchBronBron (approved):

Issue #1372, opened 2024-12-13
Eliminate the `fractionalHexLiteralsAreNotSupported` diagnostic


TwitchBronBron:
Let's eliminate this `fractionalHexLiteralsAreNotSupported` diagnostic entirely. The lexer doesn't do a great job safely detecting this (i.e. it'll show this error for `&HFF.asString()` (i know it doesn't exist, but still, that's the parser job to say a method doesn't exist), so we shouldn't even show this error (and instead show normal parser errors)

_Originally posted by @TwitchBronBron in https://github.com/rokucommunity/brighterscript/pull/1241#discussion_r1882131602_

Issue #1329, opened 2024-10-17
Eliminate bs1107 (unnecessary import)


TwitchBronBron:
This feels more like a lint rule than a compiler error. We should move it to bslint.

![Image](https://github.com/user-attachments/assets/5b6f59b9-247d-45bd-8b80-7b8d01e47ac8)

PR #1377, opened 2024-12-17
Eliminates unnecessary import and fractional hex diagnostics

markwpearce:
Addresses #1372
Addresses #1329

### DIFF
--- a/src/bscPlugin/validation/ScopeValidator.spec.ts
+++ b/src/bscPlugin/validation/ScopeValidator.spec.ts
@@ -2004,6 +2004,43 @@ describe('ScopeValidator', () => {
                 DiagnosticMessages.cannotFindName('outer')
             ]);
         });
+
+        it('allows method call on hex literal', () => {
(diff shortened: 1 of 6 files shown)


TwitchBronBron (commented):
Can you add a unit test to verify that the fractional hex literal is still flagged by the validation flow (just with different diagnostic messages now?)

src/bscPlugin/validation/ScopeValidator.spec.ts
```diff
@@ -2004,6 +2004,43 @@ describe('ScopeValidator', () => {
...
+                    x = &HFF.toStr()
+                    return x
+                end function
+            `);
+
+            program.validate();
+            expectZeroDiagnostics(program);
+        });
+
+        it('has no validation errors with print statement with hex followed by dot <number>', () => {
```
TwitchBronBron, line 2033: @markwpearce I was hoping to see a test or two that showed diagnostics for this type of syntax error.

image

PR #1378, opened 2024-12-17
Properly types 'invalid' as InvalidType

### LINKED ISSUE #1367: Catch addition type mismatch
Image

markwpearce:
Previously, `invalid` was typed as `DynamicType`

This PR changes it so it is `InvalidType`

This fixes the BinaryOperator validation and fixes #1367

In addition:
- when the default value for function params and fields is set to `invalid`, then the type of the param/field is set to Dynamic.
- Changes `AllowableTypes` to NOT include 'invalid' because that is a compile error:

image


TwitchBronBron (approved):
Nice!

Issue #1360, opened 2024-12-03
Validation errors when assigning from built in `as void` function/method


markwpearce:
Built in Brightscript functions that are `as void` do not return invalid like developer created functions.

When you call built-in as void functions that are on the native roku objects, they return uninitialized. When you call developer-written as void functions, they return invalid.
So the runtime behavior is different even though you'd expect as void to behave the same regardless of where they're defined.
And since uninitialized causes runtime crashes when you try to read them and have slightly different rules than invalid, it's important to know when you're using uninitialized vs invalid.

To fix in V1 -- we could just make the doc scraper build a correct `TypedFunctionType` that returns `UninitializedType`  when the return is `as void`
And then validate against uninitialized RHS of assignment statements.

TwitchBronBron:
Interestingly enough, the assignment itself is not the error. It's when you try to _use_ it that it crashes. Like this:

```brighterscript
thing1 = noop()
'this code is fine
if thing1 = invalid then
  print "thing1 is invalid"
end if

thing2 = arr.Append([])
'this code will crash
if thing2 = invalid then
  print "thing2 is invalid"
end if
```


![Image](https://github.com/user-attachments/assets/aca08925-9e47-422a-9b6e-8537c29169c4)

PR #1380, opened 2024-12-20
Augments handling of void returns

markwpearce:
- Adds `isBuiltIn` ExtraSymbolData for symbols that are built in types/functions
- For Built in functions & methods that return  `as void`, return type is `void`
- for User defined functions that return `as void`, the return type is actually `invalid`
- Validates against using `void`-typed vars in Binary expressions and as arguments
- Changes Binary (and Unary) result type code to be more lenient - if one of the types is `dynamic`, then it assumes the actual type is something that provides a better result type... 

eg.

```
function test(foo)
   bar = foo + "Hello"
  ' bar is typed as `string`
end function
```

Addresses #1360


TwitchBronBron:
This might just be a naming thing, but I'm not sure that `void` is the right term here? At runtime, it seems that these built-in objects actually return `uninitialized` instead of `void`, and the runtime errors are when trying to dereference `uninitialized` values. So should we align on the term `UninitializedType` instead of `VoidType`?

markwpearce:
Yeah, uninitialized type is better for this 

- [ ] Change to use `UninitializedType`

Issue #1348, opened 2024-11-18
Require plugins to declare their annotations


TwitchBronBron:
Require plugins to declare their annotations ahead of time so we can provide better editor help and also validation.

This is a breaking change to bsc because annotations would be validated for their existence, so perhaps it's a good candidate for v1.0? 

**Requirements:**
- plugins must declare all of their annotations, which include `name` and any arguments and their types. For future compatibility, we should support an array of signatures (even if we only support one signature at the moment). 

**Benefits:**
 - we can provide intellisense for annotations
 - we can show diagnostics when using unknown annotations

**Possible implementation:**
There's a new plugin lifecycle hook where plugins can declare various settings. Perhaps something like this:

```typescript
onPluginConfigure(event){
    //should this be brighterscript style code? Or json? 
    this.annotations = [
         `inline()`,
        `suite(suiteConfig as { name: string, timeout: number})`
    ]
}

```

markwpearce:
Can this be v1.1 ?

TwitchBronBron:
Maybe? The risk is that if we launch _without_ it, it would be a breaking change to land it later in v1, becuase we would essentially _enforce_ that all annotations are pre-registered.

PR #1386, opened 2025-01-01
Plugins Must Declare Annotations

markwpearce:
Addresses #1348 

Plugins must declare their annotations, so that can be validated.

- [x] Add Documentation about Annotations
- [x] Fix Deprecated docs with references to old plugin events
- [x] Add validation on annotation names and args
- [x] Ensure Annotation args are literals 
- [ ] Agree on format of annotation declarations

Future PR:
- Enhance hovers/completions/Signature help for annotations

image

image

image

image

Extra `info` logging:

```
[11:24:45:357 AM] [prj0] Adding annotation 'takesString' (testPlugin)
```

### DIFF
--- a/docs/plugins.md
+++ b/docs/plugins.md
@@ -50,6 +50,7 @@ While there are no restrictions on plugin names, it helps others to find your pl
 
 Full compiler lifecycle:
 
+- `onPluginConfigure`
 - `beforeProgramCreate`
 - `afterProgramCreate`
     - `afterScopeCreate` ("source" scope)
@@ -75,13 +76,13 @@ Full compiler lifecycle:
     - `afterProgramValidate`
 - `beforePrepublish`
 - `afterPrepublish`
-- `beforePublish`
-    - `beforeProgramTranspile`
+- `beforeSerializeProgram`
+    - `beforeBuildProgram`
     - For each file:
-        - `beforeFileTranspile`
-        - `afterFileTranspile`
-    - `afterProgramTranspile`
-- `afterPublish`
+        - `beforePrepareFile`
+        - `afterPrepareFile`
+    - `afterBuildProgram`
+- `afterSerializeProgram`
 - `beforeProgramDispose`
 
 ### Language server
@@ -90,15 +91,15 @@ Once the program has been validated, the language server runs a special loop - i
 
 When a file is removed:
 
-- `beforeFileDispose`
+- `beforeFileRemove`
 - `beforeScopeDispose` (component scope)
 - `afterScopeDispose` (component scope)
-- `afterFileDispose`
+- `afterFileRemove`
 
 When a file is added:
 
-- `beforeFileParse`
-- `afterFileParse`
+- `beforeProvideFile`
+- `afterProvideFile`
 - `afterScopeCreate` (component scope)
 - `afterFileValidate`
 
@@ -157,10 +158,25 @@ The top level object is the `ProgramBuilder` which runs the overall process: pre
 Here are some important interfaces. You can view them in the code at [this link](https://github.com/rokucommunity/brighterscript/blob/ddcb7b2cd219bd9fecec93d52fbbe7f9b972816b/src/interfaces.ts#L190:~:text=export%20interface%20CompilerPlugin%20%7B).
 
 ```typescript
-export type CompilerPluginFactory = () => CompilierPlugin;
+export type CompilerPluginFactory = () => CompilerPlugin;
 
 export interface CompilerPlugin {
     name: string;
+
+    /**
+     * A list of brighterscript-style function declarations of allowed annotations
+     * Eg.: [
+     *   `inline()`,
+     *   `suite(suiteConfig as object)`
+     * ]
+     */
+    annotations?: string[];
+
+    /**
+     * Called when plugin is initially loaded
+     */
+    onPluginConfigure?(event: onPluginConfigureEvent): any;
+
     /**
      * Called before a new program is created
      */
@@ -240,7 +256,6 @@ export interface CompilerPlugin {
     afterScopeDispose?(event: AfterScopeDisposeEvent): any;
 
     beforeScopeValidate?(event: BeforeScopeValidateEvent): any;
-
     /**
      * Called before the `provideDefinition` hook
      */
@@ -255,7 +270,6 @@ export interface CompilerPlugin {
      * @param event
      */
     afterProvideDefinition?(event: AfterProvideDefinitionEvent): any;
-
 
     /**
      * Called before the `provideReferences` hook
@@ -304,8 +318,6 @@ export interface CompilerPlugin {
      */
     afterProvideWorkspaceSymbols?(event: AfterProvideWorkspaceSymbolsEvent): any;
 
-
-    onGetSemanticTokens?: PluginHandler<OnGetSemanticTokensEvent>;
     //scope events
     onScopeValidate?(event: OnScopeValidateEvent): any;
     afterScopeValidate?(event: BeforeScopeValidateEvent): any;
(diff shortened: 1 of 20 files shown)


markwpearce:
@TwitchBronBron 

The docs I added follow along with the description from the issue, but I'm thinking it may be easier to deal with annotations that are provided as Types?

eg:
```
const myAnnotation = new TypedFunctionType()
myAnnotation.name = 'MyAnnotation';
myAnnotation.params = [
    {name: 'id', type: StringType.instance},
    {name: 'extraData', type: new AssociativeArrayType(), optional: true}
];

this.annotations = [myAnnotation]
```
(perhaps we could update the constructor or create a helper function to create it all at once?)

This would mean we wouldn't have to parse the declaration, but instead use it directly... and wouldn't need to have extra error handling on bad parsed values, etc.

docs/plugins.md
```diff
@@ -75,13 +76,13 @@ Full compiler lifecycle:
...
+    - `beforeBuildProgram`
     - For each file:
-        - `beforeFileTranspile`
-        - `afterFileTranspile`
-    - `afterProgramTranspile`
-- `afterPublish`
+        - `beforePrepareFile`
+        - `afterPrepareFile`
+    - `afterBuildProgram`
+- `afterSerializeProgram`
```
luis-soares-sky, line 85: The remaining API methods are all named with a specific order of "time" > "subject" > "action", i.e. `beforeProgramCreate`, `afterFileValidate`, etc. But these either disregard the "subject" entirely, or have their "subject" and "action" reversed.

Also, not sure I understand the difference between the program's `serialize` and `build` steps, and the need for the `beforePrepublish` and `afterPrepublish` steps. Is there a need for all of these?

```markdown
- `afterProgramCreate`
    - (...)
    - `afterProgramValidate`
- `beforeProgramPublish`
    - For each file:
        - `beforeFilePrepare`
        - `afterFilePrepare`
- `afterProgramPublish`
- `beforeProgramDispose`
```

docs/plugins.md
```diff
@@ -90,15 +91,15 @@ Once the program has been validated, the language server runs a special loop - i
...
 - `afterScopeDispose` (component scope)
-- `afterFileDispose`
+- `afterFileRemove`
 
 When a file is added:
 
-- `beforeFileParse`
-- `afterFileParse`
+- `beforeProvideFile`
+- `afterProvideFile`
```
luis-soares-sky, line 102: Same as above:

```markdown
- `beforeFileProvide`
- `afterFileProvide`
```

luis-j-soares (commented):
Mostly off-topic feedback 👼

docs/plugins.md
markwpearce: I agree the naming should be more consistent. 

@TwitchBronBron  can we make this change? 
also, maybe you could add a little more to these docs to explain what these events represent?

TwitchBronBron: Yeah, I'm definitely open to tweaking some of these names so they're more consistent. 

With regard to the `prepublish`, `serialize`, `build` steps, they each have their own purpose, but we should definitely add more documentation around them to better explain their differences.

docs/plugins.md
TwitchBronBron: Here, file isn't the subject—we're asking plugins to provide a file, not the files to provide something. So the term _subject_ is a bit different, though I can see it being read slightly that way.

But perhaps it doesn't matter, and we should just align on the naming convention anyway?

docs/plugins.md
```diff
@@ -632,6 +644,64 @@ export default function plugin() {
...
+Plugins may provide [annotations](annotations.md) that can be used to add metadata to any statement in the code.
+
+Plugins must declare the annotations they support, so they can be validated properly. To declare an annotation, it must be listed in the `annotations` property - a list of Brighterscript-style function declarations.
+
+For example:
+
+```typescript
+   this.annotations = [
+        'inline()',
+        'log(prefix as string, addLineNumbers = false as boolean)'
```
luis-j-soares, line 658: Already itching to use this for our internal plugins once v1 arrives. I do have a couple of questions, maybe both could be reflected in the docs?

- Can a plugin author expect the full type system to be supported, or will it only support the built-in types? (I see `onlyAllowLiterals: true`, so maybe that's a "no"? I'm not familiar with the type system yet)
- If I have annotation declaration that is implemented by more than one plugin, should a developer assume that the prevalent annotation is based on the order in which the plugins are declared in bsconfig?

src/types/TypedFunctionType.ts
```diff
@@ -46,6 +47,11 @@ export class TypedFunctionType extends BaseFunctionType {
         return this;
     }
 
+    public seVariadic(variadic: boolean) {
```
luis-j-soares, line 50: Small typo:

```suggestion
    public setVariadic(variadic: boolean) {
```

src/interfaces.ts
```diff
@@ -506,6 +526,9 @@ export interface OnGetCodeActionsEvent<TFile extends BscFile = BscFile> {
     codeActions: CodeAction[];
 }
 
+export interface onPluginConfigureEvent {
```
luis-j-soares, line 529: Small casing inconsistency:

```suggestion
export interface OnPluginConfigureEvent {
```

docs/plugins.md
markwpearce: Annotation arguments can only be literals - there’s no way of knowing at compile time what the value of a variable is. 

That being said, we could support `const` and `enum` values as annotation args too. Plan on that in v1.1

so that means, literal strings, integers, floats, literal arrays (with literal values) and literal AA’s can be used as args. 

This would be perfectly fine:

`@annotation(“test”, {values: [1.1, 4, 7.2], data:{name: “Luis”}}, [{abc: 123}])`

markwpearce: I’ll update the docs to reflect that

TwitchBronBron:
> ...I'm thinking it may be easier to deal with annotations that are provided as Types?

I think the challenge comes when we get into more advanced types for these annotations. Consider an annotation that takes one string, or exactly 2 arguments, or exactly 4 arguments (i know we don't have this in bsc yet, but might someday). In typescript we would represent that as:
```typescript
function someAnnotation(someArg: string | [number, number] | [number, number, number, number]) { }
```

We'd be pushing plugin authors to all start learning a more about the type system than perhaps necessary depending on their annotation argument needs.

It feels like declaring annotations as brighterscript code would be more approachable for plugin authors, and might also have a few other benefits: 
 - avoids the plugins using a slightly older brighterscript internally and producing types that don't _quite_ work right with whatever latest brighterscript is currently running
 - no new syntax or AST or type system knowledge necessary
 - much easier to define complex argument types (whenever they're eventually implemented in bsc)

Maybe I'm making this into a bigger deal than it needs to be? Thoughts?

TwitchBronBron:
Perhaps a counter-argument to my argument above would be to provide plugin authors with a helper function to build types from code. Something like this:

```
const complexType = util.createType(`string | [integer, integer] | [integer, integer, integer, integer]`);
```

But it's not too hard to just do this with existing paradigms either...

```brighterscript
const type = Parser.parse(`
    function test(arg as string | [integer, integer] | [integer, integer, integer, integer]): end function
`).ast.findChild(isFunctionStatement).getType();
```

docs/plugins.md
```diff
@@ -632,6 +640,78 @@ export default function plugin() {
...
+## Providing Annotations via a plugin
+
+Plugins may provide [annotations](annotations.md) that can be used to add metadata to any statement in the code.
+
+Plugins must declare the annotations they support, so they can be validated properly. To declare an annotation, it must be listed in the `annotations` property - a list of instances of `TypedFunctionType`, or, if you want to include a description as well, use `AnnotationDeclaration`.
+
+For example:
+
+```typescript
+   this.annotations = [
```
TwitchBronBron, line 652: It would be _really_ nice to enforce which statement types this annotation could be declared on. For example, rooibos `@suite()` _only_ works when attached to classes. Adding `@suite()` to a namespace should be an error. BrighterScript could provide consistent validations for these. I envision configuring it something like this:

```
this.annotations.push({
    restrictTo: ['class'], //this would be a list of all types the annotation maybe declared above
    func:  new TypedFunctionType(VoidType.instance).setName('inline'),
        { 
            description: 'Add a log message whenever this function is called',
            type: new TypedFunctionType(VoidType.instance)
                .setName('log')
                .addParameter('prefix', StringType.instance)
                .addParameter('addLineNumbers', BooleanType.instance, true)
        }

PR #1389, opened 2025-01-03
Assign invalid types as dynamic

### LINKED ISSUE #1383: Too strict cannot find name for type `invalid or dynamic`
This should not be an error, because the variable is `dynamic`. 

Image

Discovered in the latest commits in v1 on  Dec 23, 2024

markwpearce:
Changes handling of `someVar = invalid` so that `someVar` is typed as `dynamic`

fixes #1383


TwitchBronBron (approved):
Looks great!

PR #1390, opened 2025-01-06
Fix class transpile issue with child class constructor not inherriting parent params

TwitchBronBron:
When a child class extends a super class, but does not include its own `new` function, and the parent class has parameters, we are missing those parameters in the child class's factory. 

- [x] unit test
- [x] fix the bug

### DIFF
--- a/src/parser/Statement.ts
+++ b/src/parser/Statement.ts
@@ -2180,9 +2180,13 @@ export class ClassStatement extends Statement implements TypedefProvider {
      */
     private getConstructorParams(ancestors: ClassStatement[]) {
         for (let ancestor of ancestors) {
-            const ctor = ancestor.getConstructorFunction();
-            if (ctor) return ctor.func.parameters;
+            // @todo: somehow, ancestors can have a list where the last element is null.
+            //        this is exposed by the test named "extending namespaced class transpiles properly".
+            if (ancestor) {
+                const ctor = ancestor.getConstructorFunction();
+                if (ctor) return ctor.func.parameters;
             }
+        }
         return [];
     }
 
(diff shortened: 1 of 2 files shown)


src/parser/Statement.ts
```diff
@@ -2173,6 +2173,23 @@ export class ClassStatement extends Statement implements TypedefProvider {
...
 
+    /**
+     * Return the parameters for the first constructor function for this class
+     * @param ancestors The list of ancestors for this class
+     * @returns The parameters for the first constructor function for this class
+     */
+    private getConstructorParams(ancestors: ClassStatement[]) {
+        for (let ancestor of ancestors) {
+            // @todo: somehow, ancestors can have a list where the last element is null.
+            //        this is exposed by the test named "extending namespaced class transpiles properly".
```
luis-soares-sky, line 2184: There's likely to be a bug in `getAncestors`, specifically with this part:

```
const namespace = this.findAncestor<NamespaceStatement>(isNamespaceStatement);
stmt = state.file.getClassFileLink(
    stmt.parentClassName.getName(ParseMode.BrighterScript),
    namespace?.getName(ParseMode.BrighterScript)
)?.item;
```

In some cases (namely the test mentioned in the comment), `stmt` is invalid and still being pushed to the result array. I'll try and figure out where exactly it's becoming null.

luis-soares-sky: I believe the problem is: 

```typescript
this.findAncestor<NamespaceStatement>(isNamespaceStatement)
```

Which should actually be:

```typescript
stmt.findAncestor<NamespaceStatement>(isNamespaceStatement)
```

@TwitchBronBron is it okay if I make that change in this PR? Or would you rather have a separate one for this fix?

TwitchBronBron: Yeah that's fine, go ahead and fix it in this PR.

src/parser/Statement.ts
```diff
@@ -1,7 +1,7 @@
 /* eslint-disable no-bitwise */
 import type { Token, Identifier } from '../lexer/Token';
 import { CompoundAssignmentOperators, TokenKind } from '../lexer/TokenKind';
-import type { BinaryExpression, NamespacedVariableNameExpression, FunctionExpression, FunctionParameterExpression, LiteralExpression } from './Expression';
+import { type BinaryExpression, type NamespacedVariableNameExpression, FunctionExpression, type FunctionParameterExpression, type LiteralExpression } from './Expression';
```
TwitchBronBron, line 4: I've avoided these `import { type ....}` style imports because some other versions of typescript don't like them. 

```suggestion
import type { BinaryExpression, NamespacedVariableNameExpression, FunctionParameterExpression, LiteralExpression } from './Expression';
import { FunctionExpression } from './Expression';
```

Issue #1396, opened 2025-01-13
`util.binaryOperatorResultType` does not handle UnionTypes correctly.


markwpearce:
If one of the operands of `binaryOperatorResultType` is a UnionType, it potentially returns `undefined`

- see problem code below -

This caused a problem in `BinaryOperatorReferenceType` where `binaryOpResolver` would return `undefined`

TwitchBronBron:
I couldn't reproduce the issue with your code above, but this one caused the crash:

```brighterscript
namespace date
    function timeElapsedInDay()
        time = 1
        if true then
            time = getInteger()
        end if
        clockSeconds = getInteger() + 1
        assumedMidnight = time - clockSeconds
        offset = assumedMidnight - 1
    end function

    function getInteger()
        return 1
    end function
end namespace
```

![Image](https://github.com/user-attachments/assets/30a28255-8ca0-483a-b783-20894211867a)

PR #1400, opened 2025-01-14
Fixes How BinaryOperators work on Unions

markwpearce:
New logic:
- the result of a binary op on unions depends on what types are in the union
- Basically, the "highest priority" type wins
- If two types are "equal priority", but not the same, treat it as Dynamic
-  `(float or integer) + (Integer) => Float`
- Treats "ObjectType" basically like "DynamicType"


markwpearce:
fixes #1396

PR #1404, opened 2025-01-28
Fix issues with boxed types

### LINKED ISSUE #1403: Operators not allowed on boxed types
If I use boxed types like `roInt` or `roLongInteger`, the compiler throws errors.

```brighterscript
function math(a as rolonginteger, b as longinteger)
    c = a + b
    c = a - b
    c = a * b
    c = a / b
    c = a \ b

    d = a mod b
    d = a ^ b

    a++
    a--
    a += 1
    a -= 1
    a *= 1
    a /= 1
    a \= 1
    a <<= 1
    a >>= 1

    ? 1 << (2 as roInt)
    ? (1 as roInt) << 2
    ? a and b
    ? a or b

    j = a = b
    j = a <> b
    j = a < b
    j = a <= b
    j = a > b
    j = a >= b
end function
```

Image

markwpearce:
Fixes #1403

Basically, any place where a simple `integer` or `float` could be, the equivalent boxed type is allowed. BrightScript does automatic type conversion between boxed and un-boxed types, so we should treat them the same.

### DIFF
--- a/src/astUtils/reflection.ts
+++ b/src/astUtils/reflection.ts
@@ -309,40 +309,91 @@ export function isPrintSeparatorExpression(element: any): element is PrintSepara
 }
 
 // BscType reflection
-export function isStringType(value: any, includeBoxed = true): value is StringType {
-    const isBoxed = includeBoxed ? isBuiltInType(value, 'roString') : false;
-    return value?.kind === BscTypeKind.StringType || isBoxed;
+export function isStringType(value: any): value is StringType {
+    return value?.kind === BscTypeKind.StringType;
 }
+export function isRoStringType(value: any): value is InterfaceType {
+    return isBuiltInType(value, 'roString');
+}
+export function isStringTypeLike(value: any): value is StringType | InterfaceType {
+    return isStringType(value) || isRoStringType(value);
+}
+
 export function isTypedFunctionType(value: any): value is TypedFunctionType {
     return value?.kind === BscTypeKind.TypedFunctionType;
 }
+
 export function isFunctionType(value: any): value is FunctionType {
     return value?.kind === BscTypeKind.FunctionType;
 }
-export function isBooleanType(value: any, includeBoxed = true): value is BooleanType {
-    const isBoxed = includeBoxed ? isBuiltInType(value, 'roBoolean') : false;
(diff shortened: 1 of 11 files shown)


src/astUtils/reflection.ts
```diff
@@ -309,32 +309,39 @@ export function isPrintSeparatorExpression(element: any): element is PrintSepara
 }
 
 // BscType reflection
-export function isStringType(value: any): value is StringType {
-    return value?.kind === BscTypeKind.StringType;
+export function isStringType(value: any, includeBoxed = true): value is StringType {
```
TwitchBronBron, line 312: Shouldn't these be `false` by default, except in the locations where you know it's allowed? Also, i don't see anywhere that you're passing `false`, so the entire codebase will now be supporting boxed by default. Is all of that okay? Feels a little aggressive, but I don't have all the context.

markwpearce: yeah.. I was thinking this too (`includeBoxed = false` by default)... But everywhere in the code where I check a primitive type, I was going to have to  add `true` anyway

The places I check for these primitive types are in typeCompatibility checks,  operator checks, etc... and as far as I know, BrightScript basically treats these the same anyway.

Can you think of a time a Boxed type is not allowed, when a non-boxed type would be?

TwitchBronBron: You're right. it's probably fine. I'm imagining some future box/not-boxed validations, but you still have the boolean, and in 99.9% of the time, nobody will care. 

So this is fine as-is

TwitchBronBron: Actually, I still have a few thoughts on this.

I think what I'd really like to see is something like:
```
function isStringType()
function isRoStringType()
function isStringTypeLike()
```

I know that's a lot more work to add these to all the boxables, but in the future I envision adding some extra opt-in diagnostics to flag boxed/non-boxed interactions in some cases, and overloading this `isStringType` feels just a little "incorrect" since we _are_ actually comparing two separate types. 

Thoughts?

markwpearce: awesome.. this is better :) 

I'll make the change

TwitchBronBron (approved):

PR #1410, opened 2025-02-06
Fixes `onDidChangeWatchedFiles` flow in LanguageServer

markwpearce:
Some computers (particularly macs, apparently?) trigger ALL THE FILES in an `onDidChangeWatchedFiles` event after starting a debug session, this would cause a HUGE bottleneck and things would crash/halt.

This PR changes the flow for `onDidChangeWatchedFiles` (and the way it subsequently handled file changes) so that:

1. `LanguageServer.handleFileChanges()` -- only call `validateAllThrottled` after handling all file changes, if there actually were any file changes, instead of once per change
2. `LanguageServer.handleFileChange()` -- check the contents of the file vs. the known content for any known file. If the same, ignore it.

This fixes the issue, and the LanguageServer doesn't crash or halt when debugging!

Note: if the LSP changes can't get merged v0, this change could be used a stop-gap measure.

### DIFF
--- a/src/LanguageServer.ts
+++ b/src/LanguageServer.ts
@@ -851,7 +851,7 @@ export class LanguageServer {
     }
 
     @AddStackToErrorMessage
-    private async onDidChangeConfiguration(event: any) {
+    private async onDidChangeConfiguration() {
         if (this.hasConfigurationCapability) {
             //if the user changes any config value, just mass-reload all projects
             await this.reloadProjects(this.getProjects());
--- a/src/LanguageServer.ts
+++ b/src/LanguageServer.ts
@@ -925,9 +925,10 @@ export class LanguageServer {
                 .filter(change => change.type === FileChangeType.Created)
                 //keep only the directories
                 .filter(change => util.isDirectorySync(change.srcPath));
-            const stagingDirs = projects.map(p => p.builder?.options?.stagingDir ?? '').filter(stagingDir => stagingDir);
-            stagingDirs.push('.roku-deploy-staging');
-            stagingDirs.push('node_modules');
+
+            //remove the created directories from the changes array (we will add back each of their files next)
+            changes = changes.filter(x => !directoryChanges.includes(x));
+
             //look up every file in each of the newly added directories
             const newFileChanges = directoryChanges
                 //take just the path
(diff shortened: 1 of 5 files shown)


src/SymbolTable.ts
```diff
@@ -213,7 +214,7 @@ export class SymbolTable implements SymbolTypeGetter {
         if (!resolvedType || originalIsReferenceType) {
             const symbolTypes = this.getSymbolTypes(name, options);
             data = symbolTypes?.[0]?.data;
-            foundFlags = symbolTypes?.[0].flags;
+            foundFlags = symbolTypes?.[0]?.flags;
```
markwpearce, line 217: found this one other exception

src/Program.ts
```diff
@@ -1099,21 +1098,27 @@ export class Program {
                 //sort the scope names so we get consistent results
                 const scopeNames = this.getSortedScopeNames();
                 for (const file of filesToBeValidatedInScopeContext) {
+                    this.logger.debug(`Invalidating scopes for file:`, file.srcPath);
```
markwpearce, line 1101: Program.ts --- just logging changes

src/LanguageServer.ts
```diff
@@ -952,6 +952,16 @@ export class LanguageServer {
             //add the new file changes to the changes array.
             changes.push(...newFileChanges as any);
 
+            //remove the created directories from the changes array (we will add back each of their files next)
+            changes = changes.filter(x => {
+                for (const stagingDir of stagingDirs) {
+                    if (x.srcPath.includes(`/${stagingDir}/`)) {
```
TwitchBronBron, line 958: what's with the forward slashes here?

src/LanguageServer.ts
```diff
@@ -924,10 +925,9 @@ export class LanguageServer {
                 .filter(change => change.type === FileChangeType.Created)
                 //keep only the directories
                 .filter(change => util.isDirectorySync(change.srcPath));
-
-            //remove the created directories from the changes array (we will add back each of their files next)
-            changes = changes.filter(x => !directoryChanges.includes(x));
-
+            const stagingDirs = projects.map(p => p.builder?.options?.stagingDir ?? '').filter(stagingDir => stagingDir);
+            stagingDirs.push('.roku-deploy-staging');
+            stagingDirs.push('node_modules');
```
TwitchBronBron, line 930: I'm not sure this is actually safe. There are valid use cases for doing things like this in a bsconfig.json files array:

```{
    "files": [{
          "src": "../.roku-deploy-staging/someFile.brs",
           "dest": "source/someFile.brs"
       }, {
           "src": "node_modules/some-project/someLib.brs",
           "dest": "source/someLib.brs"
        }]
}
```

There's also the concept of standalone files (i.e. a file that's not represented by a bsconfig.json files array). So if you open the staging folder, it should spin up a new project for that file on its own and provide validation and editor help.

TwitchBronBron (approved):

PR #1411, opened 2025-02-13
Add lookup tables to DiagnosticsManager to speed up diagnostic clearing

markwpearce:
TBH, I'm not sure if this will speed up validation. However, there are many times in the validation process when the Diagnostic Manager needs to clear all diagnostics based on scope, file, tag, etc. Previously, this was done by looping all diagnostics, so if there are lots, it probably went slower and slower. This change should make it much more quick to find the diagnostics that need to be cleared.

Also added some more fine-grained Scope Validation metrics that are printed out with log-level debug:

```
[02:25:25:633 PM] [prj0] Validating scope components/search/SearchRow.xml
[02:25:25:634 PM] [prj0] Validating scope components/search/SearchRow.xml finished. (1.46ms)
[02:25:25:634 PM] [prj0] components/search/SearchRow.xml segment metrics:
[02:25:25:634 PM] [prj0]  -  components/search/SearchRow.brs 11
[02:25:25:634 PM] [prj0] components/search/SearchRow.xml total segments validated 11
[02:25:25:634 PM] [prj0] Validation Metrics (Scope: components/search/SearchRow.xml): fileWalkTime=0.953ms, flagDuplicateFunctionTime=0.21ms, classValidationTime=0.13ms, scriptImportValidationTime=0.40ms, xmlValidationTime=0.17ms
[02:25:25:634 PM] [prj0] BscPlugin onScopeValidate finished. (1.72ms)
```

### DIFF
--- a/src/CrossScopeValidator.ts
+++ b/src/CrossScopeValidator.ts
@@ -255,7 +255,7 @@ export class CrossScopeValidator {
     }
 
     resolutionsMap = new Map<UnresolvedSymbol, Set<{ scope: Scope; sourceFile: BscFile; providedSymbol: BscSymbol }>>();
-    providedTreeMap = new Map<Scope, { duplicatesMap: Map<string, Set<FileSymbolPair>>; providedTree: ProvidedNode }>();
+    providedTreeMap = new Map<string, { duplicatesMap: Map<string, Set<FileSymbolPair>>; providedTree: ProvidedNode }>();
 
 
     private componentsMap = new Map<string, FileSymbolPair>();
@@ -274,10 +274,11 @@ export class CrossScopeValidator {
     }
 
     getProvidedTree(scope: Scope) {
-        if (this.providedTreeMap.has(scope)) {
-            return this.providedTreeMap.get(scope);
+        const lowerScopeName = scope.name?.toLowerCase();
+        if (this.providedTreeMap.has(lowerScopeName)) {
+            return this.providedTreeMap.get(lowerScopeName);
         }
-        const providedTree = new ProvidedNode('');//, this.componentsMap);
+        const providedTree = new ProvidedNode('');
         const duplicatesMap = new Map<string, Set<FileSymbolPair>>();
 
         const referenceTypesMap = new Map<{ symbolName: string; file: BscFile; symbolObj: ProvidedSymbol }, Array<{ name: string; namespacedName?: string }>>();
@@ -376,7 +377,7 @@ export class CrossScopeValidator {
         }
 
         const result = { duplicatesMap: duplicatesMap, providedTree: providedTree };
-        this.providedTreeMap.set(scope, result);
+        this.providedTreeMap.set(lowerScopeName, result);
         return result;
     }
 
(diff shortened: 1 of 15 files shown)


src/util.ts
```diff
@@ -2785,6 +2785,23 @@ export class Util {
         }
         return resultType;
     }
+
+    public findCommonObjects<T>(...arrays: T[][]): T[] {
```
TwitchBronBron, line 2789: Might be something to address later, but array.includes will basically do a sequential loop over every element in the array. Could be pretty expensive. Is there any chance we could change the storage of these items into `Set` instead? Then the .has will be a hash lookup (much faster). But adding to those sets is slightly slower than arrays. 

I'm not sure the context here, just a thought.

src/CrossScopeValidator.ts
```diff
@@ -274,10 +274,11 @@ export class CrossScopeValidator {
     }
 
     getProvidedTree(scope: Scope) {
-        if (this.providedTreeMap.has(scope)) {
-            return this.providedTreeMap.get(scope);
+        const lowerScopeName = scope.name?.toLowerCase();
```
TwitchBronBron, line 277: Just thinking about performance for a moment, could we avoid calling .toLowerCase on the scope name? I think it's already all lower case, but if it's not, then we could create a new field on scope, and just use that. Could cut down on some garbage collection churn?

src/util.ts
markwpearce: Changed to finding Set intersections...

src/files/BrsFile.ts
```diff
@@ -1287,32 +1296,14 @@ export class BrsFile implements BscFile {
                         nameRange: namespaceStatement.nameExpression.location?.range,
                         lastPartName: part.text,
                         lastPartNameLower: lowerPartName,
-                        functionStatements: new Map(),
                         namespaceStatements: [],
-                        namespaces: new Map(),
```
markwpearce, line 1292: This data was not used at all, and so there were a lot of maps created (and cached) that were not used

src/Scope.ts
```diff
@@ -714,33 +732,24 @@ export class Scope {
...
-                        symbolTable: new SymbolTable(`Namespace Aggregate: '${nsContainer.fullName}'`)
+                    const newScopeNsContainer: ScopeNamespaceContainer = {
+                        namespaceContainers: [],
+                        symbolTable: new SymbolTable(`Namespace Scope Aggregate: '${nsContainer.fullName}'`),
+                        firstInstance: nsContainer
                     };
-
-                    clonedNsContainer.symbolTable.mergeSymbolTable(nsContainer.symbolTable);
-                    namespaceLookup.set(lowerNamespaceName, clonedNsContainer);
-                } else {
```
markwpearce, line 734: eliminates a bunch of empty/useless maps

src/Scope.ts
```diff
@@ -914,6 +925,8 @@ export class Scope {
             dispose();
         }
         this.linkSymbolTableDisposables = [];
+
+        this.cache.delete('namespaceLookup');
```
markwpearce, line 929: The `namespaceLookup` was stored in the cache, and not needed outside of when the scope was linked... deleting it here makes sense

markwpearce:
Some of the work here is to make the overall memory footprint smaller:

Here's before & after snapshots of the heap memory when validating & building Jellyfin (with an extra file that adds 4000+ namespaces):
image

TwitchBronBron (approved):

Issue #1413, opened 2025-02-14
"removeParameterTypes" causes compile error for "function () as void"


addison-adler:
slack thread: https://rokudevelopers.slack.com/archives/CKF1QQGTY/p1739548669597599

noticed a bad catch-22 with bsconfig..

preconditions:
1. If you use interfaces (for nicer code hints), `function (args as YourType)` transpiles to `function (arg as object)`
2. If you do not want those runtime checks you have to set `removeParameterTypes: true` 
  - at least personally I avoid using as because it causes runtime crashes that are otherwise easily avoidable with simple no-ops 

However, this also strips down:

```
Function () as void
 if noop return
End Function
```
into:
```
Function ()
  if noop return
End Function
```

resulting in compile errors "return must return a value"

PR #1414, opened 2025-02-14
Fix `removeParameterTypes` compile errors for return types

addison-adler:
relates to bug report https://github.com/rokucommunity/brighterscript/issues/1413

- [ ] Still need to address unit tests! This is only a draft

### DIFF
--- a/src/parser/Expression.ts
+++ b/src/parser/Expression.ts
@@ -177,13 +177,15 @@ export class FunctionExpression extends Expression implements TypedefProvider {
 
         if (this.returnTypeToken) {
             this.returnType = util.tokenToBscType(this.returnTypeToken);
+
         } else if (isSub) {
             this.returnType = new VoidType();
+
         } else {
             this.returnType = DynamicType.instance;
         }
 
-        if (isSub) { // format (4) and (5)
+        if (isSub) { // format (5) and (6)
             this.requiresReturnType = true;
 
         } else if (this.returnTypeToken && isVoidType(this.returnType)) { // format (3)
--- a/src/parser/Expression.ts
+++ b/src/parser/Expression.ts
@@ -393,7 +393,7 @@ export class FunctionExpression extends Expression implements TypedefProvider {
             this.returnType = DynamicType.instance;
         }
 
-        if (isSub) { // format (5) and (6)
+        if (isSub && !isVoidType(this.returnType)) { // format (6)
             this.requiresReturnType = true;
 
         } else if (this.returnTypeToken && isVoidType(this.returnType)) { // format (3)
--- a/src/parser/Expression.ts
+++ b/src/parser/Expression.ts
@@ -158,40 +158,7 @@ export class FunctionExpression extends Expression implements TypedefProvider {
         readonly returnTypeToken?: Token
     ) {
         super();
-
-        /**
-         * RokuOS methods can be written 5 ways:
-         * 1. Function() : return withValue
-         * 2. Function() as type : return withValue
-         * 3. Function() as void : return
-         *
-         * 4. Sub() : return
-         * 5. Sub () as void : return
-         * 6. Sub() as type : return withValue
-         *
-         * Formats (1), (2), and (6) throw a compile error if there IS NOT a return value in the function body.
-         * Formats (3), (4), and (5) throw a compile error if there IS a return value in the function body.
-         */
-
-        const isSub = this.functionType?.text.toLowerCase() === 'sub';
-
-        if (this.returnTypeToken) {
-            this.returnType = util.tokenToBscType(this.returnTypeToken);
-
-        } else if (isSub) {
-            this.returnType = new VoidType();
-
-        } else {
-            this.returnType = DynamicType.instance;
-        }
-
-        if (isSub) { // format (5) and (6)
-            this.requiresReturnType = true;
-
-        } else if (this.returnTypeToken && isVoidType(this.returnType)) { // format (3)
-            this.requiresReturnType = true;
-
-        }
+        this.setReturnType(); // set the initial return type that we parse
 
         //if there's a body, and it doesn't have a SymbolTable, assign one
         if (this.body && !this.body.symbolTable) {
@@ -296,6 +263,7 @@ export class FunctionExpression extends Expression implements TypedefProvider {
             state.transpileToken(this.rightParen)
         );
         //as [Type]
+        this.setReturnType(); // check one more time before transpile
         if (this.asToken && !(state.options.removeParameterTypes && !this.requiresReturnType)) {
             results.push(
                 ' ',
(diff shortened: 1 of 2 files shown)


src/parser/Expression.ts
```diff
@@ -158,11 +158,36 @@ export class FunctionExpression extends Expression implements TypedefProvider {
...
+         * 2. Function () as type
+         * 3. Function () as void
+         * 4. Sub ()
+         * 5. Sub () as type
+         *
+         * Formats (1), (2), and (5) — these throw a compile error if there IS NOT a return value in the function body.
+         * Format (4) is a 'shortcut' to writing format (3) — these throw a compile error if there IS a return value in the function body.
+         */
+
+        this.returnType = null;
```
TwitchBronBron, line 175: Let's move all of this logic into the `transpile()` function (or a private method). Plugins can change all sorts of AST stuff, so I'd like to defer the logic until _right before_ we transpile rather than much earlier, in the constructor.

src/parser/Expression.ts
```diff
@@ -268,7 +294,7 @@ export class FunctionExpression extends Expression implements TypedefProvider {
             state.transpileToken(this.rightParen)
         );
         //as [Type]
-        if (this.asToken && !state.options.removeParameterTypes) {
+        if (this.asToken && !(state.options.removeParameterTypes && !this.requiresReturnType)) {
```
TwitchBronBron, line 297: Could we just do a simple "if this is `function` and it has `as void`, leave it alone`?

addison-adler: There are two scenarios, not one
- if `sub` leave it alone otherwise you compile error
- if `function` and `as void` leave it alone

addison-adler:
I'm pushing up a re-organized code because I also didn't like it 😅

src/parser/Expression.ts
addison-adler: I've moved the code into more logical groupings and extracted it out into its own function 👍

instead of completely removing this from the constructor though, I have it called once at the constructor so that any readers of this object have its initial values, and then it calls one more time before transpile to cover any possible changes

src/parser/Expression.ts
addison-adler: should be easier to see the scenarios now; and the extra function serves as documentation for the scenarios we have to cover

addison-adler:
> I'm pushing up a re-organized code because I also didn't like it 😅

done reorganizing; its stable for review now (granted I still haven't looked at tests yet)

src/parser/Expression.ts
addison-adler: actually there's a 3rd case:
`onKeyEvent` requires `as boolean` otherwise there are OS errors. So I think this is best in its own function 😅

addison-adler:
@TwitchBronBron @chrisdp are there any other changes desired on this PR (besides amending unit test coverage)?

I added a call to `setReturnType()` in the constructor so that plugins/readers have the initial state _and_ during `transpile()` for a final check, but I wasn't sure if the call in the constructor is undesirable overhead?

or if there's anything about this y'all just don't like or want different?

I think keeping this in its own function (rather than a long in-line if condition) is better for documentation since there are so many ways a function/sub can be written and a subset of 3 formats that require keeping the return type in-tact; but other than that I'm not strongly opinionated about the rest

src/parser/Expression.ts
```diff
@@ -346,6 +342,49 @@ export class FunctionExpression extends Expression implements TypedefProvider {
...
+         *
+         * 4. Sub() : return
+         * 5. Sub () as void : return
+         * 6. Sub() as type : return withValue
+         *
+         * Formats (1), (2), and (6) throw a compile error if there IS NOT a return value in the function body.
+         * Formats (3), (4), and (5) throw a compile error if there IS a return value in the function body.
+         *
+         * 7. Additionally, as a special case, the OS requires that `onKeyEvent()` be defined with `as boolean`
+         */
```
chrisdp, line 361: Love the detailed comments. 💪

src/parser/Expression.ts
```diff
@@ -346,6 +342,49 @@ export class FunctionExpression extends Expression implements TypedefProvider {
...
+
+        } else if (isSub) {
+            this.returnType = new VoidType();
+
+        } else {
+            this.returnType = DynamicType.instance;
+        }
+
+        if ((isFunctionStatement(this.parent) || isMethodStatement(this.parent)) && this.parent.name?.text.toLowerCase() === 'onkeyevent') {
+            // onKeyEvent() requires 'asBoolean' otherwise RokuOS throws errors
```
chrisdp, line 376:
```suggestion
            // onKeyEvent() requires 'as Boolean' otherwise RokuOS throws errors
```

src/parser/Expression.ts
```diff
@@ -346,6 +342,49 @@ export class FunctionExpression extends Expression implements TypedefProvider {
...
+
+        const isSub = this.functionType?.text.toLowerCase() === 'sub';
+
+        if (this.returnTypeToken) {
+            this.returnType = util.tokenToBscType(this.returnTypeToken);
+
+        } else if (isSub) {
+            this.returnType = new VoidType();
+
+        } else {
```
chrisdp, line 371:
```suggestion
        if (this.returnTypeToken) {
            this.returnType = util.tokenToBscType(this.returnTypeToken);
        } else if (isSub) {
            this.returnType = new VoidType();
        } else {
```

src/parser/Expression.ts
```diff
@@ -346,6 +342,49 @@ export class FunctionExpression extends Expression implements TypedefProvider {
         return functionType;
     }
 
+    setReturnType() {
```
chrisdp, line 345: Could use a doc block

src/parser/Expression.ts
```diff
@@ -346,6 +342,49 @@ export class FunctionExpression extends Expression implements TypedefProvider {
...
+            // onKeyEvent() requires 'asBoolean' otherwise RokuOS throws errors
+            this.requiresReturnType = true;
+
+        } else if (isSub && !isVoidType(this.returnType)) { // format (6)
+            this.requiresReturnType = true;
+
+        } else if (this.returnTypeToken && isVoidType(this.returnType)) { // format (3)
+            this.requiresReturnType = true;
+
+        }
```
chrisdp, line 385:
```suggestion
            this.requiresReturnType = true;
        } else if (isSub && !isVoidType(this.returnType)) { // format (6)
            this.requiresReturnType = true;
        } else if (this.returnTypeToken && isVoidType(this.returnType)) { // format (3)
            this.requiresReturnType = true;
        }
```

TwitchBronBron:
@addison-adler anything you need from me to help move this one forward? I think we're still waiting for some unit tests yet?

addison-adler:
> @addison-adler anything you need from me to help move this one forward? I think we're still waiting for some unit tests yet?

Oh, I was waiting on you (from the last message [here](https://github.com/rokucommunity/brighterscript/pull/1414#issuecomment-2665464619)) 😆

I wanted to make sure the implementation was up-to-par / correct before writing tests for it. Didn't want to write tests that just had to get thrown out

TwitchBronBron (changes requested):
This implementation looks good to me. Go ahead and add the unit tests, and then we can merge this. :)

src/parser/Expression.ts
TwitchBronBron: the inner description is fine for me. plus I'm converting this to a private method, so the doc block matters even less.

TwitchBronBron (approved):

PR #1424, opened 2025-03-10
Adds diagnosticFiltersV0Compatibility flag to BSConfig.json

markwpearce:
Adds the `diagnosticFiltersV0Compatibility` flag to BSConfig

When that is true, the method `DiagnosticsFilterer.getDiagnosticFilters` will use the v0 format for the `diagnosticFilters` option.

Also re-added all the `DiagnosticsFilterer` tests from v0, with that flag turned on.

### DIFF
--- a/bsconfig.schema.json
+++ b/bsconfig.schema.json
@@ -109,7 +109,7 @@
             "default": false
         },
         "deploy": {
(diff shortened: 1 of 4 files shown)


TwitchBronBron (approved):
This is awesome!

PR #1477, opened 2025-04-23
Fix erraneous diagnostic for component name on first validation

TwitchBronBron:
Fixes a bug with component interface names not being recognized inside of typecasts, when accessing a property on that object, but only on first startup.

### DIFF
--- a/src/CrossScopeValidator.ts
+++ b/src/CrossScopeValidator.ts
@@ -277,7 +277,7 @@ export class CrossScopeValidator {
         if (this.providedTreeMap.has(scope.name)) {
             return this.providedTreeMap.get(scope.name);
         }
-        const providedTree = new ProvidedNode('');
+        const providedTree = new ProvidedNode('', this.componentsMap);
         const duplicatesMap = new Map<string, Set<FileSymbolPair>>();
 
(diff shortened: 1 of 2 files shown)


chrisdp (approved):

src/CrossScopeValidator.ts
```diff
@@ -461,9 +461,10 @@ export class CrossScopeValidator {
     }
 
     clearResolutionsForScopes(scopes: Scope[]) {
+        const lowerScopeNames = scopes.map(scope => scope.name.toLowerCase());
```
TwitchBronBron, line 464: Switching this to a `Set` because that's significantly faster when looking something up.

https://jsperf.app/pikuzu/7

TwitchBronBron: @markwpearce just thinking about general performance. We're going to run `.toLowerCase()` on every scope name every time we do this logic (and probably other logic elsewhere). It would probably be a good idea to store a permanent lower-case name for the scope (or just always force scope.name to _be_ lower case permanently). Not something to solve here, but just a thought. I'll open a ticket to track this so we can talk about it.

PR #1512, opened 2025-06-13
Improve `bsconfig.json` auto-discovery

elsassph:
## Problems

There are issues with `bsconfig.json` auto-discovery:

- Language server may crash with large monorepos making use of symlinks; implementation uses `rokuDeploy.getFilePaths` which is quite inefficient for our need,
- `bsconfig.json` are looked into Search-excluded locations.

## Changes

- replaces `rokuDeploy.getFilePaths` with a direct `fastGlob` call with `followSymbolicLinks` disabled,
- add Search `exclude` to the filters

## Behaviour change

Some users may rely on symlinks or search-hidden locations; this can be worked around by adding manually folders in the Workspace.

### DIFF
--- a/src/LanguageServer.spec.ts
+++ b/src/LanguageServer.spec.ts
@@ -1,7 +1,7 @@
 import { expect } from './chai-config.spec';
 import * as fsExtra from 'fs-extra';
 import * as path from 'path';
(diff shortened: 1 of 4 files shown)


TwitchBronBron (commented):
Looks good! Can you fix the failing test?

https://github.com/rokucommunity/brighterscript/actions/runs/15635246068/job/44048622228?pr=1512

PR #1520, opened 2025-06-18
Add `enableDiscovery` language server option

elsassph:
## Problem with automatic discovery

The language server aggressively tries to discover Roku projects by default.

This is fine for Roku-only developers, and developers working in small to medium repositories.

This is causing issues for developers working on large codebases, like monorepos, and developers working on a lot of non-Roku projects, where the extension goes in the way. For instance in our monorepo, the language server goes out of memory while validating the projects, repeatedly as VS attempts to restart the server.

## Adding a setting allowing to disable automatic discovery

This PR adds an `enableDiscovery` lsp option, which defaults to `true` but can be disabled.

## Uncovered issues

While adding the option, it turned out some refactoring had been done incorrectly; for instance `enableThreading` was never enabled due to incorrect workspace config preparation using unsafe casting.

Fixing the workspace config caused some unit tests to fail (some rather mysteriously).

While fixing the unit tests, I found that we have some questionable logic: why not ensure a `name` token is available in `BrsValidator`?

### DIFF
--- a/src/lsp/ProjectManager.ts
+++ b/src/lsp/ProjectManager.ts
@@ -646,6 +646,11 @@ export class ProjectManager {
      * If none are found, then the workspaceFolder itself is treated as a project
      */
     private async getProjectPaths(workspaceConfig: WorkspaceConfig) {
+        //automatic discovery disabled?
+        if (!workspaceConfig.languageServer.enableDiscovery) {
+            return [workspaceConfig.workspaceFolder];
+        }
+
         //get the list of exclude patterns, negate them so they actually work like excludes), and coerce to forward slashes since that's what fast-glob expects
         const excludePatterns = (workspaceConfig.excludePatterns ?? []).map(x => s`!${x}`.replace(/[\\/]+/g, '/'));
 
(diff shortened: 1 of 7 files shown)


src/lsp/ProjectManager.ts
```diff
@@ -284,7 +284,7 @@ export class ProjectManager {
                         projectPath: s`${projectPath}`,
                         workspaceFolder: s`${workspaceConfig.workspaceFolder}`,
                         excludePatterns: workspaceConfig.excludePatterns,
-                        enableThreading: workspaceConfig.enableThreading
+                        enableThreading: workspaceConfig.languageServer.enableThreading
```
TwitchBronBron, line 287: >enableThreading was never enabled due to incorrect workspace config preparation using unsafe casting.

Wow, nice catch! I just verified this is definitely a problem in the published vscode extension. Our huge app is running both our projects on the same "main" thread instead of in background threads. Yikes!

TwitchBronBron (approved):

TwitchBronBron (approved):

PR #1548, opened 2025-08-05
Support transpiling class methods as named functions

luis-j-soares:
This PR changes the transpiled output of classes from this:

```brightscript
function __MyKlass_builder()
    instance = {}
    instance.new = sub()
    end sub
    instance.abc = function() as integer
    end function
    instance.def = function() as object
    end function
    return instance
end function
function MyKlass()
    instance = __MyKlass_builder()
    instance.new()
    return instance
end function
```

To this:

```brightscript
sub __MyKlass_new()
end sub
function __MyKlass_abc() as integer
end function
function __MyKlass_def() as object
end function
function __MyKlass_builder()
    instance = {}
    instance.new = __MyKlass_new
    instance.abc = __MyKlass_abc
    instance.def = __MyKlass_def
    return instance
end function
function MyKlass()
    instance = __MyKlass_builder()
    instance.new()
    return instance
end function
```

Notes:

- On a call with Bronley and Chris, it was decided that the order of transpilation would be:
  1. Methods (`__[ClassName]_[methodName]`)
  2. Builder (`__[ClassName]_builder`)
  3. Factory (`[ClassName]`)

### DIFF
--- a/src/files/BrsFile.Class.spec.ts
+++ b/src/files/BrsFile.Class.spec.ts
@@ -251,12 +251,13 @@ describe('BrsFile BrighterScript classes', () => {
                     end sub
                 end class
             `, `
+                sub __Animal_new()
+                    m.species1 = "Animal"
+                    print "From Animal: " + m.species
+                end sub
                 function __Animal_builder()
                     instance = {}
-                    instance.new = sub()
-                        m.species1 = "Animal"
-                        print "From Animal: " + m.species
-                    end sub
+                    instance.new = __Animal_new
                     return instance
                 end function
                 function Animal()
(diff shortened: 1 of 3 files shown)


luis-j-soares:
> To keep the diff a little smaller, it might be good to keep any existing functions in their current locations rather than moving them around. Good example is `getTranspiledBuilder`, it seems like it moved? This'll make the merge into v1 a lot easier as well.

Good point. I didn't exactly move `getTranspiledBuilder`, but I added a few new methods above it and the diff viewer doesn't seem to take that well. 😅

I've now moved those new methods below the builder, and the diff seems easier to parse.

TwitchBronBron (approved):

PR #1557, opened 2025-08-21
#1397 Align event names

### LINKED ISSUE #1397: Align v1 Event names
As discussed here: https://github.com/rokucommunity/brighterscript/pull/1386#pullrequestreview-2527246022

We should align plugin event names to match the following format:

```
(before/''/after)<verb>(Program/File/Scope, etc.)
```

This means the following changes:

```
*ProgramCreate -> *CreateProgram
*ScopeCreate -> *CreateScope
*ScopeValidate -> *ValidateScope
*FileAdd -> *AddFile
*FileRemove -> *RemoveFile
etc.
```

Additionally, any plugin event that has a name that starts like `onGet` (eg. `onGetSemanticTokens`) should be changed to `provide...`

 
@luis-j-soares

Christian-Holbrook:
Resolves #1397 

- Standardize the names for the plugin interface. 
  - Use 'provide' instead of 'create'. 
  - Use verbs before nouns. 
  - Use 'remove' instead of 'dispose'
- Add `before`, current (which has no prefix), and `after` prefixes for most events.
- all old names were added to the [plugin sanitizer](https://github.com/rokucommunity/brighterscript/pull/1557/files#diff-cf2d7eaa6cf9bed3dcd3bdd2a5eda4d7579368998985430a81b6cfd44ce1fd85R145-R168) to be upgraded gracefully (with warning) when possible.

### DIFF
--- a/src/BsConfig.ts
+++ b/src/BsConfig.ts
@@ -63,6 +63,16 @@ export interface BsConfig {
     outDir?: string;
 
     /**
+     * @deprecated use `outDir` instead
+     */
+    stagingDir?: string;
+
+    /**
+     * @deprecated use `outDir` instead
+     */
+    stagingFolderPath?: string;
+
+    /**
      * A list of error codes the compiler should NOT emit, even if encountered.
      */
     ignoreErrorCodes?: (number | string)[];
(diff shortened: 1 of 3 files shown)


markwpearce:
I like the consistency of the naming.

TwitchBronBron (approved):

TwitchBronBron (approved):

Issue #1555, opened 2025-08-15
Remove zipping and sideloading from BrighterScript


TwitchBronBron:
Almost nobody uses the `zip` and `sideload` features of BrighterScript. These features should likely reside in other tools (such as roku-deploy), so we should remove them from BrighterScript itself. As a result, the following behaviors will also change:

- By default, the only "output" BrighterScript will create is a folder with transpiled code (including `.d.bs` and `.map` files). This requires:
  - removing all of the code for these features, including any related bsconfig and options properties, as well as related CLI arguments and commands
  - removing the `retainStagingDir` prop and just always retaining it
  - maybe renaming the `stagingDir` name to something like `out` instead? 

Plugins could potentially provide ways to re-implement this behavior for those who need it, but BrighterScript itself should not provide it. Give this some thought

This will be a breaking change for users who rely on these features, but it should simplify the core tool and encourage use of more specialized tooling for deployment and zipping.

PR #1562, opened 2025-08-28
#1555 Remove zip and sideload

Christian-Holbrook:
We are removing the `zip` and `sideload` functionality from `brighterscript`. `roku-deploy` should be used instead.

With this removal, the following parameters are being removed from bsconfig and cli:

1. create-package
2. copy-to-staging -> renamed to no-emit
3. deploy
4. host
5. password
6. retain-staging-folder
7. staging-folder-path -> renamed to out-dir
8. username


src/cli.ts
```diff
@@ -18,27 +18,21 @@ const rl = readline.createInterface({
...
     .option('out-file', { type: 'string', description: 'Path to the zip folder containing the bundled project. Defaults to `./out/[YOUR_ROOT_FOLDER_NAME].zip' })
-    .option('password', { type: 'string', description: 'The password for deploying to a Roku.' })
     .option('project', { type: 'string', description: 'Path to a bsconfig.json project file.' })
-    .option('retain-staging-folder', { type: 'boolean', defaultDescription: 'false', description: 'Prevent the staging folder from being deleted after creating the package.' })
     .option('root-dir', { type: 'string', description: 'Path to the root of your project files (where the manifest lives). Defaults to current directory.' })
-    .option('staging-folder-path', { type: 'string', description: 'The path where the files should be staged (right before being zipped up).' })
-    .option('username', { type: 'string', defaultDescription: '"rokudev"', description: 'The username for deploying to a Roku.' })
+    .option('out-dir', { type: 'string', description: 'TODO Better description The path where the transpiled files should are placed.' })
```
TwitchBronBron, line 35:
```suggestion
    .option('outDir', { type: 'string', description: 'TODO Better description The path where the transpiled files should are placed.' })
```

src/cli.ts
```diff
@@ -18,27 +18,21 @@ const rl = readline.createInterface({
...
     .help('help', 'View help information about this tool.')
-    .option('create-package', { type: 'boolean', defaultDescription: 'true', description: 'Creates a zip package. This setting is ignored when deploy is enabled.' })
     .option('source-map', { type: 'boolean', defaultDescription: 'false', description: 'Enables generating sourcemap files, which allow debugging tools to show the original source code while running the emitted files.' })
     .option('allow-brighterscript-in-brightscript', { alias: 'allow-brighter-script-in-bright-script', type: 'boolean', defaultDescription: 'false', description: 'Allow brighterscript features (classes, interfaces, etc...) to be included in BrightScript (`.brs`) files, and force those files to be transpiled..' })
     .option('cwd', { type: 'string', description: 'Override the current working directory.' })
-    .option('copy-to-staging', { type: 'boolean', defaultDescription: 'true', description: 'Copy project files into the staging folder, ready to be packaged.' })
+    .option('no-emit', { type: 'boolean', defaultDescription: 'true', description: 'TODO better description Do not produce transpiled files' })
```
TwitchBronBron, line 24:
```suggestion
    .option('noEmit', { type: 'boolean', defaultDescription: 'true', description: 'TODO better description Do not produce transpiled files' })
```

src/ProgramBuilder.ts
```diff
@@ -409,17 +409,13 @@ export class ProgramBuilder {
                 return errorCount;
             }
 
-            //create the deployment package (and transpile as well)
-            await this.createPackageIfEnabled();
+            await this.transpile();
```
TwitchBronBron, line 412: Probably need to wrap this in `if(options.noEmit !== true){ ... }`

src/util.ts
```diff
@@ -392,17 +391,13 @@ export class Util {
...
             //use default files array from rokuDeploy
             files: config.files ?? [...DefaultFiles],
-            createPackage: config.createPackage === false ? false : true,
             outFile: config.outFile ?? `./out/${rootFolderName}.zip`,
             sourceMap: config.sourceMap === true,
-            username: config.username ?? 'rokudev',
             watch: config.watch === true ? true : false,
             emitFullPaths: config.emitFullPaths === true ? true : false,
-            retainStagingDir: retainStagingDir,
-            copyToStaging: config.copyToStaging === false ? false : true,
```
TwitchBronBron, line 405: For backwards compatibility (just to make things easier on devs), we should keep `copyToStaging`. Do some type of hierarchy like:
```
 if (options.noEmit is defined){
    use it and ignore copyToStaging
} else if (options.copyToStaging is defined)
    use it
} else {
    set the default
}
```

bsconfig.schema.json
```diff
@@ -100,7 +100,12 @@
             "default": false
         },
         "copyToStaging": {
-            "description": "If true, the files are copied to staging. This setting is ignored when deploy is enabled or if createPackage is enabled",
+            "description": "(Deprecated use 'noEmit') If true, the files are copied to staging. This setting is ignored when deploy is enabled or if createPackage is enabled",
```
TwitchBronBron, line 103:
```suggestion
            "description": If true, the files are copied to staging. This setting is ignored when deploy is enabled or if createPackage is enabled",
                "deprecationMessage": "Deprecated. Use `noEmit` instead.",
```

Should use `deprecationMessage` for the rest of these below as well

src/cli.ts
Christian-Holbrook: Same comment as above. Yargs should provide `noEmit` as an alias.

src/cli.ts
Christian-Holbrook: Alias is the wrong term. It does look like the yargs `argv` object will have a `outDir` property. That's what I was remembering.

If we did change to camelCase here would we change `root-dir` too?

benchmarks/targets/validate.ts
```diff
@@ -8,8 +8,8 @@ module.exports = async (options: TargetOptions) => {
     //run the first run so we we can focus the test on validate
     await builder.run({
         cwd: projectPath,
-        createPackage: false,
-        copyToStaging: false,
+        noEmit: true,
+        enableTypeValidation: true,
```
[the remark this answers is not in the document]
Christian-Holbrook: It might have slipped through autoComplete. Removing it.

src/cli.ts
Christian-Holbrook: After a discussion we landed on camelCase for all the values.

TwitchBronBron (approved):
Looks great!

Issue #1450, opened 2025-04-04
Files set at beforeProgramValidate stage are no longer picked up during validation of other real .bs files


iBicha:
Context: https://rokudevelopers.slack.com/archives/CKF1QQGTY/p1739742183565309
Creating an issue in case we lose access to Slack post after a while

I believe this was introduced here https://github.com/rokucommunity/brighterscript/pull/993

repro project
https://github.com/iBicha/bsc-sample-repro

> I was able to simplfy the sample even further to be very minimalist. It comes down to this: from 0.68.5 to 0.69.0, .bs files set at beforeProgramValidate stage are no longer picked up during validation of other real .bs files. Here's the full example:
> main.bs
> import "pkg:/source/AsyncTask/Tasks.bs"
> 
> sub main()
>     print Tasks.MyTask
> end sub
> plugin
> export class AsyncTaskPlugin implements CompilerPlugin {
>     public name = 'AsyncTaskPlugin';
> 
>     beforeProgramValidate(program: Program) {
>         const fileContent = `enum Tasks
>     MyTask = "_AsyncTask"
> end enum
> `
>         program.setFile('source/AsyncTask/Tasks.bs', fileContent);
>     }
> }
> Works with 0.68.5 but with 0.69.0 Tasks is not defined in the source scope
> Does this ring any bells?

> I can see running bslint that some things are now running out of order (or perhaps is a race condition)
> E.g.
> with 0.68.5 Validating project finished was printed after afterProgramValidate, with 0.69.0 Validating project finished is printed before afterProgramValidate [0.69.0.txt](https://github.com/user-attachments/files/19608155/0.69.0.txt) [0.68.5.txt](https://github.com/user-attachments/files/19608154/0.68.5.txt)

PR #1568, opened 2025-09-09
Fix for adding files on beforeProgramValidate

iBicha:
Closes https://github.com/rokucommunity/brighterscript/issues/1450
Based on suggestion https://github.com/iBicha/brighterscript/pull/1#discussion_r2303476484

### DIFF
--- a/src/Program.spec.ts
+++ b/src/Program.spec.ts
@@ -669,6 +669,27 @@ describe('Program', () => {
             }
             expect(error?.message).to.eql('Crash for test');
         });
+
+        it('includes files added during beforeProgramValidate in validation', () => {
+            program.setFile('source/a.brs', `
+                sub a()
+                    b()
+                end sub
+            `);
+
+            program.plugins.add({
+                name: 'add file in beforeProgramValidate',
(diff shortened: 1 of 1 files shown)


TwitchBronBron (changes requested):
The logic looks good to me. Could you possibly add a unit test or two that ensures the behavior you're fixing is fixed?

iBicha:
> The logic looks good to me. Could you possibly add a unit test or two that ensures the behavior you're fixing is fixed?

Yeah I missed that, added

iBicha:
I don't know why CI is failing, it looks like it could be an unsuitability (I see it related to workers?)
It's passing on my mac

iBicha:
I cleared node_modules and redid
```
npm ci
npm run build
npm run test
```
Things passed
The only difference I could think of is node version - which is [16.20.2](https://github.com/rokucommunity/brighterscript/blob/8d6a4dbf939414f768be21831b74e752c7b76752/.github/workflows/build.yml#L21) that's pretty old, should it be at least at LTS version?

PR #1591, opened 2025-11-17
Adds Inline interfaces

markwpearce:
From the new docs:

Interfaces can be declared inline in any type expression. This provides a quick way to declare the members of an object without having to specify the name of the object or interface.

```brighterscript

function getId(obj as {id as string}) as string
    return obj.id ' obj.id must exist, and it must be a string
end function

```

- [x] Add documentation


docs/interfaces.md
```diff
@@ -146,25 +146,27 @@ end interface
 
 </details>
 
-## Optional members
```
markwpearce: This whole section was duplicated

TwitchBronBron (approved):

PR #1620, opened 2026-02-11
Allow typed functions in type expressions

markwpearce:
- Allows syntax like:

```brighterscript
sub useCallBack(callback as function(name as string, num as integer) as string)
    print "Result is: " + callback("hello", 7)
end sub
```

- Adds validation on typed functions used as arguments
- Provides better hovers for typed functions
- Fixes some issues with using Type statements
- Adds docs

image

image

### DIFF
--- a/docs/typecasts.md
+++ b/docs/typecasts.md
@@ -24,9 +24,9 @@ function getPrefixedId(node, prefix as string) as string
 end function
 ```
 
-## Typecast Statments
+## Typecast Statements
 
-In order to specify all usages of an expression are explicity set to a certian type, you can use a `typecast` statement.
+In order to specify all usages of an expression are explicitly set to a certain type, you can use a `typecast` statement.
 
 Syntax:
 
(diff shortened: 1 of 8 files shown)


docs/readme.md
```diff
@@ -79,6 +79,14 @@ catch ' look, no exception variable!
 end try
 ```
 
+## [Function Types](function-types.md)
+
+```brighterscript
+function useCallback(callback as function(string) as integer) as integer
```
luis-j-soares: Outdated syntax:

```suggestion
function useCallback(callback as function(input as string) as integer) as integer
```

markwpearce:
> Still trying to nail down how the entire type system works, but fwiw it looks good, and works without issue from the little I tested on my project.
> 
> A potential gap is recursive function types, which is not reflected in the tests. Also an outdated example in the docs.

Good point! I'll add a test to show this.

TwitchBronBron (approved):
This looks great! The tests really seem to cover all the bases. Fantastic work! Feel free to merge once you're done making changes.

PR #1624, opened 2026-02-14
Add `relativeSourceMaps` option for portable sourcemaps

### LINKED ISSUE #1156: [sourcemaps] Use relative paths for source file reference in the the generated sourcemaps.
## Summary

Use relative paths for source file reference in the the generated sourcemaps.

## Repro steps

given a project that is located in
`/my-machine/git/my-bs-project`

using the below `./bsconfig.json`:
```json
  "stagingDir": "build"
```
I expect the sourcemaps to be generated with a relative path to the project `cwd` by default, so that when I move the build output from `/my-machine/git/my-projects/build` to `/other-machine/git/my-projects/build` the source references are still valid. 

This has a small added benefit that it also reduces verbosity of logs for users that have long absolute paths, for example `rooibos` uses the source maps to print the test file references in the debugger console.

 **Expected result:**
 
```json 
{
  "version":3,
  "sources": ["../src/my-file.bs"]
   ...
}
```
 
 **Actual result:**
  
```json 
{
  "version":3,
  "sources": ["/my-machine/git/my-bs-project/src/my-file.bs"]
   ...
}
```

bartvandenende-wm:
## Summary

Adds a new `sourceMapRelativePaths` bsconfig option that emits relative file paths in sourcemap `sources` instead of absolute paths. This makes sourcemaps portable across different machines and CI environments.

Resolves #1156

## Problem

Sourcemaps currently embed absolute paths (e.g. `/home/runner/work/project/src/main.bs`). When build output is moved to another machine -- for example when libraries are published to npm and consumed elsewhere -- these paths no longer resolve and debugging tools cannot locate the original source files.

## Solution

A new boolean config option `sourceMapRelativePaths` (default: `false`) post-processes the generated sourcemap at serialization time, converting absolute `sources` entries to paths relative to the `.map` file location. For example:

```json
// Before (absolute)
{ "sources": ["/Users/dev/project/src/source/main.bs"] }

// After (relative, with sourceMapRelativePaths: true)
{ "sources": ["../../src/source/main.bs"] }
```

## VSCode extension compatibility

In [the issue](https://github.com/rokucommunity/brighterscript/issues/1156#issuecomment-2098229498), @TwitchBronBron noted:

> I think we needed absolute paths because of the way the vscode extension moves files around.

After inspecting the [roku-debug](https://github.com/rokucommunity/roku-debug) source, this concern does not apply. The [`SourceMapManager.set()`](https://github.com/rokucommunity/roku-debug/blob/master/src/managers/SourceMapManager.ts) method already resolves source paths using `path.resolve()`:

```typescript
parsedSourceMap.sources = parsedSourceMap.sources.map(source => fileUtils.standardizePath(
    path.resolve(
        //use the map's sourceRoot, or the map's folder path (to support relative paths)
        parsedSourceMap.sourceRoot || path.dirname(sourceMapPath),
        source
    )
));
```


TwitchBronBron:
I pushed a few changes as well to handle `sourceRoot` based on this new relative concept so that it aligns with the spec.

TwitchBronBron (approved):

PR #1652, opened 2026-03-20
Fix for transpile issue with typecast statement with intersection and inline interface

markwpearce:
Fix for: #1651 

- [x] Added test for broken transpile
- [ ] Fix issue


TwitchBronBron (approved):
I think I got it fixed. We'll ship this, and feel free to open a new PR with any further issues you might encounter.

PR #1657, opened 2026-03-24
added selection range provider support to lsp capabilities

### LINKED ISSUE #26: implement grow/shrink selection
should be fairly trivial given the quick and marvellous bright parser

chrisdp:
Implements the LSP [textDocument/selectionRange](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_selectionRange) request, which allows clients to request a set of expanded selection ranges for given cursor positions. Each response is a linked list of ranges from innermost to outermost, giving clients a syntactically-aware path from the cursor outward through the AST.

The server now advertises selectionRangeProvider: true in its capabilities. When a request comes in, it finds the deepest AST node at each position, prepends the exact lexer token range as the tightest step (handling cases like AssignmentStatement.name which is stored as a raw Token rather than a child AST node), then walks the .parent chain collecting each node's range — deduplicating any consecutive identical ranges along the way.

### DIFF
--- a/src/bscPlugin/selectionRanges/SelectionRangesProcessor.spec.ts
+++ b/src/bscPlugin/selectionRanges/SelectionRangesProcessor.spec.ts
@@ -1,0 +1,355 @@
+import { expect } from '../../chai-config.spec';
+import { Program } from '../../Program';
+import { util } from '../../util';
+import { rootDir, trim } from '../../testHelpers.spec';
+import type { SelectionRange } from 'vscode-languageserver-types';
+
+describe('SelectionRangesProcessor', () => {
+    let program: Program;
+
+    beforeEach(() => {
+        program = new Program({ rootDir: rootDir });
+    });
+
+    afterEach(() => {
+        program.dispose();
+    });
+
+    /**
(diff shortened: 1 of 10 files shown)


src/bscPlugin/selectionRanges/SelectionRangesProcessor.spec.ts
```diff
@@ -0,0 +1,355 @@
...
+    function getSelectionRange(code: string, line: number, character: number) {
+        const file = program.setFile('source/main.brs', code);
+        program.validate();
+        const ranges = program.getSelectionRanges(file.pathAbsolute, [util.createPosition(line, character)]);
+        return ranges[0];
+    }
+
+    // -------------------------------------------------------------------------
+    // Basic identifier / variable
+    // -------------------------------------------------------------------------
```
luis-j-soares, line 59: All of these could be "describe" blocks

luis-j-soares (commented):
Can't spot anything obviously wrong, just a minor code org for the tests

TwitchBronBron (approved):

PR #1664, opened 2026-03-26
potental perf fix with file cache useage in scope

chrisdp:
for @TwitchBronBron to review more closely


TwitchBronBron:
No, this isn't the correct fix. The point was to create a cache `'fileMaps-srcPath'` (i.e. all filemaps, cached by srcPath)

PR #1667, opened 2026-03-28
Support line continuation

luis-j-soares:
A future OS release will introduce automatic line continuation in BrightScript, allowing expressions and function call arguments to span multiple lines after binary operators (`+`, `-`, `*`, `/`, `\`, `mod`, `^`, `and`, `or`, and relational operators).

For BrighterScript users, this PR adds parser support for that syntax in BrighterScript mode without introducing a separate feature flag.

Standard `.brs` behaviour remains unchanged unless `allowBrighterScriptInBrightScript` is already enabled, and transpiled output is still normalized back to ordinary single-line BrightScript.

## Behaviour

- Enabled in `.bs` files (BrighterScript mode)
- Enabled in `.brs` files only when `allowBrighterScriptInBrightScript` is set, because those files are then parsed in BrighterScript mode
- Standard `.brs` files still emit syntax errors for binary-operator continuation and multi-line call arguments
- Continued expressions and continued call arguments transpile back to single-line BrightScript output
- Existing multi-line collection literals used as call arguments remain supported in standard `.brs`

## Coverage

- precise negative tests for standard `.brs` syntax errors
- transpile tests proving continued binary expressions and continued call arguments flatten back to one line
- regression tests proving multi-line array and associative-array literals as call arguments are still valid in standard `.brs`

## Examples

**Arithmetic operators**
```brightscript
result = value1 +
         value2 *
         value3

ratio = totalWidth -
        padding /
        columns
```

**Boolean operators**
```brightscript
condition = isValid and
            isEnabled or
            isRequired
```

**Relational operators**
```brighterscript
isInRange = x >=
            minValue and
            x <=
            maxValue
```

**Chained across multiple lines**
```brightscript
total = basePrice +
        tax +
        shippingCost -
        discount
```

**Function call arguments split across lines**
```brightscript
result = calculateValue(
    param1,
    param2
)
```

**Inline collections as arguments (supported since BrightScript's inception — unaffected by this change)**
```brightscript
foo({
    option1: true,
    option2: false
})

foo([
    1,
    2
])
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 
 
+## [0.71.1](https://github.com/rokucommunity/brighterscript/compare/0.71.0...v0.71.1) - 2026-04-14
+### Changed
+ - upgrade to [roku-deploy@3.16.5](https://github.com/rokucommunity/roku-deploy/blob/master/CHANGELOG.md#3165---2026-04-13). Notable changes since 3.16.3:
+     - Resolve files array staging option ([#249](https://github.com/rokucommunity/roku-deploy/pull/249))
+     - Fix `zipFolder` bug that wasn't properly handling absolute stagingDir path ([#245](https://github.com/rokucommunity/roku-deploy/pull/245))
+
+
+
 ## [0.71.0](https://github.com/rokucommunity/brighterscript/compare/0.70.4...v0.71.0) - 2026-03-27
 ### Added
  - Add computed property names (compile-time support only) ([#1658](https://github.com/rokucommunity/brighterscript/pull/1658))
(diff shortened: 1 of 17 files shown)


luis-j-soares:
Note: as the release notes stated that line breaks and whitespace was allowed after binary operators, I originally included relational operators (`>`, `>=`, `<`, `<=`, `<>`, `=`). But because there was no documented example of that, and I can't use OS 15.2 to test the language feature, I decided to keep those operators out of this PR for now. We can add them back easily once we get confirmation that they should also be included.

luis-j-soares:
Converting to draft as I got confirmation that relational operators are supported so I need to reinstate them.

markwpearce:
Will we need to update the formatter as well?

luis-j-soares:
> Will we need to update the formatter as well?

@markwpearce I would say so. I can open an issue and work on it once this PR is merged.

TwitchBronBron (approved):

PR #1669, opened 2026-03-28
Add named argument support and duplicate parameter name warning

chrisdp:
## Summary

- **Named arguments**: allows calling functions using named parameter syntax (e.g. `greet(name: "Bob")`)
  - Validates unknown named arguments against the function definition
  - Errors on duplicate named arguments in the same call
  - Errors on positional arguments following named arguments
  - Errors when named args are used on an unknown function (can't verify param names)
- **Duplicate parameter warning**: emits a warning (code 1150) when a `function`/`sub` definition has two or more parameters with the same name (case-insensitive)
- Fixes `Expression` import in `Scope.ts` to pull from `parser/AstNode` where it is declared

## New diagnostic codes

| Code | Severity | Message |
|------|----------|---------|
| 1146 | Error | Unknown named argument `'{name}'` for function `'{func}'` |
| 1147 | Error | Named argument `'{name}'` was already provided |
| 1148 | Error | Positional arguments cannot follow named arguments |
| 1149 | Error | Cannot use named arguments when calling `'{func}'` because the function definition cannot be found |
| 1150 | Warning | Duplicate parameter name `'{name}'` |

## Test plan

- [ ] Named argument tests in `src/parser/tests/expression/NamedArgument.spec.ts`
- [ ] Duplicate parameter tests in `src/files/BrsFile.spec.ts`

### DIFF
--- a/src/Scope.ts
+++ b/src/Scope.ts
@@ -8,15 +8,17 @@ import type { CallableContainer, BsDiagnostic, FileReference, BscFile, CallableC
 import type { Program } from './Program';
 import { BsClassValidator } from './validators/ClassValidator';
 import type { NamespaceStatement, FunctionStatement, ClassStatement, EnumStatement, InterfaceStatement, EnumMemberStatement, ConstStatement, TypeStatement } from './parser/Statement';
-import type { NewExpression } from './parser/Expression';
+import type { NewExpression, NamedArgumentExpression } from './parser/Expression';
+import type { Expression } from './parser/AstNode';
 import { ParseMode } from './parser/Parser';
+import { createInvalidLiteral } from './astUtils/creators';
 import { util } from './util';
 import { globalCallableMap } from './globalCallables';
 import { Cache } from './Cache';
 import { URI } from 'vscode-uri';
 import type { BrsFile } from './files/BrsFile';
 import type { DependencyGraph, DependencyChangedEvent } from './DependencyGraph';
-import { isBrsFile, isMethodStatement, isClassStatement, isConstStatement, isCustomType, isEnumStatement, isFunctionStatement, isFunctionType, isXmlFile, isNamespaceStatement, isEnumMemberStatement } from './astUtils/reflection';
+import { isBrsFile, isMethodStatement, isClassStatement, isConstStatement, isCustomType, isEnumStatement, isFunctionStatement, isFunctionType, isXmlFile, isNamespaceStatement, isEnumMemberStatement, isNamedArgumentExpression, isVariableExpression } from './astUtils/reflection';
 import { SymbolTable } from './SymbolTable';
 import type { Statement } from './parser/AstNode';
 import { LogLevel } from './logging';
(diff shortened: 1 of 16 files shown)


src/parser/Expression.ts
```diff
@@ -89,6 +89,12 @@ export class CallExpression extends Expression {
 
     public readonly range: Range | undefined;
 
+    /**
+     * When named arguments are used and successfully resolved during validation, this holds the
+     * positional-ordered argument list used for transpilation. If undefined, `args` is used directly.
+     */
+    public resolvedArgs?: Expression[];
```
TwitchBronBron: Remove this, we don't want to be caching this type of information.

src/Scope.ts
```diff
@@ -984,12 +987,159 @@ export class Scope {
         this.diagnostics.push(...validator.diagnostics);
     }
 
+    /**
+     * Validate calls that use named arguments, and resolve them to positional order for transpilation.
+     * Only handles simple function calls (VariableExpression callee); method calls are not supported.
+     */
+    private validateNamedArgCalls(file: BscFile, callableContainerMap: CallableContainerMap) {
```
TwitchBronBron: move this into `ScopeValidator.ts`

TwitchBronBron (changes requested):
We need to change a few things about this:
 - named arguments should be evaluted in the order they are written in the text. So that will require some hoisting to local `__bs*` vars. only need to hoist up until the last call expression. (and can likely skip hoisting VariableExpression and literals (maybe even AA and Array literals as long as they're only composed of literals or VariableExpressions themselves)
 - sequence the local hoisted vars on a per-function basis. (maybe use a WeakMap to use the functionExpression as the key and the sequence as the value)?
 - move the code into ScopeValidator.ts instead of living in Scope.ts
- the transpiler should be the one responsible for reordering the args, and only on the way out....we don't want to manipulate the AST for this if we can help it, because that makes things complicated for other plugins earlier and later in the flow. 
- instead of modifying the transpiler directly, we should do similar to the Ternary transpilation. We modify the CallExpression arguments array during the `beforeFileTranspile` (see BrsFilePreTranspileProcessor)

luis-j-soares:
Shouldn't this be a v1 feature instead?

chrisdp:
@luis-j-soares doesn't need to be. Just means the amount of places that it can be used in isn't quite as many as it would be in v1. But since this is just a call site feature it can be expanded on in v1 to support more locations.

chrisdp:
@copilot i want to update this pr to remove the param reordering for now. 

Enforce the order of params

If someone were to do something like this example we would transpile and populated the missing optional details for them. 

```
sub greet(name = “Unknown” as string, excited = false as boolean)
    if excited
        print "Hello, " + name + "!!!"
    else
        print "Hello, " + name
    end if
end sub

sub main()
    greet(excited: true)
end sub
```
So greet would become:

    greet("Unknown", true)

chrisdp:
@copilot lets also have code actions available to the user to to fix the order.

PR #1682, opened 2026-04-28
Share BscSymbol references in SymbolTable.mergeSymbolTable

chrisdp:
## Summary

Reduce memory pressure during validation by sharing `BscSymbol` references when merging symbol tables, rather than allocating a fresh `{name, range, type}` wrapper per merged symbol.

## Why

`SymbolTable.mergeSymbolTable` was allocating a large amount of `BscSymbol` wrapper objects during validation of large projects, contributing to OOM crashes on brighterscript builds of ~1300+ source-file codebases. Investigation showed:

- `BscSymbol` is a 3-field interface (`name`, `range`, `type`) and is never mutated post-creation in production code (grepped `\.range\s*=` / `\.type\s*=` against symbol-typed values: zero hits).
- `addSymbol` already shared `range` and `type` by reference. Only the wrapper object was being duplicated.
- The only public reader, `hasSymbol` (called from `ScopeValidator` and `BrsFileValidator`), only checks existence and never reads symbol contents.

Removing the redundant wrapper allocation costs nothing semantically and saves significant GC pressure during validation of large codebases.

## Heap profile delta

Measured on a large real-world project (~1300 source files), sampled allocations:

|                            | before    | after     | delta     |
|----------------------------|-----------|-----------|-----------|
| total sampled allocations  | 1933 MB   | 1854 MB   | -80 MB    |
| `mergeSymbolTable` self    | 270 MB    | 226 MB    | -44 MB    |

The win is modest because the wrapper objects are only one part of the merge cost (Map node growth and array slot growth remain). A follow-up PR addresses the larger cost in `buildNamespaceLookup`.

## Changes

- `SymbolTable.mergeSymbolTable` rewritten to push source symbol refs into the destination array directly, instead of calling `addSymbol` per entry.
- Doc-comment on `BscSymbol` declaring it immutable.
- Three new specs in `SymbolTable.spec.ts`:
  - reference identity preservation across merge
  - destination isolation against later source mutations
  - accumulation across multiple sources sharing a key

## Test plan

- [x] `npm run test:nocover` passes (2660 -> 2663 specs; the deltas are the new specs)
- [x] `npx eslint src/SymbolTable.ts src/SymbolTable.spec.ts` clean
- [x] Heap profile re-run on the real-world project shows the documented delta

## Notes

This is the first of several PRs targeting brighterscript memory growth on large codebases.

### DIFF
--- a/src/SymbolTable.ts
+++ b/src/SymbolTable.ts
@@ -115,18 +115,21 @@ export class SymbolTable {
     }
 
     /**
-     * Adds all the symbols from another table to this one
-     * It will overwrite any existing symbols in this table
+     * Adds all the symbols from another table to this one.
+     * Source symbols are shared by reference (not cloned) since BscSymbol is treated as immutable.
+     * The destination still owns its own array per key, so subsequent addSymbol calls on either
+     * table do not leak across.
      */
     mergeSymbolTable(symbolTable: SymbolTable) {
-        for (let [, value] of symbolTable.symbolMap) {
-            for (const symbol of value) {
-                this.addSymbol(
-                    symbol.name,
-                    symbol.range,
-                    symbol.type
-                );
+        for (const [key, sourceSymbols] of symbolTable.symbolMap) {
+            let destSymbols = this.symbolMap.get(key);
+            if (!destSymbols) {
+                destSymbols = [];
+                this.symbolMap.set(key, destSymbols);
             }
+            // for (const symbol of sourceSymbols) {
+            destSymbols.push(...sourceSymbols);
+            // }
         }
     }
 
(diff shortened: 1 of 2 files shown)


src/SymbolTable.ts
```diff
@@ -115,18 +115,21 @@ export class SymbolTable {
...
-                );
+        for (const [key, sourceSymbols] of symbolTable.symbolMap) {
+            let destSymbols = this.symbolMap.get(key);
+            if (!destSymbols) {
+                destSymbols = [];
+                this.symbolMap.set(key, destSymbols);
             }
+            // for (const symbol of sourceSymbols) {
+            destSymbols.push(...sourceSymbols);
+            // }
```
TwitchBronBron:
```suggestion
            destSymbols.push(...sourceSymbols);
```

TwitchBronBron (approved):

PR #1684, opened 2026-04-28
Share per-file namespace data via lazy ScopeNamespaceLookup view

chrisdp:
## Summary

Replace the eager per-scope `buildNamespaceLookup` with a lazy view that shares per-file namespace data across every scope that pulls in the file. Continues the brighterscript memory work from #1682 and #1683.

## Why

After #1683, every Scope's `cache` still holds its own copy of every namespace's data: containers, Maps, SymbolTables. On a project with hundreds of XmlScopes that all import the same `pkg:/source/...` files, the namespace data for those shared files is rebuilt and held N times. That duplication dominates validation-phase allocation throughput on large codebases (the original 2 GB OOM reports we've been chasing).

Most of the data on a `NamespaceContainer` is intrinsic to the **file** that declared the namespace, not to the scope viewing it. Moving that data to the file level lets every scope share the heavy bits.

## Architecture

Three coupled pieces:

### 1. `BrsFile.getNamespaceContributions()` — per-file cache

Lazy `Map<nameLower, NamespaceFileContribution>`. For each namespace name part the file declares (including dotted prefixes — `namespace A.B.C` produces entries for `A`, `A.B`, `A.B.C`), the contribution holds:
- The structural fields (`fullName`, `lastPartName`, `nameRange`, `file`)
- The file's slice of `statements` / `classStatements` / `functionStatements` / `enumStatements` / `constStatements`
- A single-file `SymbolTable` (no parent provider — see audit below)

Cached on the parser-level cache (same pattern as `parser.references`), so it invalidates automatically when the file re-parses.

### 2. `Program.getFilesContributingToNamespace(nameLower)` — reverse index

Lazy `Map<nameLower, Set<BrsFile>>` mapping each name part to the files that contribute to it. Built once per program by walking every `BrsFile.getNamespaceContributions()`. Invalidated by `setFile` / `removeFile` via `Program.invalidateNamespaceIndex()`.

### 3. `ScopeNamespaceLookup` — lazy view

Extends `Map<string, NamespaceContainer>` so plugin types still satisfy. Overrides `get` / `has` / iteration to lazy-build per-key:

- **Fast path** (single in-scope contributor): wraps the file's contribution directly. The container's `statements` / `classStatements` / `functionStatements` / `enumStatements` / `constStatements` / `symbolTable` all point at the contribution's pre-built data. Two scopes that pull in the same file produce wrapper containers with the same inner data refs.

- **Slow path** (multiple in-scope contributors): merges per-scope. Rare in practice but handles cases where a namespace is genuinely split across multiple in-scope files (e.g. a shared declaration file plus theme-specific extensions).

The `namespaces` (children) field is built per-scope for both paths, since which children are in scope depends on the scope's file set.

`Scope.buildNamespaceLookup()` becomes a one-liner returning a fresh `ScopeNamespaceLookup`. Type signature is unchanged.

## Aggregate `SymbolTable` parent provider — dropped

The previous code allocated each aggregate `SymbolTable` with `() => scope.symbolTable` as its parent provider. Under Phase 4, contributions are built per-file (no scope context at construction time) and shared across scopes — so a scope-coupled provider can't be safely set on a shared aggregate.

Audit of every production caller across brighterscript and the four widely-used external plugins:

| Plugin | Aggregate `.symbolTable` reads | Aggregate `.symbolTable.getSymbol/.parent` walks |
|---|---|---|
| brighterscript itself | only `addSibling`/`removeSibling` in `Scope.linkSymbolTable`/`unlinkSymbolTable` | none |
| `rokucommunity/bslint` | none | none |
| `rokucommunity/bsc-plugin-auto-findnode` | none | none |
| `slheavner/brighterscript-xml-plugin` | none | none |
| `rokucommunity/rooibos` | none (`scope.symbolTable.hasSymbol` is on the Scope's table, not a container's) | none |

Sibling resolution in `SymbolTable.getSymbol` checks `sibling.symbolMap` directly without recursing into `sibling.parent`, so the parent provider was dead code in every observed call path. Dropping it is safe.

## Heap profile delta

Measured on the same large real-world project as #1682 / #1683 (~1300 source files), sampled allocations, cumulative on top of the prior two PRs:

|                            | master    | after #1682 | after #1683 | this PR    | total Δ |
|----------------------------|-----------|-------------|-------------|------------|---------|
| total sampled allocations  | 1933 MB   | 1854 MB     | 1468 MB     | **854 MB** | **-56%** |
| `mergeSymbolTable` self    | 270 MB    | 226 MB      | 221 MB      | (off list) | -       |
| NamespaceContainer literal | 511 MB    | 511 MB      | 345 MB      | (off list) | -       |
| `Map` constructor          | 442 MB    | 442 MB      | 250 MB      | 119 MB     | -323 MB |

This PR alone is responsible for ~614 MB of reduction (42% on top of #1683).

## Wall-clock delta

Phase timings on the same real-world build:

| Phase | master | this PR | Δ |
|---|---|---|---|
| Parsing | 600 ms | 600 ms | (noise) |
| **Validating** | **14,054 ms** | **10,902 ms** | **-22%** |
| Copying to staging | 1,134 ms | 753 ms | -34% (mostly I/O noise) |
| Transpiling | 2,304 ms | 1,950 ms | -15% |

Single-run measurements, not averaged — the directional story is robust but exact percentages will vary. For projects that currently OOM at the default 2 GB heap, the wall-clock improvement is likely much larger than these numbers suggest, since GC mark-compact thrashing (which the original crash logs showed dominating CPU at `mu = 0.190`) is what eats the time in the failing case.

## Plugin compatibility

No external plugin needs changes. `Scope.buildNamespaceLookup()` still exists with the same return type. `NamespaceContainer` shape is unchanged from #1683 (optional fields stay optional, in the same way). Plugins that iterate `scope.namespaceLookup.entries()` (e.g. rooibos) keep working — the new lazy view triggers full population on `entries()`, matching today's eager behavior.

## Test plan

- [x] `npm run test:nocover` passes (2666 specs, 0 failing; 3 new specs added in this PR for fast path / slow path / file-replace invalidation)
- [x] `npx eslint src/Scope.ts src/Scope.spec.ts src/Program.ts src/files/BrsFile.ts` clean
- [x] Heap profile re-run on the real-world project shows the documented delta
- [x] Build wall-clock measured

## Notes for reviewers

- This is the third PR in the series targeting brighterscript memory growth on large codebases. #1682 and #1683 are the precursors; combined they take total sampled allocations from 1933 MB to 854 MB.
- The lazy view is a behavioral change for non-build consumers: if a plugin queries `namespaceLookup.size` or iterates `.keys()` repeatedly, it now triggers `populateAll` on first access and pays the same cost as today's eager build. After that it's free.
- The audit-derived decision to drop the aggregate parent provider is the most "load-bearing" change here and the part most worth a careful read. If a reviewer can identify any code path that walks `container.symbolTable.parent` or calls `container.symbolTable.getSymbol(...)` directly, that's the case where the parent provider drop matters and we should switch to a file-level fallback.


src/files/BrsFile.ts
```diff
@@ -160,6 +162,74 @@ export class BrsFile {
...
+     * Each entry covers a single dotted name part (so a `namespace A.B.C` declaration
+     * produces entries for `A`, `A.B`, and `A.B.C`). Intermediates carry only structural
+     * fields; leaves populate the relevant statement collections plus a single-file
+     * symbolTable. The symbolTable has no parent provider (see Phase 4 design notes).
+     *
+     * The returned Map is shared across every Scope that pulls in this file, which is
+     * the core sharing primitive Phase 4 relies on.
+     */
+    public getNamespaceContributions(): Map<string, NamespaceFileContribution> {
+        return this.cache?.getOrAdd('BrsFile_namespaceContributions', () => {
```
TwitchBronBron: This is a brsFilie instance cache already. scrap the `BrsFile_` prefix

src/files/BrsFile.ts
```diff
@@ -160,6 +162,74 @@ export class BrsFile {
...
+            const contributions = new Map<string, NamespaceFileContribution>();
+            for (const namespaceStatement of this.parser.references.namespaceStatements) {
```
TwitchBronBron: let's use this.ast.walk(createVisitor({ NamespaceStatement: ()=>{}}, { walkMode: WalkMode.visitStatements}});

src/Scope.ts
```diff
@@ -1365,3 +1339,262 @@ export interface NamespaceContainer {
...
+        this.ensureNameIndex();
+        return this.nameSet!.has(lower);
+    }
+
+    public override get(key: string): NamespaceContainer | undefined {
+        const lower = key?.toLowerCase();
+        if (!lower) {
+            return undefined;
+        }
+        if (super.has(lower)) {
```
TwitchBronBron: Would be nice ot have comments calling out that we're using a pre-computed value. it's easy to skim it.

TwitchBronBron (approved):

PR #1685, opened 2026-04-29
Memoize getName/getNameParts on namespaced AST nodes

chrisdp:
## Summary

Memoize the result of `getName` / `getNameParts` on `NamespacedVariableNameExpression`, `NamespaceStatement`, `FunctionStatement`, `ClassStatement`, and `InterfaceStatement`. Validators call these methods per potential namespace reference per scope, and each call walks the AST (`findAncestor`) and rebuilds the qualified name string from scratch. The result is a pure function of the parsed AST, so caching it on the node is safe and eliminates a large amount of repeated work.

## Measured impact

Profiled on a brighterscript build of ~1300 source files (using `node --cpu-prof --heap-prof`):

| Metric | Before | After | Δ |
|---|---|---|---|
| Total CPU sampled time | 13.58 s | **12.03 s** | **-1.55 s (-11.4%)** |
| Validate phase wall-clock | 8.74 s | **7.60 s** | **-1.14 s (-13%)** |
| GC self-time | 1.37 s | **1.08 s** | -290 ms (-21%) |
| Total heap allocations | 876 MB | **869 MB** | flat (within variance) |

CPU profile attribution before this change: `NamespacedVariableNameExpression.getName` 455 ms self, `getNameParts` 247 ms self, `Statement.getName` and the `(get name)` getter chain together another ~800 ms self. After this change those frames drop out of the top-25 entirely.

Heap stays essentially flat: each relevant AST node grows by two `string | undefined` slots (~16 bytes), but that's offset by the eliminated intermediate string and array allocations from repeated calls.

## Implementation notes

**`getNameParts` returns a defensive slice.** Two existing callers mutate the returned array — `BrsFile.ts:999` uses `.shift()`, `symbolUtils.ts:141` uses `.pop()`. The slice on each call is trivial relative to the AST walk it replaces, and preserves the original mutable-array contract for any external callers.

**Statement-level `getName` only writes to the cache when `this.parent` is set.** Without that guard, `NamespaceStatement`'s constructor (which calls `this.name` to build a SymbolTable label *before* the AST walk that establishes parent links) would lock in the parentless local name forever. The guard defers caching until after parent linking, so a nested namespace correctly caches its full qualified name on first post-walk read.

**Cache fields use the `_cached*` prefix.** `AstNode.spec`'s clone-equality test was updated to skip those slots: the cache populates lazily and need not match between a node and its clone (each populates independently on first read).

### DIFF
--- a/src/parser/AstNode.spec.ts
+++ b/src/parser/AstNode.spec.ts
@@ -217,6 +217,9 @@ describe('AstNode', () => {
                     //skip these properties
                     if (
                         ['parent', 'symbolTable', 'range'].includes(key) ||
+                        //internal memoization slots are populated lazily and need not match between
+                        //a node and its clone — they'll repopulate independently on first read
+                        key.startsWith('_cached') ||
                         //this is a circular reference property or the `returnType` prop, skip it
                         (isFunctionExpression(original) && (key === 'functionStatement' || key === 'returnType')) ||
                         //circular reference property for annotations
--- a/src/parser/Expression.ts
+++ b/src/parser/Expression.ts
@@ -504,6 +504,15 @@ export class NamespacedVariableNameExpression extends Expression {
     }
     range: Range | undefined;
 
+    //memoized name results. The AST is treated as readonly post-parse, so once these
+    //are computed they remain valid for the lifetime of the expression. Validators
+    //call `getName` per potential namespace reference per scope, often dozens of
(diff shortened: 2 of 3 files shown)


TwitchBronBron:
I don't really want to do this at this time. Too much difficulty knowing when to clear these cached values anytime AST edits are made. We need to think about this more and come up with a better more broad caching system.

PR #1696, opened 2026-05-03
Recover from mismatched loop terminators with quick fixes

chrisdp:
## Summary

When a user writes a common loop-terminator mistake, the parser used to emit two cascading diagnostics that didn't connect the user's intent to the actual problem. This PR adds error recovery in three loop-parsing paths and a single parameterized diagnostic with a structured `data` payload that drives the quick fixes.

### Before

```
while n <= 3
    n = n + 1
next
```

emitted:
- `Unexpected token 'next'`
- `Could not find matching 'end while'`

### After

Same input emits **one** diagnostic:
- `Expected 'end while' but found 'next'` (code 1147)

with a quick fix `Convert 'next' to 'end while'`.

## Cases handled

| Input | Diagnostic message | Quick fix(es) |
|---|---|---|
| `while ... next` | `Expected 'end while' but found 'next'` | `Convert 'next' to 'end while'` |
| `for ... end while` | `Expected 'end for' or 'next' but found 'end while'` | `Convert 'end while' to 'end for'` (preferred), `Convert 'end while' to 'next'` |
| `for each ... end while` | same as `for` | same two fixes |

When the same mistake appears multiple times in a file, fix-all variants are also offered.

## Implementation

- New `mismatchedEndingToken(expected: string[], found: string)` diagnostic (code 1147). The `expected` array is preferred-first, and both fields are kept on `diagnostic.data` so consumers can build fixes mechanically rather than dispatching on per-pair codes.
- `whileStatement` accepts `Next` as a recovery terminator; consumes it on the bogus path and stores it on `WhileStatement.tokens.endWhile` so quick fixes can target the token's range. Calls `mismatchedEndingToken(['end while'], 'next')`.
- `forStatement` and `forEachStatement` accept `EndWhile` as a recovery terminator; same shape. They call `mismatchedEndingToken(['end for', 'next'], 'end while')`. `forEachStatement` also refactored from "throw on missing terminator" to recover-and-continue.
- `CodeActionsProcessor.suggestMismatchedEndingTokenQuickFixes` reads `expected[]`/`found` from the diagnostic and emits one fix per legal terminator (first entry is `isPreferred`). No per-pair handler.
- New `isDiagnosticOfType(diagnostic, key)` type guard refines `BsDiagnostic` to the concrete `DiagnosticMessageType<K>` when its code matches, used by the code-action dispatcher.

## Test plan

- [x] `npm test` — 2838 passing, 0 failing
- [x] `npm run lint` — clean
- [x] Parser specs assert the new diagnostic code and `data: { expected, found }` payload for `while`, `for`, and `for each`
- [x] Code-action specs cover single-fix, two-fix (with preferred), and fix-all variants

### DIFF
--- a/src/DiagnosticMessages.ts
+++ b/src/DiagnosticMessages.ts
@@ -771,16 +771,23 @@ export let DiagnosticMessages = {
         code: 1146,
         severity: DiagnosticSeverity.Error
     }),
-    whileLoopTerminatedWithNext: () => ({
-        message: `Expected 'end while' but found 'next'`,
-        code: 1147,
-        severity: DiagnosticSeverity.Error
-    }),
-    forLoopTerminatedWithEndWhile: () => ({
-        message: `Expected 'end for' or 'next' but found 'end while'`,
-        code: 1148,
-        severity: DiagnosticSeverity.Error
-    })
+    /**
+     * Emitted when a block recovers from a wrong terminator (e.g. `while ... next` or `for ... end while`).
+     * `expected` lists the legal terminators in preferred-first order; `found` is the actual text.
+     * Quick fixes consume the structured `data` to build "Convert '<found>' to '<expected[i]>'" actions.
+     */
+    mismatchedEndingToken: (expected: string[] = [], found = '') => {
+        const expectedList = Array.isArray(expected) ? expected : [];
+        return {
+            message: `Expected ${expectedList.map(text => `'${text}'`).join(' or ')} but found '${found}'`,
+            code: 1147,
+            data: {
+                expected: expectedList,
+                found: found
+            },
+            severity: DiagnosticSeverity.Error
+        };
+    }
 };
 
 export const DiagnosticCodeMap = {} as Record<keyof (typeof DiagnosticMessages), number>;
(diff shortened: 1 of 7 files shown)


src/DiagnosticMessages.ts
```diff
@@ -770,6 +770,16 @@ export let DiagnosticMessages = {
...
+    whileLoopTerminatedWithNext: () => ({
+        message: `Expected 'end while' but found 'next'`,
+        code: 1147,
+        severity: DiagnosticSeverity.Error
+    }),
+    forLoopTerminatedWithEndWhile: () => ({
+        message: `Expected 'end for' or 'next' but found 'end while'`,
+        code: 1148,
+        severity: DiagnosticSeverity.Error
     })
```
TwitchBronBron: Let's combine these into a single "mismatched ending token" thing. pass in the array of "expected" texts that will be included in the message, and keep it on the diagnostic to help with code help.

src/DiagnosticMessages.ts
```diff
@@ -770,6 +770,16 @@ export let DiagnosticMessages = {
...
+    }),
+    whileLoopTerminatedWithNext: () => ({
+        message: `Expected 'end while' but found 'next'`,
+        code: 1147,
+        severity: DiagnosticSeverity.Error
+    }),
+    forLoopTerminatedWithEndWhile: () => ({
+        message: `Expected 'end for' or 'next' but found 'end while'`,
+        code: 1148,
+        severity: DiagnosticSeverity.Error
```
TwitchBronBron: Let's combine these into a single "mismatched ending token" thing. pass in the array of "expected" texts that will be included in the message, and keep it on the diagnostic to help with code help.

TwitchBronBron (changes requested):

TwitchBronBron (approved):

TwitchBronBron:
@copilot fix the merge conflicts with master. If there's a conflict with diagnostics, this one needs to add its diagnostic to the end of the list and ensure we have the latest number in this pr, since we are the incoming change.

TwitchBronBron (approved):

PR #1697, opened 2026-05-03
Diagnose reserved BrightScript builtins used as values

chrisdp:
## Summary

Adds a new diagnostic (`reservedBuiltinUsedAsValue`, code 1147) that fires when a reserved call-only BrightScript builtin appears in any non-call position. These builtins parse cleanly today but produce a device compile error (`Syntax Error. Builtin function call expected`, &h9d, or generic `Syntax Error` &h02 for some).

Builtins covered:
- `box`, `createobject`, `eval`, `getglobalaa`
- `getlastruncompileerror`, `getlastrunruntimeerror`
- `objfun`, `pos`, `run`, `tab`, `type`

## Before / after

| Code | Before | After |
|---|---|---|
| `x = ObjFun` | (no diagnostic — broke at device compile time) | `'ObjFun' is a reserved builtin and can only be used as a function call (e.g. 'ObjFun(...)')` |
| `x = type` | (no diagnostic) | same shape |
| `x = eval` | (no diagnostic) | same shape |
| `print foo(Box, 2)` | (no diagnostic) | flagged |
| `ObjFun()`, `type(123)`, `Box(1)`, `eval(...)` etc. | (no diagnostic) | unchanged — call sites are not flagged |
| `m.type = 1`, `{ type: 1 }` | (no diagnostic) | unchanged — property access and AA keys not flagged |
| `type Name = ...` (BrighterScript type alias) | (no diagnostic) | unchanged — different parse path |

## Implementation

- New `CallOnlyBuiltins` set in [src/lexer/TokenKind.ts](src/lexer/TokenKind.ts), keyed by lowercased text.
- New diagnostic `reservedBuiltinUsedAsValue` (code 1147) in [src/DiagnosticMessages.ts](src/DiagnosticMessages.ts).
- 11-line guard inside `primary()` at [src/parser/Parser.ts](src/parser/Parser.ts) that fires when `text.toLowerCase()` is in `CallOnlyBuiltins` and `peek().kind !== LeftParen`.

## Why these specific builtins

Each was tested on a Roku device by writing the various forms (`x = Foo`, `Foo(...)`, `m.Foo = 1`, `{ Foo: 1 }`, `function f(Foo)`) and observing whether each form was a device compile error, runtime error, or successful execution. The 11 names above are the set where non-call usage is a hard compile error on device. `eval` is included because referencing it as a value has always been a compile error across firmware versions; its separate deprecation/removal lifecycle for *call* sites is handled in #1698.

## Test plan

- [x] `npm test` — 2806 passing, 0 failing
- [x] `npm run lint` — clean
- [x] Per-builtin RHS-read flag tests + canonical-call no-flag tests (including eval)
- [x] Per-builtin parameter-name regression tests (existing 1045 already covers these)
- [x] Property-access / AA-key / BrighterScript-type-alias / case-insensitivity tests
- [x] Existing `type statement` regression test updated to expect the new diagnostic for `return type`


src/lexer/TokenKind.ts
```diff
@@ -577,6 +577,27 @@ export const DisallowedLocalIdentifiersText = new Set([
...
 ]);
 
+/**
+ * Reserved BrightScript builtins that compile only when invoked as a function call.
+ * Any non-call reference (e.g. `x = type`, `print foo(ObjFun)`) is a device compile error
+ * (most produce `Syntax Error. Builtin function call expected`, hex code &h9d;
+ * `createobject` and `tab` produce a generic `Syntax Error` &h02 with the same outcome).
+ * Names are lowercased for case-insensitive matching.
+ */
+export const CallOnlyBuiltins = new Set([
```
TwitchBronBron: Don't love this name. What about `UnreferencableBuiltins  `instead?

src/parser/Parser.ts
```diff
@@ -3018,7 +3019,20 @@ export class Parser {
                 return this.templateString(true);
 
             case this.matchAny(TokenKind.Identifier, ...this.allowedLocalIdentifiers):
-                return new VariableExpression(this.previous() as Identifier);
+                const ident = this.previous() as Identifier;
```
TwitchBronBron: Don't think this is the right place to validate. Should move to file or scope validation so we can capture any plugin-injected code using these names as well.

TwitchBronBron (changes requested):

TwitchBronBron (approved):

PR #1698, opened 2026-05-03
Validate eval/rsg_version against firmware lifecycle

chrisdp:
## Summary

Adds firmware-aware diagnostics for `eval(...)` calls and the manifest's `rsg_version` entry, driven by lifecycle metadata in a new [src/RokuConstants.ts](src/RokuConstants.ts) plus per-callable `availability` metadata in [globalCallables.ts](src/globalCallables.ts).

## What this catches

| Code | Manifest | minFirmwareVersion | Outcome |
|---|---|---|---|
| 1149 `globalCallableRemoved` | (any) | resolves to rsg_version >= 1.2 | flag every `eval(...)` call as removed |
| 1150 `globalCallableDeprecated` | `rsg_version=1.1` | modern firmware | warn to migrate `eval(...)` (os-axis nudge when the rsg axis falls back) |
| 1151 `rsgVersionRequiresMinFirmware` | `rsg_version=1.3` | < 15.0 | flag manifest |
| 1152 `invalidRsgVersionFormat` | `rsg_version=banana` | (any) | warning on the manifest line |
| 1153 `rsgVersionDeprecated` | `rsg_version=1.2` | >= 15.1 | suggest upgrading to 1.3 |
| 1154 `rsgVersionRemoved` | `rsg_version=1.0` | >= 9.0 | error: device no longer honors the entry |
| 1154 `rsgVersionRemoved` | `rsg_version=1.1` | >= 14.5 | error: silently treated as 1.2 |

## Default-firmware behavior change

When the user hasn't configured `minFirmwareVersion`, brighterscript now assumes a default of **15.0.0** so these diagnostics are useful out of the box. Users targeting older firmware can opt out by setting `minFirmwareVersion` explicitly.

This also affects the existing optional-chaining firmware check ([BrsFileValidator.ts](src/bscPlugin/validation/BrsFileValidator.ts)): the same default is applied uniformly via the new `Program.getEffectiveMinFirmwareVersion()` helper.

## Implementation

- **[src/RokuConstants.ts](src/RokuConstants.ts)**: single source of truth for RSG version lifecycle (`introducedAt` / `becameDefaultAt` / `deprecatedAt` / `removedAt` / `replacement`), plus `DEFAULT_MIN_FIRMWARE_VERSION` and `OPTIONAL_CHAINING_MIN_FIRMWARE_VERSION`. Adding a new RSG version is one entry in `RSG_VERSIONS`; downstream validators pick it up.
- **[globalCallables.ts](src/globalCallables.ts)**: global callables now carry an `availability` block (`os` / `rsg` axes, each with `deprecated` / `removed` thresholds). `eval` declares `os.deprecated=9.0.0` and `rsg.removed=1.2.0`; the validator derives diagnostics from this metadata, so the check generalizes to any callable that gains availability data.
- **[Manifest.ts](src/preprocessor/Manifest.ts)**: adds `parseManifestEntries()` (line-aware) alongside the existing `parseManifest()`. Public API unchanged.
- **[Program.ts](src/Program.ts)**: adds `getEffectiveMinFirmwareVersion()` (sanitizes garbage input to the default) and `getRsgVersion()` (data-driven default selection). Both `getManifestEntries()` and `getManifestPath()` are `protected` for now (no stable public API yet, will firm up when the manifest gets proper file-tracking support).
- **[ProgramValidator.ts](src/bscPlugin/validation/ProgramValidator.ts)**: new manifest validator deriving diagnostics from `RSG_VERSIONS`. Removal takes precedence over deprecation; deprecation only fires when the user's effective firmware actually meets the deprecation threshold (so users on pre-deprecation firmware aren't nagged).
- **[BrsFileValidator.ts](src/bscPlugin/validation/BrsFileValidator.ts)**: adds a callable-availability check (covering `eval`) next to the existing optional-chaining check; both route through `getEffectiveMinFirmwareVersion()`.

## Sources for lifecycle data

- [Declare RSG 1.3 support by October 1, 2026](https://blog.roku.com/developer/rsg-1-3) (Roku dev blog)
- [Roku SceneGraph 1.1 apps sunset](https://blog.roku.com/developer/rsg-1-1-sunset)
- [Roku Channel Manifest docs](https://developer.roku.com/dev/docs/channel-manifest)
- Roku OS 9 / 14.5 release notes for the eval and 1.1-substitution timelines

## Known gaps / TODOs

These are documented inline so we can revisit as the manifest infrastructure matures:

- **Manifest isn't a tracked file**: diagnostics use a `{ srcPath, pkgPath }` stub. This should be replaced by proper file tracking. Code actions for manifest diagnostics are deferred until then.
- **`getManifestEntries` / `getManifestPath` accessed via bracket notation**: protected for now to avoid committing to a public API.
- **`becameDefaultAt` semantically broadened**: for 1.0/1.1/1.2 it means "device silent fallback"; for 1.3 it means "Roku's cert-policy expected default" (1.3 isn't a runtime default but is required by Roku's static analysis tool). A TODO comment in RokuConstants flags this for confirmation with Roku.


src/bscPlugin/validation/BrsFileValidator.ts
```diff
@@ -62,6 +63,7 @@ export class BrsFileValidator {
                 if (node.openingParen?.kind === TokenKind.QuestionLeftParen) {
                     this.validateMinFirmwareVersionForOptionalChaining(node.openingParen.range);
                 }
+                this.validateEvalIsNotDeprecated(node);
```
TwitchBronBron: Let's call this type of stuff more generically. All the GlobalCallables could have this type of information, so we can generically determine whether any function is available, deprecated, or removed.

src/Program.ts
```diff
@@ -1782,6 +1790,82 @@ export class Program {
...
+    }
+
+    /**
+     * The minimum Roku firmware version brighterscript should assume the user is targeting.
+     * If `options.minFirmwareVersion` is set AND parseable as semver, that wins; otherwise
+     * (unset or unparseable) falls back to {@link DEFAULT_MIN_FIRMWARE_VERSION}. This
+     * guarantees the return is always a coerceable version string, so downstream callers
+     * never have to handle malformed input.
+     */
+    public getEffectiveMinFirmwareVersion(): string {
```
TwitchBronBron:
```suggestion
    public getMinFirmwareVersion(): string {
```

src/Program.ts
```diff
@@ -1782,6 +1790,82 @@ export class Program {
...
+    }
+
+    /**
+     * Returns the effective `rsg_version` for this program. If the manifest declares a value
+     * explicitly, that's returned verbatim (including malformed values, so callers can validate
+     * format themselves). Otherwise, the highest known rsg_version whose `becameDefaultAt` is
+     * `<=` the effective minimum firmware version is returned — driven entirely by the data in
+     * {@link RSG_VERSIONS}.
+     */
+    public getRsgVersion(): string {
```
TwitchBronBron, line 1884: cache the entire results of this call, store it on this, and return that if set.

src/bscPlugin/validation/ProgramValidator.ts
```diff
@@ -33,4 +36,81 @@ export class ProgramValidator {
...
+
+        const info = RSG_VERSIONS[value];
+        if (!info) {
+            //version is parseable as semver but not in our known map — trust the user;
+            //we don't have firmware-compat data for it.
+            return;
+        }
+
+        const effectiveFw = this.program.getEffectiveMinFirmwareVersion();
+        const coercedFw = semver.coerce(effectiveFw);
```
TwitchBronBron: don't have to coerce here, cuz we're gonna coerce in program.getMinFirmwareVersion

src/Program.ts
```diff
@@ -1782,6 +1790,82 @@ export class Program {
...
+
+    /**
+     * The minimum Roku firmware version brighterscript should assume the user is targeting.
+     * If `options.minFirmwareVersion` is set AND parseable as semver, that wins; otherwise
+     * (unset or unparseable) falls back to {@link DEFAULT_MIN_FIRMWARE_VERSION}. This
+     * guarantees the return is always a coerceable version string, so downstream callers
+     * never have to handle malformed input.
+     */
+    public getEffectiveMinFirmwareVersion(): string {
+        const userValue = this.options.minFirmwareVersion;
```
TwitchBronBron: Let's cache the results of this file, no need to keep recomputing it.

src/bscPlugin/validation/ProgramValidator.ts
```diff
@@ -33,4 +36,81 @@ export class ProgramValidator {
...
+            if (coercedFw && coercedDeprecated && semver.gte(coercedFw, coercedDeprecated)) {
+                this.program.addDiagnostics([{
+                    ...DiagnosticMessages.rsgVersionDeprecated(value, info.replacement),
+                    file: { srcPath: manifestPath, pkgPath: 'manifest' } as any,
+                    range: rsgEntry.range
+                }]);
+            }
+        }
+
+        const coercedRequired = semver.coerce(info.introducedAt);
```
TwitchBronBron: Let's make all of our hardcoded values coerced by default so no need to call it at the usage sites

src/DiagnosticMessages.ts
```diff
@@ -770,6 +770,31 @@ export let DiagnosticMessages = {
         message: `'${featureName}' requires Roku firmware version ${minimumVersion} or higher (current target is ${configuredVersion})`,
         code: 1146,
         severity: DiagnosticSeverity.Error
+    }),
+    evalIsDeprecatedAtRsgVersion: (rsgVersion: string) => ({
```
TwitchBronBron: Need a more generic diagnostic, this is too specific to this function name.

TwitchBronBron (changes requested):

TwitchBronBron (approved):

PR #1703, opened 2026-05-08
Add parameter name inlay hints

chrisdp:
## Summary
- Adds parameter-name inlay hints at call sites: `foo(name: "bar", count: 5)` style ghost annotations next to each argument.
- New `provideInlayHints` plugin event mirrors the existing `provideHover` / `provideSelectionRanges` shape, with `before` / `after` variants for plugins that want to filter or augment results.
- LSP server advertises `inlayHintProvider` and registers `textDocument/inlayHint` via `connection.languages.inlayHint.on(...)`.
- Resolves call targets across plain calls, namespace calls (`ns.foo`), `m.method()` inside a class, `new ClassName(...)` constructors, and `callfunc` dispatches. Suppresses hints when the argument is a bare identifier already matching the parameter name.

## Out of scope (future PRs)
- Variable / inferred-type hints next to declarations.
- Function return-type hints.
- Per-category opt-in settings (`brighterscript.inlayHints.parameterNames`, etc.).


chrisdp:
@copilot link this PR to the following issue. https://github.com/rokucommunity/vscode-brightscript-language/issues/745

TwitchBronBron (approved):

PR #1705, opened 2026-05-11
Token leadingTrivia handling and reduce memory retention in long-lived compiler/LSP caches

markwpearce:
## Summary

This PR tightens memory behavior and trivia handling across lexer/parser/transpile paths.

It does two main things:
1. Reduces retained memory in long-lived objects and caches.
2. Makes trivia access resilient now that `leadingTrivia` is optional when empty.

## What changed

- Added disposal cleanup to release cached references earlier:
  - clear BrsFile lookup caches during dispose
  - clear Scope cache during dispose
  - unlink Scope symbol table during dispose
- Reduced cache growth pressure:
  - bounded/managed standardizePath cache behavior
  - added URI conversion caching for path-to-URI conversion
- Reduced per-token overhead:
  - lexer now omits `leadingTrivia` when there is no trivia
  - `Token.leadingTrivia` is now optional
- Updated downstream consumers for optional trivia:
  - parser/transpile/comment helpers now use null-safe trivia access
  - completion/comment checks updated to handle missing trivia arrays
- Small correctness/cleanup updates:
  - typo fix (`isPostionInComment` -> `isPositionInComment`)
  - lazy allocation of duplicate-symbol map in cross-scope validation

## Why

In editor-driven and long-running language server sessions, small retained references and unbounded cache growth can accumulate over time.  
These changes keep behavior intact while lowering memory pressure and making trivia handling more explicit and robust.

## Testing

- Added Scope disposal cache-clearing test.
- Added BrsFile disposal lookup-cache test.
- Added lexer test to verify trivia arrays are only present when needed.
- Updated existing tests to support optional trivia arrays.

## Results

This code was run on an app with these metrics:

|       | Memory | Load Time | Validation Time | Build Time |
| --- | --- | --- | --- | --- |
| Before | 890MB | 884ms | 6483ms | 2551ms |
| After | 774MB | 786ms | 6420ms | 2606ms |

### DIFF
--- a/src/CrossScopeValidator.ts
+++ b/src/CrossScopeValidator.ts
@@ -278,7 +278,7 @@ export class CrossScopeValidator {
             return this.providedTreeMap.get(scope.name);
         }
         const providedTree = new ProvidedNode('', this.componentsMap);
-        const duplicatesMap = new Map<string, Set<FileSymbolPair>>();
+        let duplicatesMap: Map<string, Set<FileSymbolPair>> = null;
 
         const referenceTypesMap = new Map<{ symbolName: string; file: BscFile; symbolObj: ProvidedSymbol }, Array<{ name: string; namespacedName?: string }>>();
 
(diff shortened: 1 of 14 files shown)


TwitchBronBron (commented):
Hmm. So this path caching was added to improve performance. I benchmarked it to verify that it was significantly faster to cache these lookups. 

Adding a function call in the middle of this hot function might negate a significant amount the benefit of caching in the first place. 

Just curious, did you do any profiling to see how much memory this cache actually gets? Is it actually significant or just a guess?

markwpearce:
> Hmm. So this path caching was added to improve performance. I benchmarked it to verify that it was significantly faster to cache these lookups.
> 
> Adding a function call in the middle of this hot function might negate a significant amount the benefit of caching in the first place.
> 
> Just curious, did you do any profiling to see how much memory this cache actually gets? Is it actually significant or just a guess?

Yeah.. That was just AI.. I'll revert that change.

TwitchBronBron (approved):

PR #1706, opened 2026-05-12
Support plugin config values in bsconfig and CLI

TwitchBronBron:
Adds plugin configuration support to both `bsconfig.json` and the CLI, plus a new `Plugin.onSetConfiguration(config)` lifecycle hook that fires immediately after every plugin is loaded.

## bsconfig.json

`plugins` entries can now be objects with `src`, `name?`, and `config?` (the previous string shorthand still works):

```jsonc
{
    "plugins": [
        "./scripts/myPlugin.js",
        { "src": "@rokucommunity/bslint", "name": "bslint", "config": "./bslint.jsonc" },
        { "src": "./scripts/custom.js", "config": { "severity": "error" } }
    ]
}
```

`config` is either an inline object or a path to a JSONC file (comments and trailing commas allowed).

## CLI

**Load and name** with `--plugin <src> [<name>]` (repeatable):

```bash
bsc --plugin "@rokucommunity/bslint" bslint \
    --plugin ./scripts/custom.js house-rules
```

**Override config** with `--plugin.<id>.<prop>=<value>` (deep-merge):

```bash
bsc --plugin.bslint.severity=error --plugin.bslint.rules.no-print=off
```

**Replace config entirely** with a bare `--plugin.<id>=<value>` (no dotted path); if the value is a path, the JSONC file is loaded. Combinable with merge — replace is always applied first regardless of CLI order:

```bash
# Loads ./ci-bslint.jsonc as the whole config, then overrides `severity`
bsc --plugin.bslint=./ci-bslint.jsonc --plugin.bslint.severity=hint
```

**Quoted segments** anywhere in the dotted key handle ids/props containing dots:

```bash
bsc --plugin."my.module".enabled=false
bsc --plugin.bslint."rules.no-underscores"=warn
```

## ID resolution

`<id>` is matched against loaded plugins with strict precedence — a higher-priority match wins outright, and lower-level collisions are ignored:

1. bsconfig user-supplied `name`
2. Plugin factory `name`
3. bsconfig `src` (as written)

Ambiguity at the chosen level hard-fails with a message listing the conflicting plugins.

Full docs (with more examples) in `docs/plugins.md`.

### DIFF
--- a/docs/plugins.md
+++ b/docs/plugins.md
@@ -49,9 +49,48 @@ The supported properties are:
 - **`name`** *(optional)* — overrides the plugin's own `name` property. Useful for distinguishing multiple instances of the same plugin loaded with different configs.
 - **`config`** *(optional)* — plugin-specific configuration, either:
   - **An inline object** — passed directly to the plugin.
-  - **A path to a JSONC file** (relative to `cwd`) — the file is parsed and its contents are passed to the plugin. JSONC files support JavaScript-style comments (`//` and `/* */`).
+  - **A path to a JSONC file** — the file is parsed and its contents are passed to the plugin. JSONC files support JavaScript-style comments (`//` and `/* */`). Paths in `bsconfig.json` are resolved relative to the `bsconfig.json` file's directory.
 
 Both the string shorthand and the object form can be mixed in the same `plugins` array.
+
+#### Example: JSONC config file
+
+A JSONC config file lets you keep large or commented plugin configuration out of `bsconfig.json`. Given this `bsconfig.json`:
+
+```json
+{
+    "plugins": [
+        {
+            "src": "@rokucommunity/bslint",
+            "config": "./bslint.jsonc"
+        }
+    ]
+}
+```
+
+…the referenced `bslint.jsonc` might look like this:
+
+```jsonc
+{
+    // overall severity for all rules
+    "severity": "error",
+
+    // skip these files when linting
+    "ignorePatterns": [
+        "**/*.spec.bs",
+        "source/generated/**/*.brs"
+    ],
+
+    "rules": {
+        "no-print": "error",
+        "no-underscores": "warn",
+        /* allow `m.foo` style access in test files only */
+        "consistent-this": "off"
+    }
+}
+```
+
+The compiler parses the file (comments and trailing commas allowed) and hands the resulting object to the plugin via `onSetConfiguration`.
 
 ### Receiving configuration in your plugin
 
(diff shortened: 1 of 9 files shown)


markwpearce:
Should this be in v1? 

If not, can we change some stuff to match what v1 is doing with plugins? For Example changing event names.. "onSetConfiguration" -> "setConfiguration" (verb_noun format)

src/cli.ts
```diff
@@ -25,12 +28,7 @@ let options = yargs
     .option('cwd', { type: 'string', description: 'Override the current working directory.' })
     .option('copy-to-staging', { type: 'boolean', defaultDescription: 'true', description: 'Copy project files into the staging folder, ready to be packaged.' })
     .option('diagnostic-level', { type: 'string', defaultDescription: '"warn"', description: 'Specify what diagnostic types should be printed to the console. Value can be "error", "warn", "hint", "info".' })
```
TwitchBronBron, line 27: We need to discuss this, because it's technically a breaking change to remove this alias.

docs/plugins.md
```diff
@@ -21,11 +21,349 @@ Those plugins will be loaded by the VSCode extension and can provide live diagno
...
 
+### Plugin configuration objects
+
+Instead of a plain string, a plugin entry can be an object with `src`, `name`, and `config` properties. This lets you pass plugin-specific configuration directly in `bsconfig.json`.
+
+```json
+{
+    "plugins": [
+        {
+            "src": "./scripts/myPlugin.js",
```
TwitchBronBron, line 32: @chrisdp wants this named `"plugin"` instead of `"src"`

docs/plugins.md
```diff
@@ -21,11 +21,349 @@ Those plugins will be loaded by the VSCode extension and can provide live diagno
...
+            "name": "my-plugin",
```
TwitchBronBron, line 33: @chrisdp wants this named `"alias"` instead of `"name"`

docs/plugins.md
```diff
@@ -21,11 +21,349 @@ Those plugins will be loaded by the VSCode extension and can provide live diagno
...
+
+# Mix with --plugins (plural) — names from --plugin take effect on those entries only
+npx bsc --plugins alpha beta charlie --plugin "@rokucommunity/bslint" bslint
+```
+
+The name you supply on the CLI is equivalent to the `name` field in a `bsconfig.json` plugin entry — it becomes the highest-priority identifier for `--plugin.<id>...` config overrides (see [How the `<id>` is resolved](#how-the-id-is-resolved) below).
+
+```bash
+# Load and name in one place, then override config keyed by that name
+npx bsc --plugin "@rokucommunity/bslint" bslint --plugin.bslint.severity=error
```
TwitchBronBron, line 162: @chrisdp wants to consider removing the equals from all these examples, since it's optional

PR #1712, opened 2026-05-14
Reduce per-Token lexer allocation to cut GC pressure while editing

chrisdp:
## Summary

Five small lexer/parser changes targeting per-Token wrapper allocations on lexer-bound builds. Net effect on a namespace-bound ukor build is within run-to-run noise; the wins should concentrate on codebases where the lexer is the dominant allocator (e.g. peacock, where the report from PR #1685 follow-up shows ~270 MB of lexer self-heap).

## Changes

1. **`FixedTokenText` canonical strings.** `TokenKind.ts` exposes a `Partial<Record<TokenKind, string>>` mapping ~42 punctuation/operator token kinds to their canonical text (`Dot` → `'.'`, `LeftParen` → `'('`, etc.). `Lexer.addToken` uses the canonical instance instead of allocating a fresh `source.slice` wrapper for these kinds. Excludes keywords (text varies by case in source) and identifiers / literals / comments (text is content-dependent).

2. **`LexerTextCache` for Newline + Whitespace.** Lazy intern table for the two kinds with bounded-but-not-fixed text variety (3 valid newline forms, ~50 typical indent patterns in real source). Module-scope, retention is trivial.

3. **`Lexer.whitespace()` fast path when `includeWhitespace=false`.** The default path previously built a full Whitespace Token (with `Range` tree of 5 nested objects) only to immediately `pop()` it from the stream. The fast path now computes the canonical text directly and updates `start`/`lineBegin`/`columnBegin` to match what `addToken`'s `sync()` would have done. Saves a Token + Range allocation per whitespace run on default builds. **This is the change most likely to move peak RSS on peacock-shaped projects.**

4. **EOF token field order.** `Lexer.scan()`'s EOF token construction now matches `addToken`'s field order (`kind, text, isReserved, range, leadingWhitespace`), keeping all lexer-produced Tokens on a single V8 hidden class.

5. **Parser-synthesized Token shapes.** Four synthesis sites in `Parser.ts` (the recovered `Function` token and three compound-assignment `Equal` tokens) are normalized to the lexer's 5-field shape with `isReserved: false` and `leadingWhitespace: ''` to keep V8 representation type consistent with the lexer's writes.

## Test plan

- [x] All 2910 specs pass
- [ ] Heap profile on a lexer-heavy codebase to confirm the per-token wrapper savings show up where expected — the build used during development is namespace-bound so the signal isn't as visible there
- [ ] Sanity check that consumers reading `token.range`, `token.text`, `token.leadingWhitespace` still behave identically (the lexer's externally-visible Token contract is unchanged; only internal allocation patterns changed)


TwitchBronBron (approved):
lgtm. once we test this, i'm good to merge.

TwitchBronBron:
I merged `master` into this branch (it was conflicting — `master` had independently landed the canonical Token field-order comment, so **changes #4 and #5 of this PR are already upstream** and contribute nothing here) and then spent a while trying to verify the perf claims. Summary: **the allocation claim holds and is solidly measurable; I could not reproduce any reliable CPU or peak-RSS benefit.** I also pushed one small fix and one dedup. Details and method below so you can check any of it.

## The allocation reduction is real and exactly reproducible

Wall-clock benchmarking of this change is a trap — I burned a lot of time on it and got contradictory answers. Allocation *counts*, though, are a pure function of the input, so they repeat bit-for-bit.

I instrumented `String.prototype.slice` and `util.createRange` (the lexer's only token-text and Range allocation sites) and counted calls. **24 corpus/mode cells × 3 builds × 2 runs → 0 nondeterministic readings.**

Across 7 real projects (playlet, maestro-roku, rooibos, rodash, SmartList, plus two `.brs`-only trees), 588,662 tokens:

| metric | master | this PR | delta |
|---|---|---|---|
| string allocations | 1,092,562 | 885,551 | **−18.9%** |
| Range allocations | 810,383 | 588,662 | **−27.4%** |

That's ~352 fewer strings and ~377 fewer Ranges per 1,000 tokens, and it holds on every real corpus (slices −13.5% to −20.3%, ranges −20.6% to −31.2%). A nice confirmation that change #3 does exactly what you designed it to: on `master` the Range count equals tokens + whitespace runs, and on this branch it equals tokens exactly — one Range tree per whitespace run, eliminated.

## But I could not find a CPU or RSS win

On a real full build (jellyfin-roku, which pins bsc 0.73.0, 418 files, bslint active, full lex→parse→validate→transpile, 7 trials, all arms verified identical at 418 files/75 diagnostics), every delta was smaller than the run-to-run noise: wall varied 17–24% *within* a single arm, and peak RSS swung 177→261MB within `master` alone. Median wall came out −1.9% for this PR, which I do not believe is signal.

Context worth having: on that build, parsing is ~724ms of ~2010ms wall, and lexing is only a slice of parsing. So even a large lexer change has limited headroom at build scale.

Running under `node --predictable` (deterministic GC scheduling, single-threaded, no concurrent JIT) tightened spreads to 1–9% and gave the clearest CPU picture I could get on real corpora — roughly break-even, mixed sign:

| corpus | lex | lex+parse |
|---|---|---|
| maestro-roku | +2.0% | −0.4% |
| SmartList | −0.7% | −0.8% |
| playlet | −0.2% | −0.5% |
| rooibos | −10.1% | −3.8% |

## One thing worth looking at: identifier-heavy source

The one CPU result that reproduced consistently and outside the noise is a regression. On an identifier/literal-heavy corpus, `lex` is **+13% to +16% vs master** (reproduced 4x, spreads 1.5–3.9%).

The mechanism is straightforward: `FixedTokenText` deliberately excludes identifiers, literals and keywords, so on identifier-dense code every token pays the table lookup and no token benefits. Identifiers are the most common token kind in real source, which is likely why the real-project rows above hover around break-even rather than showing the allocation win as time.

Fair caveat: that corpus is synthetic (I wrote it to isolate the miss path), so treat it as evidence about the *mechanism*, not a prediction for any real project. But it's the one signal that survived every attempt to make it go away.

## What I pushed

**1. Bounded `LexerTextCache` (`LEXER_TEXT_CACHE_MAX_ENTRIES = 512`).** Your comment estimates the table stays small, and that checks out — across 1078 real files it settles at **32 entries / 2.3 KB**. But it's module-scope and lives for the process, which matters for the language server. A 5000-line file with a distinct indent width per line grows it to **~8.9 MB** with nothing to ever release it (generated or minified source can do this). Interning is a pure optimization, so past the cap the lexer just keeps the sliced string. Allocation counts on real code are byte-identical with the cap in place (279,307 slices / 190,832 ranges on maestro-roku, unchanged), so it costs nothing in the normal case.

**2. Deduped the intern logic** into one `internText` helper — it was copy-pasted between `whitespace()` and `addToken()`.

**3. Four tests** covering instance sharing for fixed-text kinds, correct text for multi-char operators (`??`, `?.`, `@.`), whitespace interning, and cap behavior. 3042 passing, lint clean.

**One correction on my own work:** I initially pushed a null-prototype wrapper for `FixedTokenText`, claiming ~2.1x faster lookups from avoiding V8 dictionary-mode misses. That was wrong and I've reverted it (96b2bd3e). The 2.1x came from a micro-benchmark run *without* `--predictable`; under it, plain and null-prototype lookups are indistinguishable (1.516ms vs 1.514ms over 252k tokens), and the entire lookup is only ~6ms of a ~600ms lex — far too small to explain what I'd attributed to it. An A/B of just that line showed the plain object was actually *faster* on punctuation-dense source (381ms vs 579ms). Sorry for the noise in the branch history.

## Correctness: clean

Differential test across **1078 files × 3 option sets** (default, `includeWhitespace: true`, `trackLocations: false`), comparing full token streams — kind, text, isReserved, leadingWhitespace, and exact ranges — against `master`: **0 mismatches.** The `includeWhitespace: false` fast path replicates `sync()`'s three field updates exactly, and `Question` vs `Print` being distinct kinds means `?`-as-print stays correctly mapped.

## Recommendation

I'd land it, but on the allocation argument rather than the speed one, and I'd suggest:

1. **Reframe the PR description.** "Reduce per-Token allocation" is accurate and well-supported (−19% strings, −27% Ranges). The peak-RSS framing on change #3 I couldn't substantiate — that corpus showed the largest *Range* reduction (−41.5%) but no reliable RSS or wall-time movement.
2. **Drop changes #4 and #5** — already in `master`.
3. **Decide on the identifier-heavy case.** If the +13–16% concerns you, gating the lookup on a cheap precondition before touching the table (something that rules out identifiers/literals/keywords without a hash lookup) would preserve the allocation win and drop the miss cost. Happy to try it if useful.
4. The real payoff is likely the **language server**, not CLI builds: fewer allocations per token matters most in a long-lived process re-lexing files on every keystroke. I haven't measured that — it needs a steady-state incremental-reparse harness — so it remains the untested (and I'd guess strongest) case for this change.

### Reproducing any of this

The harness is a handful of standalone scripts (allocation counter, differential token-stream check, `--predictable` CPU runner, full-build runner, corpus builders). It's not in the branch since it's not shippable, but I'm glad to push it to a scratch branch or paste any individual script if you want to re-run or poke at the numbers. The two load-bearing ones are the alloc counter (deterministic, gives the −19%/−27%) and the differential checker (gives the 0-mismatch correctness result).

TwitchBronBron:
Merged `master` in and benchmarked this fairly thoroughly. **Verdict: land it.** The allocation claim holds, and I pushed a follow-up that turns the CPU picture from "wash" into a real win. Numbers and method below.

Heads up: **the "EOF token field order" and "Parser-synthesized Token shapes" changes are already in master** — it landed the canonical Token field-order change independently, which is what caused the conflict. They contribute nothing here now.

## Allocation reduction: verified, zero variance

Wall-clock benchmarking of this change is unreliable — I burned a lot of time on it before giving up on it. Allocation *counts* are a pure function of the input, so they repeat bit-for-bit. Instrumenting `String.prototype.slice` and `util.createRange`: **24 cells × 3 builds × 2 runs → 0 nondeterministic readings.**

7 real projects, 588,662 tokens:

| metric | master | this PR | delta |
|---|---|---|---|
| string allocations | 1,092,562 | 885,551 | **−18.9%** |
| Range allocations | 810,383 | 588,662 | **−27.4%** |

The `whitespace()` fast path works exactly as designed: master's Range count equals `tokens + whitespace runs`; here it equals `tokens` precisely — one Range tree per whitespace run, eliminated.

## GC pressure while editing: −18% pauses

This is where the allocation win actually cashes out. The LSP re-lexes the **whole file** on every `setFile()` (`BrsFile.parse()` → `Lexer.scan()`, no incremental lexing), so every keystroke re-lexes everything.

Simulated editing session (jellyfin-roku, 232 files, 150 edits, 5 trials/arm):

| metric | master | this PR | delta |
|---|---|---|---|
| **GC count** | 390 | 320 | **−17.9%** |
| **GC pause time** | 1928 ms | 1470 ms | **−23.8%** |

Master ran 385/393/390/390/390; this PR ran 320/321/319/320/322. Non-overlapping distributions, 1–2% spread.

Total *bytes* allocated barely moves (−0.7%) while GC *count* drops 18% — expected, since the scavenger is driven by allocation count and young-gen fill rate more than volume, and this removes many small short-lived objects.

## Retained footprint: real but small

1,667 distinct-content files from a large production codebase, 2,336,914 tokens, 7 trials/arm:

| | master | this PR | delta |
|---|---|---|---|
| **lex** | 470.87 MB | 459.83 MB | **−11.05 MB (−2.35%)** |
| **lex+parse** | 777.96 MB | 767.09 MB | **−10.88 MB (−1.40%)** |

sd 0.02–0.09 MB, distributions fully separated. Genuine, but against a ~470–780 MB baseline — if LSP memory on big projects is a concern, that 767 MB is the target and this isn't the fix. It's dominated by the Tokens/Ranges that are *kept*, not the wrappers this removes.

## No accumulation over time

The intern table is module-scope and process-lifetime, so worth checking: lexing and discarding **11,676 file-lexes of real code** (2,919 files, 19.7 MB, 4 passes) leaves heap flat (0.31 → 0.39 MB) with the cache plateaued at **37 entries** — identical to master. It's keyed by whitespace *content*, and real source uses only a few dozen distinct indent strings regardless of file count.

## Follow-up I pushed: order `addToken` by real token frequency

`addToken` probed `FixedTokenText` for every token, then fell back to a slice. But `Identifier` is the single most common kind — **27.5% of ~2.9M tokens** measured across production and open-source BrightScript — and is deliberately absent from that table. So the hottest path in the lexer was paying a hash lookup that could never hit.

Replaced the probe with a switch ordered by measured frequency (`Identifier`, `Newline`, then `Dot`/`L`/`RParen`/`Equal`/`Comma`/`Colon` — together ~75% of all tokens).

Two things worth knowing, both measured:

- **V8 does not build a jump table for a string switch** (those need dense integers) — it emits a chain of pointer-equality compares. So case *order* is load-bearing: the same logic with cold cases first measured **1.8× slower** than hot-first.
- **`Newline` gets its own case.** Only `\n`, `\r\n`, `\r` are possible, so a direct compare chain replaces the Map — **6–8× cheaper per newline**. Which form is tested first is noise (~0.02 ms per 90k newlines), so it doesn't favor LF or CRLF; that mix varies by platform and git autocrlf.

`lex` results (`node --predictable`, min-of-5, 7 trials, arm order rotated):

| corpus | master | before switch | after switch | vs before |
|---|---|---|---|---|
| playlet | 113.5 | 113.5 | 96.2 | **−15.3%** |
| SmartList | 109.4 | 112.1 | 95.4 | **−15.0%** |
| maestro-roku | 113.8 | 118.7 | 102.0 | **−14.1%** |
| 1,667 real production files | 1337.2 | 1216.0 | 1125.8 | **−7.4%** |

`lex+parse` on that production corpus: 3111 ms → 2545 ms (**−18.2%**).

This also erases the small regressions the allocation work had introduced on maestro-roku (+4.4%) and SmartList (+2.5%). Allocation counts are unchanged (279,307 slices / 190,832 ranges), so it's purely CPU — the allocation win is fully preserved.

One honest note: a synthetic identifier-heavy corpus regresses ~9%. I bisected it — it's not the `Newline` or `Identifier` case, it's the switch structure itself. That corpus sends **38% of tokens to `default`** (it's generated from only 13 token kinds) versus **22–26% for real code** (~127 kinds). Tokens reaching `default` pay all 9 failed compares first, so the trade is explicit: hot kinds get faster, `default` gets slightly slower, and real code is ~75% hot kinds. It's an artifact of a corpus I wrote to stress one path, not a real-world shape.

## Also pushed

- **Bounded `LexerTextCache`** (cap 512). Your "stays small" estimate is right for real code (32–37 entries), but it's process-lifetime: a file with a distinct indent width per line grows it to **~8.9 MB** with nothing to release it. Interning is a pure optimization, so past the cap the lexer keeps the sliced string — allocation counts on real code are byte-identical.
- **Deduped the intern logic** into one `internText` helper (was copy-pasted between `whitespace()` and `addToken()`).
- **Four tests** — instance sharing, multi-char operator text (`??`, `?.`, `@.`), whitespace interning, cap behavior.

**Correction on my own work:** I pushed then reverted (96b2bd3e) a null-prototype wrapper for `FixedTokenText`, claiming ~2.1× faster lookups. That measurement was taken without `--predictable` and didn't survive re-testing — plain and null-proto are indistinguishable, and an A/B of just that line showed the plain object was actually *faster* on punctuation-dense source. Sorry for the churn in the history.

## Correctness

**5,664 files × 3 option sets** (default, `includeWhitespace: true`, `trackLocations: false`), comparing full token streams — kind, text, isReserved, leadingWhitespace, exact ranges — against master: **0 mismatches.** 3042 specs passing, lint clean.

## Suggested before merge

**Reframe the description** — drop the peak-RSS claim on the `whitespace()` fast path (I couldn't substantiate it; that corpus gave the largest *Range* reduction at −41.5% but no reliable RSS movement), and drop the two changes that are already upstream. The supportable summary is: −19%/−27% allocations, −18% GC pauses while editing, and −14–15% lexer CPU on real code.

TwitchBronBron (approved):

PR #1732, opened 2026-06-15
Adds `strict` mode.

markwpearce:
Adds `strict` mode option to `bsconfig.json`, set by default to `false`

```
"Enables a set of stricter type-checking rules and language features. When true, unknown members on Component types will be treated as errors instead of dynamic. Defaults to false."
```

What does `strict` mode do?

- **Validates unknown members on Nodes:** Normally, unknown members on nodes are typed as `dynamic`, because Node member types can be updated (eg. through `addField`, `update`, etc). When Strict mode is on, unknown members will not be resolved
- **Validates unknown CallFunc on `roSGNode`:** If strict mode is on, trying to do a `callfunc` on a generic node will give a `cannot-find-callfunc` error. By default, this is simply not validated.

If strict mode is on, it forces the developer to typecast the Nodes so they have known members.

- [x] Follow the typescript `strict` pattern, where there are multiple options that can be turned on/off. 

EG: In Typescript, there's `strict` but also `strictBuiltinIteratorReturn`, `strictFunctionTypes` etc.

`strict` sets a default value for the rest of the `strict*` settings... so you can do:
```
"compilerOptions": {
   "strict": true,
   "strictNullChecks": false
}
```

which turns on all `strict*` settings, but turns off `strictNullChecks` ... 

Add the following `strict` settings:

- [x] `strictNodeMembers` will validate unknown members on Nodes
- [x] `strictCallFunc` will disallow doing a callfunc on a generic node


src/DiagnosticFilterer.ts
```diff
@@ -71,7 +71,7 @@ export class DiagnosticFilterer {
      * Should this file be completely ignored?
      * If the file diagnostics are ignored, we do not need to validate this file
      */
-    public isFileFiltered(file: { srcPath: string; destPath: string }) {
+    public isFileCompletelyFiltered(file: { srcPath: string; destPath: string }) {
```
markwpearce, line 74: Changed name to make function more clear

src/DiagnosticManager.ts
```diff
@@ -478,13 +478,17 @@ export class DiagnosticManager {
         }
     }
 
-    public shouldFilterFile(file: BrsFile): boolean {
+    /**
+     * Are the diagnostics for this file completely filtered?
+     * If so, we can skip any Scope-based validation on this file at all, which can save a lot of time for large files
+     * with many diagnostics that are being ignored
+     */
+    public canSkipScopeValidationForFile(file: BrsFile): boolean {
```
markwpearce, line 486: changed name to make function use more clear

TwitchBronBron (approved):

PR #1741, opened 2026-07-07
Add SceneGraph XML element and attribute completions

chrisdp:
Bring first-class SceneGraph XML completions into brighterscript so `.xml` component files get completions out of the box (previously only `<script uri>` paths completed). Driven by the built-in `roku-types` node metadata plus the project's own components; no scraped data is bundled.

- Element (tag name) completions after `<` and in element content: built-in nodes and project components.
- Field (attribute name) completions inside an open start tag, resolved across the node's full `extends` chain; attributes already present are excluded and read-only built-in fields are omitted.
- New `Program` methods `getSceneGraphNodeNames`, `getSceneGraphNodeFields`, `hasSceneGraphNode`, `getSceneGraphNode`.
- `<` added as a completion trigger character.
- `SGParser` no longer throws on a malformed child element (e.g. a lone `<` while typing).

First of three stacked PRs: completions, then validation, then hover.

### DIFF
--- a/src/LanguageServer.ts
+++ b/src/LanguageServer.ts
@@ -219,8 +219,8 @@ export class LanguageServer {
                 // Tell the client that the server supports code completion
                 completionProvider: {
                     resolveProvider: false,
-                    //anytime the user types a period, auto-show the completion results
-                    triggerCharacters: ['.'],
+                    //`.` auto-shows brightscript completions; `<` auto-shows xml element completions
+                    triggerCharacters: ['.', '<'],
                     allCommitCharacters: ['.', '@']
                 },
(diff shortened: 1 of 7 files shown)


TwitchBronBron:
Missing `<function` completion. (the one `<function>` in there is a snippet
image

TwitchBronBron:
Incorrectly completing with `f` (which is not a valid thing). 

image

TwitchBronBron:
All the components should only be available from within `<children>` not at the top level. 
image

TwitchBronBron:
missing attributes for <field> and <function>  (was expecting to see `type` and `id` and such).
image

PR #1742, opened 2026-07-07
Validate SceneGraph node types and fields in xml components

chrisdp:
Validate SceneGraph node usage in `.xml` component files by walking each component's `<children>` node tree.

- Unknown component type: error (consistent with the existing `roSGNode` validation).
- Unknown field/attribute name on a known node: error.
- Field-name casing that differs from the declared field: warning.
- Clearly-invalid scalar field value: warning.

Value and field checks are conservative: only int/float/bool/color values are validated (everything else passes), and field-name checks are skipped when no fields resolve for a known node, to avoid false positives. New diagnostics use codes 1155-1158.

Stacked on #1741.


src/bscPlugin/validation/XmlFileValidator.ts
```diff
@@ -60,6 +61,67 @@ export class XmlFileValidator {
...
+     * Validate a single node instance and recurse into its children. Flags unknown node types, and (for
+     * known types) unknown/mis-cased field names and clearly-invalid field values.
+     */
+    private validateNode(node: SGNode) {
+        const nodeName = node.tag.text;
+        if (this.event.program.hasSceneGraphNode(nodeName)) {
+            this.validateNodeAttributes(node, nodeName);
+        } else {
+            this.event.file.diagnostics.push({
+                ...DiagnosticMessages.xmlUnknownComponentType(nodeName),
```
TwitchBronBron, line 82: This works great for _known_ xml files. However, for devs that use component libraries, all of their components are going to show up as errors. I think for now, we should avoid adding this diagnostic for `ComplibPrefix:Whatever` pattern components, until we have a way of declaring these in type definitions.

src/bscPlugin/validation/XmlFileValidator.ts
```diff
@@ -60,6 +61,67 @@ export class XmlFileValidator {
...
+                file: this.event.file
+            });
+        }
+        //recurse regardless, so nested typos are still reported
+        for (const child of node.children ?? []) {
+            this.validateNode(child);
+        }
+    }
+
+    private validateNodeAttributes(node: SGNode, nodeName: string) {
```
TwitchBronBron, line 93: This works great for out-of-the-box attributes. But many dev teams have custom plugins that do things like handlebars replacements. Those would all be marked as warnings (or errors) with no way to granularly say "brighterscript can validate these, I will validate those others". maybe a plugin hook for `validateSGAttribute` or something that lets plugins tap in and claim an attribute before falling back to brighterscript?

src/DiagnosticMessages.ts
```diff
@@ -833,6 +833,26 @@ export let DiagnosticMessages = {
...
+    }),
+    xmlFieldNameCaseMismatch: (actualName: string, expectedName: string) => ({
+        message: `Field '${actualName}' should be '${expectedName}' to match the declared casing`,
+        code: 1157,
+        severity: DiagnosticSeverity.Warning
+    }),
+    xmlInvalidFieldValue: (fieldName: string, expectedType: string) => ({
+        message: `Value for field '${fieldName}' is not a valid '${expectedType}'`,
+        code: 1158,
+        severity: DiagnosticSeverity.Warning
```
TwitchBronBron: This should be an error

src/DiagnosticMessages.ts
```diff
@@ -833,6 +833,26 @@ export let DiagnosticMessages = {
...
+    xmlUnknownField: (fieldName: string, componentName: string) => ({
+        message: `Field '${fieldName}' does not exist on component '${componentName}'`,
+        code: 1156,
+        severity: DiagnosticSeverity.Error
+    }),
+    xmlFieldNameCaseMismatch: (actualName: string, expectedName: string) => ({
+        message: `Field '${actualName}' should be '${expectedName}' to match the declared casing`,
+        code: 1157,
+        severity: DiagnosticSeverity.Warning
+    }),
```
TwitchBronBron: This diagnostic belongs in eslint. it's a preference thing, not something that would cause runtime errors.

src/DiagnosticMessages.ts
```diff
@@ -833,6 +833,26 @@ export let DiagnosticMessages = {
         message: `rsg_version=${rsgVersion} was removed in firmware ${removedAt}; use rsg_version=${replacement}`,
         code: 1154,
         severity: DiagnosticSeverity.Error
+    }),
+    xmlUnknownComponentType: (nodeName: string) => ({
+        message: `Unknown component type '${nodeName}'`,
+        code: 1155,
+        severity: DiagnosticSeverity.Error
+    }),
+    xmlUnknownField: (fieldName: string, componentName: string) => ({
```
TwitchBronBron, line 842: We need additional diagonstics for mismatched starting and ending xml attribute names

Image

Image

src/bscPlugin/validation/XmlFieldTypeValidator.ts
TwitchBronBron, line 1: Rename this to include xml in the file name. Maybe XmlFieldTypeValidator or something? We have lots of other validation files floating around in this folder.

TwitchBronBron (changes requested):

src/bscPlugin/validation/XmlFileValidator.ts
TwitchBronBron: I think the approach we came up with is some type of validation-only hook that lets plugins say "yeah, i know what that is...don't validate it".

PR #1744, opened 2026-07-07
Support conditional compilation (`#if`) inside class bodies

chrisdp:
## Problem

In v0, conditional compilation was handled by a preprocessor that ran before parsing, so `#if` blocks worked anywhere — including class bodies. In v1, `#if` became a first-class AST statement (`ConditionalCompileStatement`) handled by the parser at block and file-root level, but `classDeclaration()` was never taught about it. Code like this (valid in v0, common in test suites and debug tooling) produces a cascade of parse errors in v1 (`expected-end-keyword`, `Unexpected token '#end if'`, `unexpected-statement`, etc.):

```brighterscript
class Person
    sub speak()
    end sub

    #if DEBUG
        sub debugSpeak()
            print "debug"
        end sub
    #end if
end class
```

## Changes

- **Parser**: class bodies now accept `#if` blocks. The class-member parsing was extracted into `classMemberDeclaration()` and reused by a new `classBodyConditionalCompileBlock()`, which parses branch contents in a class-member context (methods, fields, annotations, nested `#if`). It plugs into the existing `conditionalCompileStatement()` machinery via an injected branch parser, so `#else if` chains, `not` conditions, false-branch diagnostic suppression, and the token-position contract all behave exactly like function-body `#if`.
- **ClassStatement**: `methods`/`fields`/`memberMap` now register members found inside conditional compile branches, so the type system and tooling can see them (e.g. calling a conditional method via `m.` from another method validates cleanly).
- **Transpile**:
  - methods and builder assignments are emitted wrapped in the equivalent `#if`/`#else if`/`#else`/`#end if` directives (resolved by the Roku compiler from manifest `bs_const`, same as function-body `#if`)
  - field initializers injected into the constructor are wrapped in an equivalent `ConditionalCompileStatement`
  - field-only branches don't emit empty `#if`/`#end if` shells in the methods/builder sections

Example transpile output:

```brightscript
sub __Animal_method_new()
    m.name = "generic"
    #if DEBUG
        m.logLevel = 4
    #else
        m.logLevel = 0
    #end if
end sub
#if DEBUG
    sub __Animal_method_debugSpeak()
        print "debug"
    end sub
#end if
function __Animal_builder()
    instance = {}
    instance.new = __Animal_method_new
    #if DEBUG
        instance.debugSpeak = __Animal_method_debugSpeak
    #end if
    return instance
end function
```

## Known limitations

- Constructors (`sub new()`) inside `#if` blocks are not supported — builder/class-function generation assumes a single known constructor signature (unchanged from current behavior).
- `getTypedef()` does not yet emit conditional members into `.d.bs` typedefs.

## Testing

- 12 new tests (8 parser, 4 transpile/validation): methods, fields, `#else if` chains, nesting, empty blocks, annotations on conditional members, class-type membership, unterminated-block diagnostics
- full suite passes (4264 passing, 0 failures), lint clean
- validated against a large production codebase (fubo's Roku app) where the v0→v1 migration surfaced this: the parse-error cascade is resolved with the full plugin stack loaded

### DIFF
--- a/src/files/BrsFile.Class.spec.ts
+++ b/src/files/BrsFile.Class.spec.ts
@@ -2151,4 +2151,158 @@ describe('BrsFile BrighterScript classes', () => {
             DiagnosticMessages.mismatchArgumentCount(1, 0)
         ]);
     });
+
+    describe('conditional compile', () => {
+        it('transpiles methods inside conditional compile blocks', async () => {
+            await testTranspile(`
+                #const DEBUG = true
+                class Animal
+                    sub speak()
+                        print "speak"
+                    end sub
+                    #if DEBUG
+                        sub debugSpeak()
+                            print "debug"
+                        end sub
+                    #end if
+                end class
+            `, `
+                #const DEBUG = true
+                sub __Animal_method_new()
+                end sub
+                sub __Animal_method_speak()
+                    print "speak"
+                end sub
+                #if DEBUG
+                    sub __Animal_method_debugSpeak()
+                        print "debug"
+                    end sub
+                #end if
+                function __Animal_builder()
+                    instance = {}
+                    instance.new = __Animal_method_new
+                    instance.speak = __Animal_method_speak
+                    #if DEBUG
+                        instance.debugSpeak = __Animal_method_debugSpeak
+                    #end if
+                    return instance
+                end function
+                function Animal()
+                    instance = __Animal_builder()
+                    instance.new()
+                    return instance
+                end function
+            `, 'trim', 'source/main.bs');
+        });
+
+        it('transpiles fields inside conditional compile blocks as conditional initializers', async () => {
+            await testTranspile(`
+                #const DEBUG = true
+                class Animal
+                    name = "generic"
+                    #if DEBUG
+                        logLevel = 4
+                    #else
+                        logLevel = 0
+                    #end if
+                end class
+            `, `
+                #const DEBUG = true
+                sub __Animal_method_new()
+                    m.name = "generic"
+                    #if DEBUG
+                        m.logLevel = 4
+                    #else
+                        m.logLevel = 0
+                    #end if
+                end sub
+                function __Animal_builder()
+                    instance = {}
+                    instance.new = __Animal_method_new
+                    return instance
+                end function
+                function Animal()
+                    instance = __Animal_builder()
+                    instance.new()
+                    return instance
+                end function
+            `, 'trim', 'source/main.bs');
+        });
+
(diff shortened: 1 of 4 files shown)


markwpearce (commented):
Ran a review pass on this, a few things worth a look:

**`src/parser/Statement.ts:2849` (`getConstructorFunction`)** - this only looks at the top-level `this.body`, so it doesn't find a `new()` that's only declared inside a `#if` block. Combined with `registerMembers` now descending into `#if` branches, you can end up with a class like:

```
class Animal
    #if DEBUG
        sub new()
        end sub
    #end if
end class
```

`getConstructorFunction()` says "no constructor exists", so a synthetic empty one gets injected... but the conditional one still transpiles too. Actually tried this - when DEBUG is true you get two `sub __Animal_method_new()` defs in the output. That's not just a "known limitation", it's invalid brs that won't compile on-device, and nothing flags it.

**`src/parser/Parser.ts:728` (`classMemberDeclaration`)** - the early return for `#if` happens before the annotation-consumption logic at the bottom of the function. So an annotation right before a `#if` block doesn't attach to anything inside it - it just leaks onto whatever comes after the block instead. Confirmed this: `@deprecated` before a `#if`-wrapped method silently attaches to the *next* method instead, no diagnostic at all. Worse than the "annotation not attached to anything" case since there's zero signal something went wrong.

**`src/parser/Statement.ts:2692` (`registerMembers`) / `:3087` (`getTranspiledConditionalCompileMembers`)** - more of a maintainability thing than a bug: the `#if`/`#else if`/`#else` branch tree gets walked by three separate hand-rolled recursive functions now (registration, transpile, field-initializer injection). Any future fix to how elseBranch chains work has to be applied in all three or one of them silently breaks. Also `registerMembers` flattens conditional members into `methods`/`fields`/`memberMap` once at construction - if `body` gets mutated later (plugins, editor splices) those arrays go stale.

Not blocking necessarily, but the first two are silent-bad-output bugs, would want those fixed or at least diagnosed before merge.

markwpearce:
Merged v1 in - one real conflict in Statement.ts's import list, nothing structural.

Addressed the review:
- constructor inside `#if` now errors instead of silently duplicating `new` in the transpiled output - `getConstructorFunction()` checks `memberMap` now so it stops injecting a synthetic ctor on top of a conditional one
- annotation before a `#if` now attaches to the `ConditionalCompileStatement` instead of leaking onto the next member
- collapsed the three `#if`/`#else if`/`#else` branch-walks into one `forEachConditionalCompileBranch` helper

4585 tests passing, added coverage for the constructor diagnostic and the annotation fix.

PR #1753, opened 2026-07-21
Flag path-like codes in diagnosticFilters

### LINKED ISSUE #1384: Flag path-like codes in diagnosticFilters
Now that https://github.com/rokucommunity/brighterscript/issues/1060 merged, we should add some diagnostics to flag when the old style diagnosticFilters pattern looks like a file path instead of a code. This might not be 100% accurate (plugins could do some crazy codes if they wanted to), but it should help almost all the time. 

This is going to be a challenge for projects that are trying to migrate from v0 to v1 because their diagnostic filters are no longer compatible. 

For example:

```json
{
    "diagnosticFilters": [
         "source/vendor/**/*",
    ]
}
```

Should be flagged because that is very clearly a glob, and those need to now be in the format:
```json
{
    "diagnosticFilters": [{
        "files": [
            "source/vendor/**/*"
        ]
    }]
}
```

We could probably flag the following:
- if there is a `./` or `../`
- if there is a globstar `**`
- if there is a `.bs` or `.brs` or `.xml` file extension at the end

markwpearce:
## Summary
- Adds a new diagnostic that warns when a v1-style `diagnosticFilters` entry looks like a file path or glob (contains `./`, `../`, `**`, or ends in `.bs`/`.brs`/`.xml`) rather than a diagnostic code — a common mistake when migrating a v0-style bsconfig, where `diagnosticFilters` entries were file globs.
- Skipped when `diagnosticFiltersV0Compatibility` is enabled, since those filters are legitimately file globs.

Closes #1384

## Test plan
- [x] Added unit tests for `DiagnosticFilterer.getPathLikeDiagnosticFilterCodes`
- [x] Added an integration test in `Program.spec.ts` verifying the diagnostic is emitted/not emitted during `program.validate()`
- [x] `npm run build`, `npm run lint`, and `npm run test:nocover` all pass

### DIFF
--- a/src/DiagnosticFilterer.ts
+++ b/src/DiagnosticFilterer.ts
@@ -421,15 +421,20 @@ export class DiagnosticFilterer {
     /**
      * Does this diagnostic filter "code" value look like a file path/glob instead of a diagnostic code?
      * v0-style `diagnosticFilters` entries were file globs, but v1 treats bare string/number entries as codes.
-     * This heuristic catches the common glob patterns left over from a v0-style config.
+     * This heuristic catches the common glob patterns left over from a v0-style config, as well as any value
+     * that exactly matches the destPath of a file actually in the project (if `knownDestPaths` is provided).
      */
-    public isCodeValuePathLike(value: number | string): boolean {
+    public isCodeValuePathLike(value: number | string, knownDestPaths?: Set<string>): boolean {
         if (typeof value !== 'string') {
             return false;
         }
         const lowerValue = value.toLowerCase();
+        if (knownDestPaths?.has(lowerValue.replace(/\\/g, '/'))) {
+            return true;
+        }
         return (
             value.includes('./') ||
+            value.includes('.\\') ||
             value.includes('**') ||
             lowerValue.endsWith('.bs') ||
             lowerValue.endsWith('.brs') ||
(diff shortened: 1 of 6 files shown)


src/DiagnosticFilterer.ts
```diff
@@ -417,4 +417,48 @@ export class DiagnosticFilterer {
...
+     * v0-style `diagnosticFilters` entries were file globs, but v1 treats bare string/number entries as codes.
+     * This heuristic catches the common glob patterns left over from a v0-style config.
+     */
+    public isCodeValuePathLike(value: number | string): boolean {
+        if (typeof value !== 'string') {
+            return false;
+        }
+        const lowerValue = value.toLowerCase();
+        return (
+            value.includes('./') ||
```
TwitchBronBron, line 436: what about windows? `.\`

src/DiagnosticManager.ts
```diff
@@ -489,6 +490,25 @@ export class DiagnosticManager {
         }
         return this.diagnosticFilterer.isFileCompletelyFiltered(file);
     }
+
+    /**
+     * Flag `diagnosticFilters` entries that look like file paths/globs rather than diagnostic codes.
+     * This is a common mistake when migrating a bsconfig.json from the v0-style filters (which were file globs)
+     */
+    public detectPathLikeDiagnosticFilterCodes(config: FinalizedBsConfig, context?: DiagnosticContext) {
+        const pathLikeCodes = this.diagnosticFilterer.getPathLikeDiagnosticFilterCodes(config);
```
TwitchBronBron: Might be worth checking to see if the pattern is exactly a path to a known file in the files array? like `source/main.json` or something?

TwitchBronBron (changes requested):

src/DiagnosticFilterer.ts
markwpearce: Good catch — fixed. `isCodeValuePathLike` now also checks for the Windows-style `.\\` relative-path separator in addition to `./`.

src/DiagnosticManager.ts
markwpearce: Done — `getPathLikeDiagnosticFilterCodes`/`isCodeValuePathLike` now accept an optional set of known project file destPaths, and `DiagnosticManager` passes through the destPaths of all files currently loaded in the program. A filter entry that exactly matches one of those (e.g. a bare `source/main.json`) is flagged even when it doesn't match the glob-like heuristics.

TwitchBronBron: Great. Did you push the changes? I'm not sure I'm seeing them.

TwitchBronBron (approved):

PR #1776, opened 2026-08-17
Cap LSP worker thread pool to fix memory scaling with project count

markwpearce:
## Problem

Opening a multi-root VSCode workspace with many BrighterScript projects (e.g. ~15+, as in `brighterscript-game-engine`'s 17 `bsconfig.json` files) can crash the language server with an out-of-memory error.

Root cause: `LanguageServer.enableThreadingDefault` defaults to `true`, and `ProjectManager` spawns one dedicated OS worker thread (a full separate V8 isolate) per project via `WorkerPool.getWorker()`, which had no cap — it always created a new worker if none were free, and a worker was only returned to the pool when its project was disposed (which doesn't happen for open, in-use projects during a normal session). Each isolate independently loads the whole `brighterscript` module graph (`roku-types` static data, `globalCallables`, type caches, etc.), so total memory scaled roughly linearly with project count, multiplying fixed interpreter/type-system overhead by N.

## Fix

Replaces `WorkerPool`'s exclusive one-worker-per-project model with a capped, load-aware assignment model:

- `WorkerPool` now has a configurable `maxWorkers` (default `os.cpus().length`). Each project still gets its own `Project`/`Program` instance and its own dedicated `MessagePort` (via `MessageChannel`), but once the cap is reached, additional projects' ports are attached to an existing worker instead of spawning a new one.
- Co-located projects share that worker's JS realm, recovering cross-project sharing of the big fixed-cost singletons that per-project threading had made impossible to share.
- A worker is terminated and dropped from the pool as soon as its last attached project releases it, so memory is reclaimed promptly rather than kept warm indefinitely.
- New `languageServer.maxProjectWorkers` setting, mirroring the existing `projectActivationConcurrencyLimit` pattern, so this is tunable per-workspace.
- An unexpected worker exit now surfaces a `critical-failure` notification (previously: requests on that worker would hang forever with no signal to the user).
- Worker lifecycle edge cases handled: idempotent dispose (a project can be disposed more than once without side effects on sibling projects sharing its worker), a crashed worker is evicted from the pool's bookkeeping (so it can't silently swallow a future project's bootstrap), `preload()`/`assignProject()` agree on the cap, and `assignProject()` defends against a misconfigured `maxWorkers <= 0`.

## Testing

- New/updated unit tests in `WorkerPool.spec.ts`, `WorkerThreadProject.spec.ts` (including real-worker-thread integration tests proving multiple projects can share one worker while remaining functionally independent, and that a co-tenant surviving a sibling's dispose/double-dispose keeps working), `MessageHandler.spec.ts`, and `LanguageServer.spec.ts`.
- Full suite: 3001 passing, 0 failing. `npm run lint` and `npm run build` clean.
- Not automated: real-world confirmation against `brighterscript-game-engine`'s multi-project workspace in VSCode (recommended as a manual follow-up check).

## Notes for reviewers

- This targets `master` (not `v1`) because the affected code (`src/lsp/worker/*`) is currently identical between the two branches — landing here first means it flows to `v1` on the next `master` merge, rather than requiring a manual backport.
- Scoped out of this PR: the separate `Program.ts` cache-leak fixes identified during investigation (`fileClusters`, `getFilePathCache`, `symbolDependencies`, `util.ts` path caches never pruned) — smaller, unrelated issue, tracked separately.
- Also scoped out: automatic reload of sibling projects when a shared worker crashes — no existing mechanism reloads a project on `critical-failure` today; this PR keeps that event's existing one-way-notification behavior rather than introducing new reload semantics.

### DIFF
--- a/src/LanguageServer.ts
+++ b/src/LanguageServer.ts
@@ -80,9 +80,10 @@ export class LanguageServer {
 
     /**
      * The default maximum number of worker threads to use for running LSP projects. Can be overridden by
-     * per-workspace settings. Multiple projects will share a single worker thread once this limit is reached.
+     * per-workspace settings. Once this limit is reached, additional projects are spread evenly across the
+     * existing worker threads instead of each getting a dedicated one.
      */
-    public static maxProjectWorkersDefault = Math.max(1, os.cpus().length);
+    public static maxWorkerThreadsDefault = Math.max(1, os.cpus().length);
 
     /**
      * The language server protocol connection, used to send and receive all requests and responses
@@ -295,7 +296,7 @@ export class LanguageServer {
         await this.syncLogLevel();
 
         this.syncProjectActivationConcurrencyLimit();
-        this.syncMaxProjectWorkers();
+        this.syncMaxWorkerThreads();
 
         try {
             if (this.hasConfigurationCapability) {
@@ -419,27 +420,27 @@ export class LanguageServer {
     }
 
     /**
-     * Get the max project workers setting from all workspaces and set the worker pool's cap to the lowest value found.
+     * Get the max worker threads setting from all workspaces and set the worker pool's cap to the lowest value found.
      * This ensures that if the user has multiple workspaces open with different limits,
      * we respect the most restrictive limit to avoid overwhelming the user's machine.
      */
-    private syncMaxProjectWorkers() {
+    private syncMaxWorkerThreads() {
         const limits = [...this.workspaceConfigsCache]
-            .map(x => x?.[1]?.languageServer?.maxProjectWorkers)
+            .map(x => x?.[1]?.languageServer?.maxWorkerThreads)
             .filter(x => typeof x === 'number');
 
         //if we don't have any limits defined, use our default value
         if (limits.length === 0) {
-            limits.push(LanguageServer.maxProjectWorkersDefault);
+            limits.push(LanguageServer.maxWorkerThreadsDefault);
         }
 
-        let maxProjectWorkers = Math.min(...limits);
+        let maxWorkerThreads = Math.min(...limits);
         //we must always support at least 1 worker, otherwise no threaded projects could ever activate
-        if (!(maxProjectWorkers >= 1)) {
-            this.logger.log(`maxProjectWorkers was set to ${maxProjectWorkers}, which is not a valid value. Defaulting to 1.`);
-            maxProjectWorkers = 1;
+        if (!(maxWorkerThreads >= 1)) {
+            this.logger.log(`maxWorkerThreads was set to ${maxWorkerThreads}, which is not a valid value. Defaulting to 1.`);
+            maxWorkerThreads = 1;
         }
-        workerPool.maxWorkers = maxProjectWorkers;
+        workerPool.maxWorkers = maxWorkerThreads;
     }
 
     @AddStackToErrorMessage
@@ -597,7 +598,7 @@ export class LanguageServer {
                         projectDiscoveryExclude: brightscriptConfig?.languageServer?.projectDiscoveryExclude,
                         logLevel: brightscriptConfig?.languageServer?.logLevel,
                         projectActivationConcurrencyLimit: brightscriptConfig?.languageServer?.projectActivationConcurrencyLimit,
-                        maxProjectWorkers: brightscriptConfig?.languageServer?.maxProjectWorkers
+                        maxWorkerThreads: brightscriptConfig?.languageServer?.maxWorkerThreads
                     }
                 };
             })
@@ -640,7 +641,7 @@ export class LanguageServer {
             this.workspaceConfigsCache = configs;
 
             this.syncProjectActivationConcurrencyLimit();
-            this.syncMaxProjectWorkers();
+            this.syncMaxWorkerThreads();
 
             //if configuration changed, rebuild the path filterer
             await this.rebuildPathFilterer();
@@ -1068,7 +1069,7 @@ export interface BrightScriptClientConfiguration {
         logLevel: LogLevel | string;
         projectDiscoveryMaxDepth?: number;
         projectActivationConcurrencyLimit?: number;
-        maxProjectWorkers?: number;
+        maxWorkerThreads?: number;
     };
 }
 
--- a/src/LanguageServer.ts
+++ b/src/LanguageServer.ts
@@ -137,6 +137,13 @@ export class LanguageServer {
             this.sendDiagnostics(event).catch(logAndIgnoreError);
         });
 
+        //if a project's worker thread crashes unexpectedly, notify the client so the user isn't left staring at a frozen project with no explanation
+        this.projectManager.on('critical-failure', (event) => {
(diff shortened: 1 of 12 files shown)


markwpearce:
This change makes a substantial difference - this allows opening the Roku Samples repo (~139 projects) without issues.

Additionally, this same fix in `v1` *also* allows opening that repo in VSCode, with no noticeable lag... Taking about 20 seconds to load and validate all the projects.

markwpearce:
I don't love that MacOS CI tests take longer, but that's due to Github's test runner hardware, I believe.

On my laptop (m3 Macbook pro) `npm run test`:
- previous master  - 8s
- this branch - 12s


Creating Worker threads is just slow, unfortunately.

src/lsp/worker/WorkerThreadProject.ts
```diff
@@ -149,10 +153,37 @@ export class WorkerThreadProject implements LspProject {
...
+     * treat it as a critical failure so the user finds out instead of every request just hanging forever.
+     */
+    private handleWorkerExit = (code: number) => {
+        if (this.isDisposed) {
+            return;
+        }
+        this.logger.error(`Worker thread for project #${this.projectNumber} exited unexpectedly with code ${code}`);
+        //emit() is async; this handler can't be, so void it
+        void this.emit('critical-failure', {
+            message: `The BrighterScript language server's worker thread crashed unexpectedly (exit code ${code}). This project (and any others sharing its worker thread) will stop responding until the language server is reloaded.`
```
TwitchBronBron: Is this message accurate? We have to fully reload the language server? Or could/would we recover?

src/lsp/worker/WorkerThreadProject.ts
```diff
@@ -325,14 +356,21 @@ export class WorkerThreadProject implements LspProject {
...
             disposable?.dispose?.();
         }
         this.disposables = [];
 
-        //move the worker back to the pool so it can be used again
+        //close our dedicated port and release our slot on the worker (which terminates the worker if it's now empty)
+        this.port?.close();
         if (this.worker) {
-            workerPool.releaseWorker(this.worker);
+            workerPool.releaseProject(this.worker);
```
TwitchBronBron, line 373: Do we have sufficient safeguards to ensure we _always_ release, even in the case of full crashes? Just want to make sure we don't end up with orphaned assignments.

src/LanguageServer.ts
```diff
@@ -77,6 +78,12 @@ export class LanguageServer {
      */
     private static projectActivationConcurrencyLimitDefault = 3;
 
+    /**
+     * The default maximum number of worker threads to use for running LSP projects. Can be overridden by
+     * per-workspace settings. Multiple projects will share a single worker thread once this limit is reached.
```
TwitchBronBron: this comment should likely say something more like "multiple projects will be spread evenly across existing worker threads once this limit is reached". (the current wording sounds like we dump them all on 1 thread.

src/lsp/worker/WorkerThreadProject.spec.ts
```diff
@@ -25,15 +46,37 @@ export function getWakeWorkerThreadPromise() {
     return wakeWorkerThreadPromise1;
 }
 
+//kick off the worker warmup at module-load time (mocha requires all spec files before running any test), so its
+//cold-boot cost overlaps with everything that runs before workerThreadWarmup's `before` hook awaits it below
+void getWakeWorkerThreadPromise().catch(() => {
```
TwitchBronBron: I think this might cause us to _always_ spin up a worker thread, even if I'm doing `it.only()` on some unrelated unit test. That's one of the things I was trying to protect against with the current implementation. 

Can we achieve this pre-baking only for tests that actually need to use workers?

TwitchBronBron (changes requested):
This is cool! 

One consideration, the term "worker" is getting a little fuzzy to me. Especially since it's going to be a user-facing setting. I think I originally used the term "worker" to represent a single thread which handled a single project. 

So it might be worth discussing/clarifying some of our terms. To clearly distinguish between a thread, a project, a specific connect of a project running inside a thread, etc. 

And maybe it's fine as is? I just know i was getting a little confused (and it's been a while since I last looked at this stuff).

src/LanguageServer.ts
```diff
@@ -403,6 +418,29 @@ export class LanguageServer {
...
     }
 
+    /**
+     * Get the max project workers setting from all workspaces and set the worker pool's cap to the lowest value found.
+     * This ensures that if the user has multiple workspaces open with different limits,
+     * we respect the most restrictive limit to avoid overwhelming the user's machine.
+     */
+    private syncMaxProjectWorkers() {
+        const limits = [...this.workspaceConfigsCache]
+            .map(x => x?.[1]?.languageServer?.maxProjectWorkers)
```
TwitchBronBron: Not crazy about the name of this setting, `maxProjectWorkers`. I could interpret to mean there's a 1-1 relationship between worker and project (but that's no longer true). 

This represents max number of threads that the language server will use. So something like `maxWorkerThreads` or `maxThreads` instead?

src/lsp/worker/WorkerThreadProject.ts
markwpearce: Yeah, accurate as of today - there's no recovery path for a crashed worker, so a reload really is required. Tweaked the wording to make clear that's a current limitation, not a permanent architectural one.

src/lsp/worker/WorkerThreadProject.ts
markwpearce: Yeah - when a worker crashes, `WorkerPool` removes the whole `WorkerEntry` from its list via the `'exit'` listener, so any sibling project's later `releaseProject(worker)` call just becomes a no-op instead of decrementing a stale count. No orphaned per-project bookkeeping survives a crash. Added a test for this in `WorkerPool.spec.ts`.

src/LanguageServer.ts
markwpearce: Fixed, reworded to say additional projects get spread evenly across the existing worker threads instead of each getting a dedicated one.

src/lsp/worker/WorkerThreadProject.spec.ts
markwpearce: You're right - moved the warmup back inside `before(workerThreadWarmup)`. Tradeoff is we lose the "start the cold boot as early as possible" for a full unfiltered suite run - it only starts once mocha actually reaches this suite's (or `ProjectManager.spec.ts`'s) hooks, same as before that optimization was added.

src/LanguageServer.ts
markwpearce: Renamed to `maxWorkerThreads`. Turns out this setting hasn't shipped in any release yet, so no need to keep a deprecated alias around - just did a clean rename.

src/lsp/worker/WorkerThreadProject.spec.ts
markwpearce: tests run (on my machine) in 14seconds now

TwitchBronBron (approved):

PR #1798, opened 2026-09-09
Modifies default max worker thread logic to be only as much as memory allows

markwpearce:
Instead of setting the default number of max threads to be the number of CPUs, this sets the default value to be as much as the machine's memory allows, considering there is an up-to 2GB V8 instance instantiated in each thread.

Logic is:
- Max of (Number of CPUs available vs. Memory in GB/2)
- at least 1 max thread

Added tests with sinon stubs for faking os.cpus() and os.totalmem()

### DIFF
--- a/src/lsp/worker/WorkerPool.ts
+++ b/src/lsp/worker/WorkerPool.ts
@@ -13,6 +13,7 @@ interface WorkerEntry {
 
 
 export const MEMORY_GB_PER_V8_INSTANCE = 2;
+export const MEMORY_GB_BUFFER = 1;
 
 /**
  * Gets the lower bound of number of CPUS available vs 2GB chunks of memory.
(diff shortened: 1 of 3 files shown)


src/lsp/worker/WorkerPool.ts
```diff
@@ -11,6 +11,20 @@ interface WorkerEntry {
...
+export const MEMORY_GB_PER_V8_INSTANCE = 2;
+
+/**
+ * Gets the lower bound of number of CPUS available vs 2GB chunks of memory.
+ * This allows as many concurrent threads as possible, while still allowing a full v8 instance per thread.
+ */
+export function getDefaultMaxWorkerThreads() {
+    const numCpus = os.cpus().length;
+    const memoryGB = Math.floor(os.totalmem() / (1024 * 1024 * 1024));
+    return Math.max(1, Math.min(numCpus, memoryGB / MEMORY_GB_PER_V8_INSTANCE));
```
TwitchBronBron: I'm thinking maybe we should allow a little buffer? For example, if we have 4 gigs free, not sure we should consume all of that because then the pc is fully swamped? So maybe shoot for 500-1000mb free? Something like: 
```ts
memoryGBAllowed = memoryGB - sevenHundredMegabytes;
```

TwitchBronBron (changes requested):

TwitchBronBron (approved):

PR #1817, opened 2026-09-11
Fix Array<Interface> callfunc argument false-positive across component scopes

### LINKED ISSUE #1816: Array<Interface> callfunc argument rejected as incompatible with itself across component scopes
## Description

Calling a component function through the `@` callfunc operator with an array-of-interface argument gets flagged as incompatible with its own declared parameter type - `Array<MyItem>` rejected as not compatible with `Array<MyItem>`.

It only shows up when:
- the interface is imported into two different component scopes (the one declaring the function, and the one calling it), and
- the callfunc target is typed as the specific generated node type (`roSGNodeA`), not the generic `roSGNode`

A scalar (non-array) interface param in the exact same setup type-checks fine, and calling the function from within its own component's scope (e.g. via `m.top@`) also works fine. It's only the combination of cross-scope + array + specifically-typed node reference that triggers it.

## Reproduction

`components/types.bs`
```brighterscript
interface MyItem
  sku as string
end interface
```

`components/A/A.xml`
```xml
<component name="A" extends="Group">
  <script type="text/brightscript" uri="A.bs" />
  <interface>
    <function name="doThing" />
  </interface>
</component>
```

`components/A/A.bs`
```brighterscript
import "pkg:/components/types.bs"

function doThing(items as MyItem[]) as void
  print items.Count()
end function
```

`components/B/B.xml`
```xml
<component name="B" extends="Group">
  <script type="text/brightscript" uri="B.bs" />
</component>
```

`components/B/B.bs`
```brighterscript
import "pkg:/components/types.bs"

sub callIt(aNode as roSGNodeA)
  item as MyItem = { sku: "abc" }
  aNode@.doThing([item])
end sub
```

`bsc --project bsconfig.json`:
```
components/B/B.bs:5:18 - error argument-type-mismatch: Argument of type 'Array<MyItem>' is not compatible with parameter of type 'Array<MyItem>'

5   aNode@.doThing([item])
                   ~~~~~~
```

Two changes independently make the error disappear:
- type `aNode` as `roSGNode` instead of `roSGNodeA` in `B.bs`
- change `doThing`'s param from `items as MyItem[]` to a scalar `item as MyItem` (and pass `item` instead of `[item]`)

## Expected behavior

`Array<MyItem>` should be recognized as compatible with itself no matter which component scope resolved the `import` - same as already works for a scalar interface param in the identical setup.

## Environment

- `brighterscript`: 1.0.0-alpha.52
- Reproduced via `bsc --project bsconfig.json` with a minimal `rootDir`/`files` config, no manifest needed

TacoJr01:
## Summary
Fixes #1816. Calling a component function through `@` callfunc with an `Array<MyItem>` argument was rejected as incompatible with its own declared parameter type whenever the interface was imported into two different component scopes and the callfunc target was typed as a specific generated node type (e.g. `roSGNodeA`) rather than generic `roSGNode`.

## Root cause
`ReferenceType.isTypeCompatible` has a fallback for comparing two same-named reference types that requires their `tableProvider` closures to be reference-identical. Two independently-resolved references to the same `.bs` interface (one from the declaring component's scope, one from the calling component's own scope) never share that closure, even though both resolve to the same declaration -- causing a spurious mismatch.

This only surfaces for arrays because `ArrayType.defaultType` permanently memoizes its element type on first access, and for a callfunc target that first access happens once against a shared, global `ComponentType` built while the declaring component's scope is transiently linked. By the time a different calling scope validates its own call, that cached reference can no longer live-resolve. A scalar param re-resolves fresh each time instead of being cached, so it self-heals.

## Fix
When two same-named references can't be proven identical only because they came from different scopes, treat them as compatible -- matching the "can't resolve, assume compatible" fallback already used elsewhere in this file.

## Test plan
- Added a test reproducing the issue (two components, shared imported interface, array param, `roSGNodeA` callfunc target) -- zero diagnostics now.
- Added a negative-case test confirming a genuinely incompatible array arg (`Array<integer>` vs `Array<MyItem>`) is still flagged.
- `npx mocha`: 4561 passing (2 new), 9 pending, 0 failing.
- `npm run lint`: 0 errors. `npm run build`: succeeds.
- Manually verified against the issue's exact repro with `bsc --project bsconfig.json`.

### DIFF
--- a/src/bscPlugin/validation/ScopeValidator.spec.ts
+++ b/src/bscPlugin/validation/ScopeValidator.spec.ts
@@ -6590,6 +6590,95 @@ describe('ScopeValidator', () => {
             ]);
         });
 
+        it('allows an array of an interface type as a callfunc arg, when the interface is imported into two different component scopes', () => {
+            program.setFile('components/types.bs', `
+                interface MyItem
+                    sku as string
+                end interface
+            `);
+
+            program.setFile('components/A.xml', trim`
+                <?xml version="1.0" encoding="utf-8" ?>
+                <component name="A" extends="Group">
+                    <script uri="A.bs"/>
+                    <interface>
+                        <function name="doThing" />
+                    </interface>
+                </component>
+            `);
+
+            program.setFile('components/A.bs', `
+                import "pkg:/components/types.bs"
+
+                function doThing(items as MyItem[]) as void
+                    print items.Count()
+                end function
+            `);
+
+            program.setFile('components/B.xml', trim`
+                <?xml version="1.0" encoding="utf-8" ?>
+                <component name="B" extends="Group">
+                    <script uri="B.bs"/>
+                </component>
+            `);
+
+            program.setFile('components/B.bs', `
+                import "pkg:/components/types.bs"
+
+                sub callIt(aNode as roSGNodeA)
+                    item as MyItem = { sku: "abc" }
+                    aNode@.doThing([item])
+                end sub
+            `);
+            program.validate();
+            expectZeroDiagnostics(program);
+        });
+
+        it('still catches a genuinely incompatible array arg to a callfunc, alongside a compatible interface array', () => {
+            program.setFile('components/types.bs', `
+                interface MyItem
+                    sku as string
+                end interface
+            `);
+
+            program.setFile('components/A.xml', trim`
+                <?xml version="1.0" encoding="utf-8" ?>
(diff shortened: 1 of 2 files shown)


markwpearce (approved):

TwitchBronBron (approved):

markwpearce:
Thanks for contributing!

PR #1820, opened 2026-09-15
Skip per-token toLowerCase+Set lookup for Token.isReserved

markwpearce:
Lexer.addToken computed \`isReserved: ReservedWords.has(text.toLowerCase())\` for every token, but only a Keyword-kind token can ever be reserved — identifier() already resolves reserved-word text to its specific keyword kind before addToken runs, so this was a wasted toLowerCase()+Set lookup on every single token (millions of them on a big project). Also — isReserved is never read anywhere internally, only carried through cloneToken for downstream consumers.

Added `ReservedTokenKinds` (derived from the existing `Keywords`/`ReservedWords` data, so it can't drift), and check kind membership instead — O(1), no allocation.

Small win, not a big one like #1819 — addToken's own CPU self-time was ~0.6% of the profiled run, and I can't isolate the allocation savings precisely from a sampling profile since it's smeared across every call rather than one hot node. Rough guess: a couple percent less heap churn. Free and zero-risk regardless.

Added tests for isReserved correctness (reserved vs. non-reserved keywords vs. identifiers, casing, and a round-trip check against every `Keywords` entry).


TwitchBronBron (approved):
Nice!

PR #1822, opened 2026-09-15
Lazily allocate SymbolTable siblings/complementsTables

markwpearce:
Every SymbolTable constructor was allocating `siblings`/`complementsTables` Sets and `symbolMap`/`typeCache` Maps unconditionally, even though most tables (a lot of ReferenceType/BscType member tables especially) never use them. Made all four lazy - allocated on first write, guarded the read sites with `?? []`/`?.`.

~117MB (~3%) of a profiled lint run's total allocation, across the two commits. Free, low-risk - all four follow the same simple "empty until written to" invariant, no external staleness like the `.parent` caching I tried and reverted earlier.


src/SymbolTable.ts
```diff
@@ -73,23 +73,25 @@ export class SymbolTable implements SymbolTypeGetter {
...
 
     /**
      * Add a sibling symbol table (which will be inspected first before walking upward to the parent
      */
     public addSibling(sibling: SymbolTable) {
+        this.siblings ??= new Set();
         this.siblings.add(sibling);
         return () => {
-            this.siblings.delete(sibling);
+            this.siblings?.delete(sibling);
```
TwitchBronBron: Splitting hairs here, but why would you need this optional chain when you clearly allocated `this.siblings` above?

src/SymbolTable.ts
```diff
@@ -179,7 +183,7 @@ export class SymbolTable implements SymbolTypeGetter {
             }
 
             //look through any sibling maps next
-            for (let sibling of currentTable.siblings) {
+            for (let sibling of currentTable.siblings ?? []) {
```
TwitchBronBron: If we're looking for improvements to memory churn, it might be worth wrapping this in an if statement instead of allocating an empty array just to skip it? 

```
if(currentTable.siblings) {
  for (....
```

TwitchBronBron (approved):
Two comments, do what you want with them, but this looks good otherwise.

src/SymbolTable.ts
markwpearce: Good call, done - if-gated the sibling/symbolMap loops instead of `?? []`.

src/SymbolTable.ts
markwpearce: Fair - fixed, that .siblings is always set there by the time the disposer runs.

